PR #19, opened 2014-12-07
Add updated XCode project

btb:
Adds an XCode project for dxx-rebirth, in contrib/, and deletes the old d1x-rebirth and d2x-rebirth projects, which did not work.
Still uses scons to build, settings are editable from within XCode.


kreatordxx:
I'm getting there. I managed to update my system to Yosemite (with a few hiccups) and have merged the 'Xcode' branch into my local copy. I can open the Xcode project and I'm now going through the process of reinstalling DXX-Rebirth's dependencies, as I did a clean install of Yosemite. A lot has changed since I last worked on this!

vLKp:
> I'm now going through the process of reinstalling DXX-Rebirth's dependencies, as I did a clean install of Yosemite.

If you encounter any dependencies where scons starts a build, but a source file fails to build or the eventual link fails, please file a ticket for it.  My goal is that the scons configure phase should catch any missing dependencies so that it can print a useful diagnostic, rather than let the build proceed and then die partway through.

Conversely, if you find that an scons configure test fails to accept a dependency that you have installed, file that too.  You can use the **sconf_**``*FOO* overrides to skip checks as needed, but ideally all checks should run and produce the result you expect.

kreatordxx:
I’m glad you mentioned that, because it’s failing to find physfs.h. It’s saying

scons: *** Header physfs.h is missing or unusable.  See build/d1x-rebirth-opengl/sconf.log for details.  Stop.

I downloaded the PhysicsFS 2.0.3 tarball and started from scratch. I tried building PhysicsFS using the Xcode project generated by CMake, but it simply wouldn’t build the ‘archive’ (figured out that means release in Xcode-6-speak) of the dynamic library. Therefore, I made it generate the ‘build’ directory (still on PhysicsFS here) for the Unix Makefiles. Everything seemed to install OK, it said it put physfs.h in /usr/local/include/physfs.h (as well as installing /usr/local/lib/libphysfs.dylib). However, I’m still getting the problem where DXX-Rebirth can’t find physfs.h. How did you go with this Bradley? Did you install PhysicsFS the Unix way?

This is going to get even more fun when I try and make a release of DXX-Rebirth that includes the PhysicsFS library inside it, which is what I used to do. I can’t expect regular Mac users to download the PhysicsFS tarball and compile from source just to get DXX-Rebirth to work. Ever since PhysicsFS went to CMake I have been having regular hassles with it. Before I found a hackish solution of directly linking to physfs in the physfs source directory but I don’t think that will work now, because the hack was in Xcode, not SCons. Now with Xcode completely overhauled, long after PhysicsFS has been updated, it just adds another ‘dimension of complexity’ to it.

kreatordxx:
Filed a separate issue. I can confirm it occurs in SCons via Terminal (Mac OS X command line), independently of the Xcode project.

I'd love to merge this Xcode project file in (I probably wouldn't have even thought to link to SCons - great way to avoid doubling up on the source file dependency lists) but would first like to confirm it works for me too.

btb:
I used homebrew to install scons, physfs, and any other dependencies there might have been. Homebrew is quite nice, it cleanly installs everything into its own hierarchy under /usr/local.

kreatordxx:
Hi Bradley, I managed to get D1X-Rebirth and D2X-Rebirth to build and work using Xcode, but when it tries to launch it in the debugger I get an error message saying it can’t run it. In the console I get
> error: executable doesn't exist: '/Users/bradleyb/src/descent/dxx-rebirth/build/d2x-rebirth-opengl/D2X-Rebirth.app'
> error: failed to launch '/Users/bradleyb/src/descent/dxx-rebirth/build/d2x-rebirth-opengl/D2X-Rebirth.app’
Looks like there are some hard coded paths to the executable, but I’m not sure where to find the editable paths with this new version of Xcode.=

btb:
Yeah, if there is a way to make it use a relative path for an external build system, I could not find it.
You can change it under Edit scheme->Run->Info->Executable->Other

general info: https://developer.apple.com/library/mac/recipes/xcode_help-scheme_editor/Articles/SchemeDialog.html

kreatordxx:
Yes, I searched and searched too and could not find a way to make it a relative path. Even changing the custom working directory (Options in the schemes dialog under ‘run’) to an absolute path had no effect on the executable path. Another guy asked the same question and got no answer: http://osdir.com/ml/general/2014-07/msg58636.html <http://osdir.com/ml/general/2014-07/msg58636.html> (Relative Path in Xcode scheme)=

kreatordxx:
Merged (#19).

PR #33, opened 2015-01-29
Recorddemosofallgames

jeffersoncarpenter:
I wasn't sure if game_setup was the best place to check for the PlayerCfg property and start recording a demo.

Anyway, let me know what you think about this code.


jeffersoncarpenter:
I reverted those questionable changes.  The patch got smaller :)

vLKp:
Implemented in 07869d9dba381d8bd1ea830282af33c245038806 as a pair of command-line options.  Users who want to use this intermittently can set the options when they want it.  Users who want it permanently can set it in their `d1x.ini` / `d2x.ini` file.

jeffersoncarpenter:
Excellent, glad the feature's in there :)

PR #34, opened 2015-02-06
more OS X issues

btb:
first commit fixes this problem:
```
Compiling d2x-rebirth build/d2x-rebirth-opengl/ similar/arch/sdl/jukebox.cpp
similar/arch/sdl/jukebox.cpp:143:10: error: use of overloaded operator '[]' is ambiguous (with operand types 'RAIIdmem<char []>' and 'uint_fast32_t' (aka 'unsigned int'))
        list_buf[length] = '\0';        // make sure the last string is terminated
        ~~~~~~~~^~~~~~~
        typename tt::enable_if<tt::rank<T>::value != 0, element_type &>::type operator[](std::size_t i) const
                                                                              ^
        typename tt::enable_if<tt::rank<T>::value != 0, element_type &>::type operator[](int i) const
                                                                              ^
        list_buf[length] = '\0';        // make sure the last string is terminated
                ^
1 error generated.
```

couple more on the way

PR #43, opened 2015-02-16
Osx issues

btb:


vLKp:
fdf7bb01ed7b13f11321cbb425c7a3a2e153854d breaks the mingw32 build.  On mingw32, `size_t` is `unsigned int`, so you cannot define a separate overload for `unsigned int`:
```
In file included from common/2d/bitmap.cpp:30:0:
common/include/u_mem.h:119:72: error: 'typename std::enable_if<(std::rank<_Tp>::value != 0), typename std::unique_ptr<T, RAIIdmem_deleter<T> >::element_type&>::type RAIIdmem<T>::operator[](unsigned int) const' cannot be overloaded
  typename tt::enable_if<tt::rank<T>::value != 0, element_type &>::type operator[](unsigned int i) const
                                                                        ^
common/include/u_mem.h:115:72: error: with 'typename std::enable_if<(std::rank<_Tp>::value != 0), typename std::unique_ptr<T, RAIIdmem_deleter<T> >::element_type&>::type RAIIdmem<T>::operator[](std::size_t) const'
  typename tt::enable_if<tt::rank<T>::value != 0, element_type &>::type operator[](std::size_t i) const
                                                                        ^
```
Why do you need a separate overload for `unsigned int`?  What breaks if we have only the overload for `std::size_t`?

btb:
I get the error referenced in #34.

I think it's because there are implicit conversions from uint_fast32_t to both size_t and unsigned int, so the compiler doesn't know which operator[] to use.

btb:
I get the error referenced in #34.

`uint_fast32_t` is an `unsigned int` for me, which seems to have implicit conversions to both `size_t` and `int`, so the compiler doesn't know which operator[] to use.

btb:
deleting the `operator[](int i)` also fixes it for me

btb:
I don't know much about `enable_if` but I suppose with the right parameters there it we could make it ignore that overload in my situation?

vLKp:
> deleting the `operator[](int i)` overload also fixes it for me

Unfortunately, deleting that overload breaks the mingw32 build also:
```
In file included from common/2d/rle.cpp:33:0:
common/include/u_mem.h:115:72: note: candidate 1: typename std::enable_if<(std::rank<_Tp>::value != 0), typename std::unique_ptr<T, RAIIdmem_deleter<T> >::element_type&>::type RAIIdmem<T>::operator[](std::size_t) const [with T = unsigned char []; typename std::enable_if<(std::rank<_Tp>::value != 0), typename std::unique_ptr<T, RAIIdmem_deleter<T> >::element_type&>::type = unsigned char&; std::size_t = unsigned int]
  typename tt::enable_if<tt::rank<T>::value != 0, element_type &>::type operator[](std::size_t i) const
                                                                        ^
common/2d/rle.cpp:601:14: note: candidate 2: operator[](RAIIdmem<unsigned char []>::pointer {aka unsigned char*}, int) <built-in>
    temp[4 + i] = ptr2 - start;
              ^

```
Fixing the build after that removal would require changing data structures in some places, since RAIIdmem is sometimes subscripted using a value from a structure (`inputbox->UI_GADGET_INPUTBOX::position`).

> I don't know much about `enable_if` but I suppose with the right parameters there it we could make it ignore that overload in my situation?

Maybe.  I tried adding ` && !tt::is_same<unsigned, std::size_t>::value` to the new declaration so that it would be disabled on mingw32.  That failed.  Guarding the parameter inside `enable_if` also failed.  I think in both cases it is because the expression is not dependent on a template parameter, so SFINAE does not trigger.  Although it would make the code uglier, I am inclined to fix this by removing `operator pointer() const` and converting the affected sites to use an explicit call to `->get()`.

See [cppreference:enable_if](http://en.cppreference.com/w/cpp/types/enable_if) for a full explanation.  The short version is that if the compile-time boolean expression passed as the first argument to `enable_if<B, T>` is false, then the expression is ill-formed and SFINAE removes it from the overload set.

vLKp:
Please checkout and test f900d4a5ee32c2eb8d0f42e216445072c2b657b1.  It removes `operator pointer() const` and both `operator[]() const`.  It builds fine for Windows and Linux.

btb:
that works fine. pull req rebased without my overload

vLKp:
I think the revert is unneeded.  On most platforms, `std::size_t` is the widest native type.  It is probably more suitable for indexing an array, since otherwise the index must be zero-extended.

btb:
yeah, it's not needed for me, I just figured you might want it for consistency

PR #47, opened 2015-03-21
RPi on unification

derhass:
With these changes, the RPi port builds again with current unification branch.


derhass:
I've creared a new version of this branch and a new pull request.

vLKp:
I believe it is possible to submit a replacement of the contents of a pull request without switching to a new pull request.

Don't worry too much about the incorrect log message.  I don't think anyone reads the old logs that closely.

PR #51, opened 2015-03-21
RPi on unification

derhass:
With these changes, the RPi port builds again with current unification branch.

This is the updated version (replacing https://github.com/dxx-rebirth/dxx-rebirth/pull/47) with the correct git message for the merge :)

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1539,10 +1539,10 @@ class DXXCommon(LazyObjectConstructor):
 		if (self.user_settings.raspberrypi == 1):
 			message(self, "Raspberry Pi: using VideoCore libs in \"%s\"" % self.user_settings.rpi_vc_path)
 			env.Append(CPPDEFINES = ['RPI', 'WORDS_NEED_ALIGNMENT'])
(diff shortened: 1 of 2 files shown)


zicodxx:
Thank you. I will take a quick look over them and merge afterwards. :)

derhass:
Please note that building d1x and d2x works with these patches, but d2x doesn't seem to work. I'll have to invesitgate this further. However, those patches won't hurt in either case.

zicodxx:
Does it only break the RPi build? in case if so, we can still patch this up later.

derhass:
It doesn't break the build. Those patches don''t break anything (as far as I can tell), and normal desktop builds are totally unaffected.

The game just crashes on the Pi when trying to start a game after the main menu. It might be the case that it is just an out of memory conditon. I plan to debugg this. But building on the RPi with the new C++ code base takes several hours (instead of half an hour before), so things are moving slowly. I should set up some cross-compile environment.

zicodxx:
Maybe it's also just a general regression. Yes, your changes don't look as if they break anything - let alone on the PC build. Merging. :)

PR #53, opened 2015-03-22
cleaned-up syncgl patch for unification

derhass:
This is my syncgl-patch ported to unification branch, and additionally cleaned up.

The OpenGL extension stuff is now moved to it's own file, as well as the sync code itself. I created the new common/arch/ogl directory for this, and updated the SConstruct file appropriately.

The new default is the conservative AUTO mode which enables FENCE_SLEEP mode if available, and does nothing if not. FENCE_SLEEP should have no negative effects (while
the FINISH modes might degrade performance a bit, and in general should only be enabled if the user needs them).

The only problem with this implementation is that we are not able to use the GL_ARB_sync extension on OSX (and are limited to the FINISH modes there), because we are using a legacy GL context and cannot access modern GL features. Fixing this would require rewriting the whole GL renderer to modern GL, which is something left for future patches...

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -179,10 +179,18 @@ class ConfigureTests:
 		if r and forced != self.sconf_assume_success:
 			caller_modified_env_flags.restore(context.env)
 			context.env.Replace(CPPDEFINES=env_flags['CPPDEFINES'])
-			for d in successflags.pop('CPPDEFINES', []):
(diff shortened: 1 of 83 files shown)


derhass:
I've updated the patch to resolve any merge conflicts with Kp's latest commits.

PR #54, opened 2015-03-25
Use Travis for CI

btb:
By adding .travis.yml and registering your repo at travis-ci.org, you can have the build automatically tested any time code is checked in. Even pull requests can be automatically checked.

I'm not sure when or if it will be possible to add a native OS X to the supported builds at travis, but having automatic checks on linux for clang and gcc should be pretty useful. I imagine it might be possible to get cross-compiles working for mingw and os x.

My fork is on travis here: https://travis-ci.org/btb/dxx-rebirth
Unfortunately it doesn't build yet. Perhaps you could help me figure out why.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,10 @@ compiler:
 - clang
 
 before_install:
+- sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
 - sudo apt-get update -qq
 - sudo apt-get install -y libphysfs-dev libsdl1.2-dev libsdl-mixer1.2-dev
+- if [ "$CXX" = "g++" ]; then sudo apt-get install -qq g++-4.8; fi
+- if [ "$CXX" = "g++" ]; then export CXX="g++-4.8" CC="gcc-4.8"; fi
 
 script: scons
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: cpp
 
 compiler:
 - gcc
-- clang
+#- clang
 
 before_install:
 - sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+language: cpp
+
+compiler:
+- gcc
+- clang
+
+before_install:
+- sudo apt-get update -qq
+- sudo apt-get install -y libphysfs-dev libsdl1.2-dev libsdl-mixer1.2-dev
+
+script: scons
--- a/contrib/dxx-rebirth.xcodeproj/project.pbxproj
+++ b/contrib/dxx-rebirth.xcodeproj/project.pbxproj
@@ -8,6 +8,7 @@
 
 /* Begin PBXFileReference section */
 		1469259F1A57A0E50014A414 /* ntstring.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = ntstring.h; sourceTree = "<group>"; };
+		14AC23E81ABE1AF60067931F /* .travis.yml */ = {isa = PBXFileReference; lastKnownFileType = text; path = .travis.yml; sourceTree = "<group>"; };
 		14F8A91B1A2D6320003AA93A /* 2dsline.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; path = 2dsline.cpp; sourceTree = "<group>"; };
 		14F8A91C1A2D6320003AA93A /* bitblt.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; path = bitblt.cpp; sourceTree = "<group>"; };
 		14F8A91D1A2D6320003AA93A /* bitmap.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; path = bitmap.cpp; sourceTree = "<group>"; };
@@ -482,6 +483,7 @@
 		141FE1011A2D62B100A4FB18 /* .. */ = {
 			isa = PBXGroup;
 			children = (
+				14AC23E81ABE1AF60067931F /* .travis.yml */,
 				14F8AA1E1A2D6320003AA93A /* common */,
 				14F8AA2B1A2D6320003AA93A /* contrib */,
 				14F8AA2C1A2D6320003AA93A /* COPYING.txt */,
--- a/similar/main/newmenu.cpp
+++ b/similar/main/newmenu.cpp
@@ -24,7 +24,7 @@ COPYRIGHT 1993-1999 PARALLAX SOFTWARE CORPORATION.  ALL RIGHTS RESERVED.
  */
 
 #include <stdio.h>
-#include <stdlib.h>
+#include <cstdlib>
 #include <string.h>
 #include <stdarg.h>
 #include <ctype.h>


vLKp:
btb wrote:
> Perhaps you could help me figure out why.

We could try, if you post some error messages. :smile:

btb:
compiling with gcc: https://travis-ci.org/btb/dxx-rebirth/jobs/55753191

```
In file included from common/arch/sdl/joy.cpp:17:0:
common/include/event.h: In constructor ‘d_change_event::d_change_event(int)’:
common/include/event.h:65:42: error: no matching function for call to ‘d_event::d_event(<brace-enclosed initializer list>)’
common/include/event.h:65:42: note: candidates are:
common/include/event.h:51:8: note: d_event::d_event()
common/include/event.h:51:8: note:   candidate expects 0 arguments, 1 provided
common/include/event.h:51:8: note: constexpr d_event::d_event(const d_event&)
common/include/event.h:51:8: note:   no known conversion for argument 1 from ‘event_type’ to ‘const d_event&’
common/include/event.h:51:8: note: constexpr d_event::d_event(d_event&&)
common/include/event.h:51:8: note:   no known conversion for argument 1 from ‘event_type’ to ‘d_event&&’
common/include/event.h: In constructor ‘d_select_event::d_select_event(int)’:
common/include/event.h:74:43: error: no matching function for call to ‘d_event::d_event(<brace-enclosed initializer list>)’
common/include/event.h:74:43: note: candidates are:
common/include/event.h:51:8: note: d_event::d_event()
common/include/event.h:51:8: note:   candidate expects 0 arguments, 1 provided
common/include/event.h:51:8: note: constexpr d_event::d_event(const d_event&)
common/include/event.h:51:8: note:   no known conversion for argument 1 from ‘event_type’ to ‘const d_event&’
common/include/event.h:51:8: note: constexpr d_event::d_event(d_event&&)
common/include/event.h:51:8: note:   no known conversion for argument 1 from ‘event_type’ to ‘d_event&&’
```

btb:
compiling with clang: https://travis-ci.org/btb/dxx-rebirth/jobs/55753192

```
In file included from common/2d/bitblt.cpp:36:
In file included from common/include/ogl_init.h:43:
In file included from common/include/3d.h:32:
common/include/vecmat.h:177:38: error: unused function 'operator*'
      [-Werror,-Wunused-function]
static constexpr vm_distance_squared operator*(const vm_distance &a, con...
                                     ^
```

btb:
The clang problems are mostly the same as what I get in OSX now, with the extra warning checks.

btb:
after fixing the warnings with miscellaneous hacks (https://github.com/btb/dxx-rebirth/compare/travis...btb:travis2), I still get this:

```
similar/main/render.cpp:1036:33: error: no member named 'emplace' in
      'std::unordered_map<short, render_state_t::per_segment_state_t,
      std::hash<short>, std::equal_to<short>, std::allocator<std::pair<const
      short, render_state_t::per_segment_state_t> > >'
        auto p = rstate.render_seg_map.emplace(segnum, render_state_t::p...
                 ~~~~~~~~~~~~~~~~~~~~~ ^
```

vLKp:
[btb wrote](https://github.com/dxx-rebirth/dxx-rebirth/pull/54#issuecomment-86252358):
> compiling with gcc: https://travis-ci.org/btb/dxx-rebirth/jobs/55753191

Travis seems to be using an old gcc.  I get the same error if I use gcc-4.6.  Using brace-initialization for base-class construction does not work before gcc-4.7, but allows initializing a base which has no user-defined constructors.  Using parentheses for base-class construction works with older compilers, but requires a base class constructor that accepts the listed initializer.

[btb wrote](https://github.com/dxx-rebirth/dxx-rebirth/pull/54#issuecomment-86252750):
> compiling with clang: https://travis-ci.org/btb/dxx-rebirth/jobs/55753192

I cannot reproduce this one with Linux/amd64 clang v3.4.2 using gcc-4.8.3 stdlibc++.  If I had to guess, I would say that it is a bug in clang's -Wunused-function handling.  They suppress unused function warnings for `static inline` functions, but apparently not for `static constexpr` functions, even though [constexpr implies inline](http://en.cppreference.com/w/cpp/language/constexpr).  I have been thinking about moving that `operator*` into class scope to suppress messages about it being rejected when the compiler tries to resolve other forms of `operator*`.  That move may also work around the overzealous clang warning.

[btb wrote](https://github.com/dxx-rebirth/dxx-rebirth/pull/54#issuecomment-86265119):
> I still get this: https://travis-ci.org/btb/dxx-rebirth/jobs/55886446

This is probably due to an outdated or incomplete standard library.  [`std::unordered_map has emplace(...)`](http://en.cppreference.com/w/cpp/container/unordered_map/emplace)

btb:
Travis uses Ubuntu 12.04 LTS and so gcc 4.6.3 is the system compiler.

I take it you're not really interested in supporting the older compiler?
I did figure out how to install and use 4.8 in Travis, so the gcc build now succeeds. 

It might be harder to get clang working

vLKp:
[btb wrote](https://github.com/dxx-rebirth/dxx-rebirth/pull/54#issuecomment-86376247):
> gcc 4.6.3 is the system compiler.
> I take it you're not really interested in supporting the older compiler?

As far as I know, GCC 4.6.x was [closed upstream](https://gcc.gnu.org/ml/gcc/2013-04/msg00132.html) after [GCC 4.6.4 was released 2013-04-12](https://gcc.gnu.org/ml/gcc/2013-04/msg00131.html).  I am not happy about supporting it, but since Ubuntu has an LTS that uses it, I will try to fix problems reported with it.  Beware that it has a [known bug with template parameters](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=51064) that will cause a bogus warning.  The bug was fixed in gcc 4.7, but the fix was never backported.  Commit 9fe3fdd2bbcb3b817aa982d416a068f0e319ac58 will try to detect this bug and suppress the warning.

[btb wrote](https://github.com/dxx-rebirth/dxx-rebirth/pull/54#issuecomment-86376247):
> It might be harder to get clang working

What is wrong with clang now?  Commit 7045b44386291d8825edaf844e7bbe52a5e8ae67 should have fixed the `static constexpr` problem you reported above.

vLKp:
After further investigation, I think GCC 4.6.x support for C++11 is insufficient for the Rebirth code as it is now.  If this had been caught months ago when I first began adding features that GCC 4.6.x could not handle, I might have written it differently.  However, no one reported issues until you began the Travis work and there are now too many things that GCC 4.6.x chokes on, some of them seriously.  Fixing those sites is not fun, and in my opinion is not necessary since even LTS systems can readily install a more modern GCC.

I will accept patches to improve GCC 4.6.x support if they do not make the code harder to maintain, but I will not spend much effort on fixing GCC 4.6.x support.  As time permits, I may add some SConf checks to detect inadequate compilers early.

vLKp:
As of 147bc4ae0a1aa6c7cf092ae3abd096430e52e310, this commit no longer merges cleanly.  I see how to fix it, but per [btb's last comment](https://github.com/dxx-rebirth/dxx-rebirth/pull/54#issuecomment-86376247), clang on Travis is still broken.  Additionally, Travis itself seems to be broken.  It returns only a blank doument no matter what URL I load.  I am open to adding support for this if Travis can be made to work, but with no error messages and no ability to use the site, there is nothing I can do to improve the situation.

btb:
Should merge cleanly again. Disabled the clang build for now.

btb:
Creating an account on travis and switching on builds for this repo is all that's needed now to get it running.

vLKp:
Once travis serves something other than a blank page, I will look into it.

btb:
hmm, it's working for me: https://travis-ci.org/btb/dxx-rebirth

PR #58, opened 2015-04-05
Fix the INTEL_*() byte swapping functions

derhass:
This fixes a regression from 6e96ff3e0073139bbb77f57c4eda4dfaa1b1c2f7 where `WORDS_BIGENDIAN` somehow was confused with `WORDS_NEED_ALIGNMENT`, so that setting `WORDS_NEED_ALIGNMENT` resulted in assuming a big endian machine. This completely broke the RPi build, which is little endian.


vLKp:
Cherry-picked as e00b318bee06f075e5af3b9ccedc65c4490172ab to add a Fixes tag on the off chance anyone looks at those.

PR #67, opened 2015-05-23
change u_int*_t to C99 standard uint*_t

btb:
This removes the vestiges of weird non-standard u_int*_t types.

You'll probably want to just delete that whole section in pstypes.h, since these types have been around since C99, but I haven't tested any platforms that might be affected.


Matt1360:
:+1: looks good to me.  I also prefer the `stdint.h` types.

vLKp:
The change looks fine.  I will merge it after I finish running build tests on my target platforms.

PR #106, opened 2015-06-14
Add the ability to specify missions directory from command line

Mako88:
This is so that Rebirth and Retro can share a missions directory as well as a data directory.

Also...something's screwy and github thinks this is still part of the nonlinear joystick stuff for some reason...


vLKp:
Could you explain why this is needed?  I thought Retro still read/wrote the `d1x-rebirth`/`d2x-rebirth` directories, in which case it should automatically be sharing missions.

Mako88:
It compiled fine for me, but I'll change that statement to prevent the conversion.

I just submitted a pull request to Retro today to change where it stores it's files (to d*x-retro). This is because configurations between Retro and Rebirth are no longer the same, so if both programs use the same player file you have to reset all your settings. Being able to set missions manually will allow the users to have a hog dir and missions dir that are shared between both programs, while everything else is put in the respective "d*x-whatever" folders.

vLKp:
> This is because configurations between Retro and Rebirth are no longer the same, so if both programs use the same player file you have to reset all your settings.

That sounds bad.  I see no recent changes in Retro though.  Has this been broken for a long time?  What was changed that broke the player file?  The binary player file cannot be extended, and the text player file should be extensible.

Mako88:
I may have mis-spoken...it may not be in the player files. But here is what must be reset every time you switch between programs:

Transparency
Colored Lighting
Joystick Sensitivity & Deadzone
Autoselect
Various misc options (not sure off-hand which all)

Anyway, that gets pretty annoying to have to set every time, (and as far as I know, this problem will just get worse as the two projects add separate settings).

Mako88:
OK, all of the issues you pointed out are fixed now.

Mako88:
I accidentally included the .ini improvements to this pull request. I was planning on making that a separate pull request after this one (since it's a totally separate thing).

vLKp:
> I may have mis-spoken...it may not be in the player files. But here is what must be reset every time you switch between programs:

Those settings predate Retro.  Do you also lose them if you switch between Rebirth 0.58.1 and current HEAD?  What was the first commit where settings were lost?

vLKp:
@Mako88: could you respond to my Jun 14 comments with regard to what settings are lost?  I would rather deal with the cause of the lost settings than add a workaround based on keeping multiple copies of the files.

Mako88:
This seems kinda screwy...why is this part of the Nonlinear Joystick pull request?

I'll check tomorrow if I get a chance and let you know. I think (and this is entirely speculation, because I haven't looked at the code), that settings are stored and read by index, and because Rebirth and Retro have different settings the index is wrong. But I could be completely wrong about that.

vLKp:
Mako88 wrote:
> This seems kinda screwy...why is this part of the Nonlinear Joystick pull request?

You set it that way.

Mako88:
After further testing, it's not that settings aren't being remembered. It's that Rebirth has settings that Retro doesn't and vice-versa, and so if someone uses port A, then switches to port B, then switches back to port A, the settings that port A has that port B doesn't, will no longer be remembered.

Mako88:
I just realized...Retro and Rebirth won't be compatible after .60 anyway because of .dll differences. If you want to implement this ability just cause it'd be nice to have, go ahead, but I'd actually like to have this closed so I can clean up the fork that use to make pull requests from.

NOTE: I would recommend implementing the help text cleanup that I did here.

vLKp:
> It's that Rebirth has settings that Retro doesn't and vice-versa

Right.  Rebirth has never had the ability to remember free-form text options that it does not understand, so the Retro fork also forgets options it does not know.  The solution here is to rebase Retro forward so that it learns our new options and to teach it not to overwrite stock Rebirth files.

> I just realized...Retro and Rebirth won't be compatible after .60 anyway because of .dll differences.

What dll differences?

> I'd actually like to have this closed so I can clean up the fork that use to make pull requests from.

You can delete branches from your repository whenever you like.  Github will maintain the required link here, so that the pull request remains viable.

Mako88:
Just updated dlls...Unification doesn't require mikmod.dll and so won't ship with it, whereas Retro does.

vLKp:
I do not think we have changed our list of required DLLs.  Even if we did, Retro and Rebirth can still be co-installed, so the settings clobber problem will persist as long as outdated Retro builds try to use Rebirth data directories.

Mako88:
Right. I've already decided I'm going to remove Retro from the installer when 0.60 comes out. So if we want to have the ability to specify mission directories, you can merge this pull request, but don't do it for Retro/Rebirth compatibility.

vLKp:
Automatic merge failed, but in looking at it, there is a simpler way to do this, if we assume support for PhysFS v2.x.  It was released ~6 years ago, so I am comfortable relying on it.  The v2.x release has support for mounting specified real directories at arbitrary virtual directories, rather than only at the root of the virtual directory tree.  I added an option `-add-missions-dir` which takes a real filesystem path (e.g. `C:\Users\Me\Descent\my-shared-missions`) and maps it to the virtual `missions/` directory, so when the game tries to enumerate `missions/*`, it searches both the traditional missions directory and the one specified to `-add-missions-dir`.  This option does not stack, so you can only add one extra directory.  PhysFS can support more, if we had a need to allow multiple directories.

I guarded the new option with `#if PHYSFS_VER_MAJOR >= 2`.  If someone tries to build with PhysFS v1.x, the new option will be compiled out and will trigger a diagnostic if the user tries to pass it.  I do not know if PhysFS v1.x is even still usable with Rebirth, but this commit will not be the one to break it.

Mako88:
OK, since -add-missions-dir was implemented and the .ini/help fixes were also added, I'm closing this pull request.

PR #144, opened 2015-09-25
Rearrange Advanced Hosting Menu

Mako88:


vLKp:
Rather than doing a revert of HEAD, you could just unwind the branch by one commit, which would make for cleaner history.  It is too late to do that now that it has been merged, though.

PR #147, opened 2015-09-27
Add Autoselect After Firing

Mako88:


Mako88:
@vLKp you can merge this whenever you'd like. Because I don't fully understand your use of the ternary operator, I reverted that commit to allow you to do it yourself. Also, you may want to implement NoFireAutoselect as a 3-way when you do it.

PR #171, opened 2015-12-22
change namespace dcx to dsx in common/include/interp.h

derhass:
Some functions only present in the `WORDS_BIGENDIAN` and `WORDS_NEED_ALIGNMENT` case were declared in `namespace dcx` in the header `common/include/interp.h`, while they were defined in `namepsace dsx` in `similiar/3d/interp.cpp`. This broke the RPi build. This changes the declaration to match the definition.

Fixes: 32051298ae3f5a1ce5cae21377b2a7f99d9a649f

### DIFF
--- a/common/include/interp.h
+++ b/common/include/interp.h
@@ -63,7 +63,7 @@ int g3_poly_get_color(const uint8_t *model_ptr);
 }
 #endif
 
-namespace dcx {
+namespace dsx {
 #ifdef WORDS_BIGENDIAN
 // routine to convert little to big endian in polygon model data
 void swap_polygon_model_data(ubyte *data);


vLKp:
Are these functions sensitive to the game being built?  Only functions which compile to different code depending on D1 vs D2 should be placed in namespace dsx.

vLKp:
This should be resolved by [Use namespace dcx for WORDS_BIGENDIAN/WORDS_NEED_ALIGNMENT functions](/dxx-rebirth/dxx-rebirth/commit/70ffee39cd09285225599bb02cf30be3275fb13d).

derhass:
I can confirm that 70ffee39cd09285225599bb02cf30be3275fb13d fixes the issue. RPi builds now work again with a vanilla unification/master checkout.

PR #190, opened 2016-03-21
Fix editor crash introduced in unification.

Jayman2000:
Imported functions and variables relating to med-mark-start and med-mark-end (both in common/include/editor/kdefs.h) that were somehow lost from v0.58.1 to unification. If dxx-rebirth was compiled with both editor=True and debug=True, it would crash as soon as the editor was opened.

### DIFF
--- a/common/include/editor/kdefs.h
+++ b/common/include/editor/kdefs.h
@@ -181,7 +181,10 @@ int med_keypad_goto();
 int med_increase_tilings();
 int med_decrease_tilings();
 int ToggleAutosave();
+#ifndef NDEBUG
+int MarkStart();
 int MarkEnd();
+#endif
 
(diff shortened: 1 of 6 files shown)


vLKp:
NAK.  Adding dead code is the wrong approach.

vLKp:
On further investigation, there is no crash here.  When an undefined function was referenced, it was reported using `Error` (and later `UserError`), both of which are fatal.  `Error` aborted the program.  `UserError` calls `exit(1)`.  [Remove UserError on unknown MED function](/dxx-rebirth/dxx-rebirth/commit/178964348079734e427147edc6d8ba66bedc46b8) fixes this by making undefined functions produce an urgent warning, then ignoring them.

Issue #180, opened 2016-03-01
.ini parsing doesn't allow inline comments with paramaters


Mako88:
With the new .ini parsing, you can't have "-maxfps 30  ;a comment". You should be able to, so that people don't have to remove the comment lines already in the .ini file when enabling an option.

vLKp:
That should already work.  Either way, ini parsing is an area (one of many) that needs a rework, but has been low priority.  I think it is backwards that arguments from the .ini file are processed **after** the command line, since it means that the .ini overrides command line arguments.  If you have `-pilot a` in your .ini file, specifying `-pilot b` on the command line has no effect.

Mako88:
I agree, command-line should take precedence. I think this should be added to the 0.60 milestone, since people will see errors when trying to start the game after just removing the semi-colon in front of an argument with a parameter.

Mako88:
Closing since #194 has been added.

PR #194, opened 2016-05-10
.ini Parsing Improvements

Mako88:
This fixes bug #180

### DIFF
--- a/similar/misc/args.cpp
+++ b/similar/misc/args.cpp
@@ -105,7 +105,7 @@ static void AppendIniArgs(const char *filename, Arglist &Args)
 		PHYSFSX_gets_line_t<1024> line;
 		while (Args.size() < MAX_ARGS && PHYSFSX_fgets(line, f))
 		{
-			const auto separator = " \t\n\v\f\r";
+			const auto separator = " \t";
 			for(char *token = strtok(line, separator); token != NULL; token = strtok(NULL, separator))
 			{
 				if (*token == ';')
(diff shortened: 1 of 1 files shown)


similar/misc/args.cpp
```diff
@@ -105,7 +105,7 @@ static void AppendIniArgs(const char *filename, Arglist &Args)
 		PHYSFSX_gets_line_t<1024> line;
 		while (Args.size() < MAX_ARGS && PHYSFSX_fgets(line, f))
 		{
-			const auto separator = " ";
+			const auto separator = " \t\n\v\f\r";
```
vLKp: You do not need `\n` or `\r` here, since `PHYSFSX_fgets` terminates lines when it sees those.  I doubt users will have `\v` or `\f` in their files either.

Mako88: I included those just because those are all considered "whitespace". I agree that " \t" should be enough though.

vLKp:
The change looks good.  I feel bad about not having written it myself back when you raised the issue.  At the time, I put off working on it because I wanted the option to have a command-line specified `-ini` **override** the default ini file, rather than supplement it as you did here.  Implementing that would have required two passes over the command line, so I never wrote it.

I almost merged this, but then noticed an oddity with your authorship.  Both commits show:
```
Author: BuildTools <afuturepilotis@gmail.com>
```
Do you want to correct this before I merge it?  I do not care since you have the right e-mail address listed.

Mako88:
It's ok, you can go ahead and merge it.

PR #196, opened 2016-05-29
fix -Wshadow warnings in get_chunks_state

derhass:
This fixes a warning when building with `WORDS_NEEDS_ALIGNMENT`.


vLKp:
I pushed a different fix, renaming the parameter instead of the member.  This reduces number of lines changed.  I did force a build with `WORDS_NEED_ALIGNMENT` to verify that my version resolved the warning you encountered.

PR #225, opened 2016-09-27
Use no code optimisation for the Debugger build

kreatordxx:
Use no code optimisation for the Debugger build, so Xcode's debugger (others?) can debug properly.

After making this change to SConstruct Apple's linker complains of missing symbols:
```
Undefined symbols for architecture x86_64:
  "newmenu_item::imenu_specific_type::nm_type", referenced from:
      newmenu_item::input_or_menu() in .d1x-rebirth.newmenu.o
  "newmenu_item::input_specific_type::nm_type", referenced from:
      newmenu_item::input_or_menu() in .d1x-rebirth.newmenu.o
  "newmenu_item::number_specific_type::nm_type", referenced from:
      newmenu_item::number_or_slider() in .d1x-rebirth.newmenu.o
  "newmenu_item::slider_specific_type::nm_type", referenced from:
      newmenu_item::number_or_slider() in .d1x-rebirth.newmenu.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)

```
Therefore, using #define constants instead of integral constants so Clang (others?) doesn't get confused when -O0 optimisation is used.

Making a branch for this in case there are any issues with using no optimisation I was unaware of.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -3379,7 +3379,12 @@ class DXXCommon(LazyObjectConstructor):
 		env = self.env
 		user_settings = self.user_settings
 
-		env.Prepend(CXXFLAGS = ['-g', '-O2'])
+		if user_settings.debug:
+			env.Prepend(CXXFLAGS = ['-O0'])
+		else:
+			env.Prepend(CXXFLAGS = ['-O2'])
+
+		env.Prepend(CXXFLAGS = ['-g'])
 		# Raspberry Pi?
 		if user_settings.raspberrypi:
 			rpi_vc_path = user_settings.rpi_vc_path
--- a/common/main/newmenu.h
+++ b/common/main/newmenu.h
@@ -135,12 +135,12 @@ public:
 		return get_union_member(nm_private_slider);
 	}
 	number_slider_common_type *number_or_slider() {
-		return (type == nm_private_number.nm_type || type == nm_private_slider.nm_type)
+		return (type == NM_TYPE_NUMBER || type == NM_TYPE_SLIDER)
 			? &nm_private_number
 			: nullptr;
 	}
 	input_common_type *input_or_menu() {
-		return (type == nm_private_input.nm_type || type == nm_private_imenu.nm_type)
+		return (type == NM_TYPE_INPUT || type == NM_TYPE_INPUT_MENU)
 			? &nm_private_input
 			: nullptr;
 	}


vLKp:
NAK on the change to the optimization level.  If you want no optimizations, that should be set in your `$CXXFLAGS`.  Many of my sanity checks require that the optimizer run so that `__builtin_constant_p` works.  I intentionally used `Prepend` to add `-O2` so that it can be overridden by user `$CXXFLAGS`.  You can pass this on the `scons` command line by doing `scons CXXFLAGS=-O0`.

Since the Xcode project files store `scons` command lines, you can pass the alternate level there, so that it does not affect non-Xcode builds.  Try this patch instead:

```diff
diff --git a/contrib/dxx-rebirth.xcodeproj/project.pbxproj b/contrib/dxx-rebirth.xcodeproj/project.pbxproj
index 6c6181b..6bd6f88 100644
--- a/contrib/dxx-rebirth.xcodeproj/project.pbxproj
+++ b/contrib/dxx-rebirth.xcodeproj/project.pbxproj
@@ -2060,7 +2060,7 @@
 				ONLY_ACTIVE_ARCH = YES;
 				SDKROOT = macosx;
 				dxx_build_dir = "builddir=build/$(PRODUCT_NAME)";
-				dxx_build_type = "debug=True editor=False";
+				dxx_build_type = "debug=True editor=False CXXFLAGS=-O0";
 				dxx_net_opts = "ipv6=False use_udp=True use_tracker=True";
 				scons_args = "$(dxx_build_target) $(dxx_build_renderer) $(dxx_build_type) $(dxx_build_dir) $(dxx_net_opts)";
 			};
@@ -2142,7 +2142,7 @@
 				ONLY_ACTIVE_ARCH = YES;
 				SDKROOT = macosx;
 				dxx_build_dir = "builddir=build/$(PRODUCT_NAME)";
-				dxx_build_type = "debug=True editor=True";
+				dxx_build_type = "debug=True editor=True CXXFLAGS=-O0";
 				dxx_net_opts = "ipv6=False use_udp=True use_tracker=True";
 				scons_args = "$(dxx_build_target) $(dxx_build_renderer) $(dxx_build_type) $(dxx_build_dir) $(dxx_net_opts)";
 			};
```

The symbol problem is a legitimate bug, not an optimizer problem.  The required integral_constant values are never defined anywhere.  When the optimizer runs, it inlines all uses and so the definition is not necessary for a successful link.  The program is technically wrong without a definition.  I will add a definition to resolve the linker error, which also manifests under gcc when optimization is disabled.

kreatordxx:
Done, I committed changes to the Xcode project you recommended. I can confirm it still debugs properly.

PR #229, opened 2016-10-01
Allow window subclasses

kreatordxx:
These changes are to the dcx::window struct and related functions to allow the window struct to be subclassed (see Issue #229). In particular, this would be useful for the editor, which is currently crashing when some dialogs are opened (for example Object->AI Properties dialog).

I made these changes so as not to impact the existing uses of the dcx::window system (at least at this stage).

### DIFF
--- a/common/arch/sdl/window.cpp
+++ b/common/arch/sdl/window.cpp
@@ -25,7 +25,7 @@ namespace dcx {
 static window *FrontWindow = NULL;
 static window *FirstWindow = NULL;
 
-window::window(grs_canvas *src, int x, int y, int w, int h, window_subfunction<void> event_callback, void *data, const void *createdata)
+window::window(grs_canvas *src, int x, int y, int w, int h, window_alloc_type type, window_subfunction<void> event_callback, void *data, const void *createdata)
 {
 	window *prev_front = window_get_front();
 	d_create_event event;
@@ -36,6 +36,7 @@ window::window(grs_canvas *src, int x, int y, int w, int h, window_subfunction<v
 	wind->w_callback = event_callback;
 	wind->w_visible = 1;	// default to visible
 	wind->w_modal =	1;		// default to modal
+	wind->w_alloc = type;
 	wind->w_data = data;
 
 	if (FirstWindow == NULL)
@@ -53,11 +54,30 @@ window::window(grs_canvas *src, int x, int y, int w, int h, window_subfunction<v
 	WINDOW_SEND_EVENT(wind, EVENT_WINDOW_ACTIVATED);
 }
 
+void window::unlink()
+{
+	if (this == FrontWindow)
+		FrontWindow = this->prev;
+	if (this == FirstWindow)
+		FirstWindow = this->next;
+	if (this->next)
+		this->next->prev = this->prev;
+	if (this->prev)
+		this->prev->next = this->next;
+}
+
+window::~window()
+{
+	if (w_alloc == window_alloc_type::subclass)
+		unlink();
+}
+
 int window_close(window *wind)
 {
 	window *prev;
 	d_event event;
 	window_event_result (*w_callback)(window *wind,const d_event &event, void *data) = wind->w_callback;
+	window_alloc_type type = wind->w_alloc;
 
 	if (wind == window_get_front())
 		WINDOW_SEND_EVENT(wind, EVENT_WINDOW_DEACTIVATED);	// Deactivate first
@@ -72,21 +92,19 @@ int window_close(window *wind)
 		return 0;
 	}
 
-	if (wind == FrontWindow)
-		FrontWindow = wind->prev;
-	if (wind == FirstWindow)
-		FirstWindow = wind->next;
-	if (wind->next)
-		wind->next->prev = wind->prev;
-	if (wind->prev)
-		wind->prev->next = wind->next;
+	if (type == window_alloc_type::separate)
+		wind->unlink();		// Remove from the window list
 
 	if ((prev = window_get_front()))
 		WINDOW_SEND_EVENT(prev, EVENT_WINDOW_ACTIVATED);
 
+	// Callback needs to recognise this is a NULL pointer! (And 'wind' will be a dangling pointer for window_alloc_type::subclass)
 	event.type = EVENT_WINDOW_CLOSED;
-	w_callback(wind, event, NULL);	// callback needs to recognise this is a NULL pointer!
-	delete wind;
+	w_callback(wind, event, NULL);
+	
+	if (type == window_alloc_type::separate)
+		delete wind;
+
 	return 1;
 }
 
--- a/common/arch/sdl/window.cpp
+++ b/common/arch/sdl/window.cpp
@@ -22,17 +22,6 @@
 
 namespace dcx {
 
-struct window
-{
-	grs_canvas w_canv;					// the window's canvas to draw to
-	window_event_result (*w_callback)(window *wind,const d_event &event, void *data);	// the event handler
-	int w_visible;						// whether it's visible
-	int w_modal;						// modal = accept all user input exclusively
-	void *data;							// whatever the user wants (eg menu data for 'newmenu' menus)
-	struct window *prev;				// the previous window in the doubly linked list
-	struct window *next;				// the next window in the doubly linked list
-};
-
 static window *FrontWindow = NULL;
 static window *FirstWindow = NULL;
 
@@ -127,16 +116,6 @@ window *window_get_first(void)
 	return FirstWindow;
 }
 
-window *window_get_next(window &wind)
-{
-	return wind.next;
-}
-
-window *window_get_prev(window &wind)
-{
-	return wind.prev;
-}
-
 // Make wind the front window
 void window_select(window &wind)
 {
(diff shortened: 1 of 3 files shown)


common/include/window.h
```diff
@@ -50,6 +50,79 @@ static inline void set_embedded_window_pointer(embed_window_pointer_t *wp, windo
...
+	{
+		return wind.w_canv;
+	}
+	
+	friend window *window_set_visible(window *wind, int visible)
+	{
+		return window_set_visible(*wind, visible);
+	}
+
+	friend int window_is_visible(window *wind)
```
vLKp: Passing a valid `dcx::window` is not optional for `window_is_visible`, so I would prefer to retain the signature `int (dcx::window &)` to enforce this.  A transitional wrapper that permits a pointer and chains to the reference form is fine, but this commit seems to have only the pointer form and to adjust callers that previously called the reference form.

common/include/fwd-window.h
```diff
@@ -36,18 +36,18 @@ int window_close(window *wind);
 int window_exists(window *wind);
 window *window_get_front();
 window *window_get_first();
-window *window_get_next(window &wind);
-window *window_get_prev(window &wind);
+static inline window *window_get_next(window &wind);
```
vLKp: `fwd-window.h` is intended only for forward declarations.  A `static inline` function must be defined in any translation unit where it is used, so the only reason to declare it would be to support calling it from before the definition is seen.  For this commit, I think it would be better to omit these declarations and rely on the definition in `window.h` to provide the declaration implicitly.

common/include/window.h
```diff
@@ -50,6 +50,79 @@ static inline void set_embedded_window_pointer(embed_window_pointer_t *wp, windo
 
 static inline void set_embedded_window_pointer(ignore_window_pointer_t *, window *) {}
 
+struct window
```
vLKp: You could use *class-key* `class` to introduce the type and set the default member access to `private`, which would remove the need to specify `private:` two lines down.  You can leave this as-is if you like.  It runs the same either way.

common/include/window.h
```diff
@@ -50,6 +50,79 @@ static inline void set_embedded_window_pointer(embed_window_pointer_t *wp, windo
...
+{
+private:
+	grs_canvas w_canv;					// the window's canvas to draw to
+	window_event_result (*w_callback)(window *wind,const d_event &event, void *data);	// the event handler
```
vLKp: This could be written as `window_subfunction<void> w_callback;` to avoid repeating the signature.

common/include/window.h
```diff
@@ -57,14 +57,20 @@ struct window
...
 	struct window *next;				// the next window in the doubly linked list
 	
 public:
+	// For creating the window, there are two ways - using the (older) window_create function
+	// or using the constructor, passing an event handler that takes a subclass of window.
+	explicit window(grs_canvas *src, int x, int y, int w, int h, window_subfunction<void> event_callback, void *data, const void *createdata);
+
+	template <typename T>
+			window(grs_canvas *src, int x, int y, int w, int h, window_subclass_subfunction<T> event_callback) :
+			window(src, x, y, w, h, reinterpret_cast<window_subclass_subfunction<window>>(event_callback), NULL, NULL) {}
```
vLKp: Please use `nullptr` instead of `NULL` here.

common/arch/sdl/window.cpp
```diff
@@ -72,21 +92,19 @@ int window_close(window *wind)
...
-	if (wind == FrontWindow)
-		FrontWindow = wind->prev;
-	if (wind == FirstWindow)
-		FirstWindow = wind->next;
-	if (wind->next)
-		wind->next->prev = wind->prev;
-	if (wind->prev)
-		wind->prev->next = wind->next;
+	if (type == window_alloc_type::separate)
+		wind->unlink();		// Remove from the window list
```
vLKp: Why do freestanding window instances need one form of unlink, but embedded instances get a different form?  Would it not be simpler to say that `window_close` will always unlink the window and that deletion must always be performed by a function which knows the most-derived type?

The current design looks like it would be easy for someone to do the wrong thing by allocating a window with the type set to `subclass`, then calling `window_close` on it and not explicitly deleting the allocated window.  This could be surprising, since `window_close` will implicitly delete some windows, but not others.

common/arch/sdl/window.cpp
```diff
@@ -72,21 +92,19 @@ int window_close(window *wind)
...
 
 	if ((prev = window_get_front()))
 		WINDOW_SEND_EVENT(prev, EVENT_WINDOW_ACTIVATED);
 
+	// Callback needs to recognise this is a NULL pointer! (And 'wind' will be a dangling pointer for window_alloc_type::subclass)
```
vLKp: [Dangling pointer](https://en.wikipedia.org/wiki/Dangling_pointer) typically refers to the case that the pointer is uninitialized or points to memory which has been freed.  As I read the code, `wind` is not dangling when `w_callback` executes, but may be dangling by the end of the function.

common/include/window.h
kreatordxx: I was going to do that on my first attempt but the number of files touched would have roughly doubled, since there are other forward declarations as struct (e.g. in mouse.h). I might do that in a later commit, since it also clarifies what dcx::window is being used as, or will be used as.

common/include/window.h
kreatordxx: Done.

common/include/window.h
kreatordxx: OK, yes I googled the reference form and I see it has the benefit (where intended) to disallow nullptr. However, where it uses the pointer form even chaining to the reference form would simply cause it to attempt to dereference nullptr where it has been passed to the pointer form (correct me if I'm wrong). Rather than changing all the uses of the pointer form later on, I would like to declare this and the other related functions as class methods. Therefore, I'm proposing to declare those class methods here.

common/include/fwd-window.h
kreatordxx: OK, I was unsure if the compiler would still inline them. Removing the `static inline` keywords but leaving the function declarations (except for `window_is_visible`, which has been replaced with the `is_visible` method), otherwise I found the friend functions are declared within the `dcx::window` namespace.

common/arch/sdl/window.cpp
kreatordxx: Yes, this was a very tricky one to resolve. The reason I did it this way is because if the handler of the `EVENT_WINDOW_CLOSE` deletes a subclassed window, it deletes the window with it. Attempting to unlink it after it's freed would cause a bad memory access. It can only know whether to unlink it if the handler returns `ignored`.

Yes, it would be a lot simpler for the handler to always delete it. I was avoiding this at first, but now doing it this way to avoid the potential issue you described.

common/arch/sdl/window.cpp
kreatordxx: `wind` would be dangling if `w_callback` frees it, which would have been the case if `wind` were subclassed (but will now always be the case - so changing to `nullptr`).

kreatordxx:
Closing and pull requesting new #230 Allow window subclasses 2.

common/include/window.h
vLKp: I see only one forward declaration as `struct window`, in `fwd-window.h`:
```
$ git grep 'struct window;'
common/include/fwd-window.h:21:struct window;
```
The forward-declaration header is included by some other files, such as `mouse.h`, but the declaration itself is not repeated.
```
$ git grep -F '"fwd-window.h"'
common/include/editor/editor.h:35:#include "fwd-window.h"
common/include/editor/info.h:29:#include "fwd-window.h"
common/include/mouse.h:20:#include "fwd-window.h"
common/include/window.h:23:#include "fwd-window.h"
common/main/automap.h:53:#include "fwd-window.h"
common/main/game.h:37:#include "fwd-window.h"
common/main/newmenu.h:39:#include "fwd-window.h"
```

common/include/window.h
vLKp: It appears you did change this in the second commit in the series.

common/include/window.h
vLKp: Yes, it is enforced only at compile time.  In some cases, a static analysis tool might warn if it discovers a code path where `nullptr` will be dereferenced and passed as a reference, but that is not required of a conforming compiler, nor is it possible in all cases.

I agree that this should not attempt to update all the pointer form calls at once.  Rather, I want to allow the existing reference form calls to retain that form, with a goal of eventually converting other places to use the reference form.  For now, having both forms and allowing the pointer form not to be checked at compile time is fine.

common/arch/sdl/window.cpp
vLKp: I misunderstood the point of the comment.  I interpreted it as a warning that `wind` might be a dangling pointer at the start of `w_callback`, and so some callbacks would need to ignore it or at least avoid dereferencing it.  However, your intent was that `wind` was guaranteed valid when `w_callback` begins, but might, depending on the callback, be invalid when `w_callback` returns.

vLKp:
I think Github allows you to replace one commit series with another for a review, so that you do not need to close and recreate the pull request when making changes to address comments.

kreatordxx:
OK I'll look into that next time.

PR #230, opened 2016-10-02
Allow window subclasses 2

kreatordxx:
These changes are to the dcx::window struct and related functions to allow the window struct to be subclassed (see Issue #229). In particular, this would be useful for the editor, which is currently crashing when some dialogs are opened (for example Object->AI Properties dialog).

Revising according to vLKp's feedback on #229 Allow window subclasses.

### DIFF
--- a/common/arch/sdl/window.cpp
+++ b/common/arch/sdl/window.cpp
@@ -22,17 +22,6 @@
 
 namespace dcx {
 
-struct window
-{
-	grs_canvas w_canv;					// the window's canvas to draw to
-	window_event_result (*w_callback)(window *wind,const d_event &event, void *data);	// the event handler
-	int w_visible;						// whether it's visible
-	int w_modal;						// modal = accept all user input exclusively
-	void *data;							// whatever the user wants (eg menu data for 'newmenu' menus)
-	struct window *prev;				// the previous window in the doubly linked list
-	struct window *next;				// the next window in the doubly linked list
-};
-
 static window *FrontWindow = NULL;
 static window *FirstWindow = NULL;
 
@@ -127,16 +116,6 @@ window *window_get_first(void)
 	return FirstWindow;
 }
 
-window *window_get_next(window &wind)
-{
-	return wind.next;
-}
-
-window *window_get_prev(window &wind)
-{
-	return wind.prev;
-}
-
 // Make wind the front window
 void window_select(window &wind)
 {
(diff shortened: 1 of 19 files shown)


common/arch/sdl/window.cpp
```diff
@@ -25,18 +25,18 @@ namespace dcx {
 static window *FrontWindow = NULL;
 static window *FirstWindow = NULL;
 
-window *window_create(grs_canvas *src, int x, int y, int w, int h, window_subfunction<void> event_callback, void *data, const void *createdata)
+window::window(grs_canvas *src, int x, int y, int w, int h, window_subfunction<void> event_callback, void *data, const void *createdata)
```
vLKp: This could, and probably should, use a *member initializer list*.  I can convert it to that after you push, to save another round of review.

vLKp:
I am uncomfortable with the idea that every subclass is independently responsible for freeing the window object when `EVENT_WINDOW_CLOSE` occurs, but the only alternative I see is to introduce virtual destructors and move the deletion logic back to where `EVENT_WINDOW_CLOSE` is sent, which has its own problems.

I say go ahead and merge this for now.

common/arch/sdl/window.cpp
kreatordxx: Yes I considered that, but didn't this time around to minimise changes. I would be happy to do this.

similar/main/credits.cpp
```diff
@@ -217,6 +217,8 @@ static window_event_result credits_handler(window *, const d_event &event, credi
 			songs_set_volume(GameCfg.MusicVolume);
 			songs_play_song( SONG_TITLE, 1 );
 			std::default_delete<credits>()(cr);
+			if (wind)
```
kreatordxx: OK, yeah I got a bit confused about some conflicting views in this article [http://stackoverflow.com/questions/6731331/is-it-still-safe-to-delete-nullptr-in-c0x](url) but the overall consensus does seem to be that it's safe. Happy to remove these checks (there were a few).

kreatordxx:
Yes, I was uncomfortable with it as well. Shall I prioritise moving all the uses of dcx::window to subclass the window, or wait until after the 0.60 release?

common/arch/sdl/window.cpp
kreatordxx: Done.

similar/main/credits.cpp
kreatordxx: Redundant checks removed.

vLKp:
If you can convert them without breaking anything, you may as well start on it whenever you have time.  I see no reason to block your work on the release.

PR #233, opened 2016-10-04
Allow UI_DIALOG to be subclassed

kreatordxx:
Allow a `UI_DIALOG` to be subclassed, by adding the constructors and destructor. Doing this before attempting to make `UI_DIALOG` a subclass of window for several reasons, including because it would require its own constructor.

Also replacing the kludge of requiring all window handlers to free the window with a better solution - adding an `event_window_result::deleted`, which would be returned to `window_close` (or potentially have other uses) if the handler deleted the window.

### DIFF
--- a/common/ui/dialog.cpp
+++ b/common/ui/dialog.cpp
@@ -134,7 +134,6 @@ static window_event_result ui_dialog_handler(window *wind,const d_event &event, 
 			ui_gadget_delete_all(dlg);
 			selected_gadget = NULL;
 			delete dlg;
-			delete wind;
 			return window_event_result::ignored;
 		default:
 			return window_event_result::ignored;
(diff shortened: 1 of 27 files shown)


common/ui/dialog.cpp
```diff
@@ -194,6 +193,12 @@ void ui_dialog_set_current_canvas(UI_DIALOG *dlg)
 	gr_set_current_canvas(window_get_canvas(*dlg->wind));
 }
 
+UI_DIALOG::~UI_DIALOG()
+{
+	ui_gadget_delete_all(this);
+	selected_gadget = NULL;
```
vLKp: There is no need to update member variables in the destructor.  They go out of scope when the destructor ends, which in this case happens before anything else runs.

kreatordxx: OK, selected_gadget is a global but yes `ui_gadget_delete_all` does only update the gadget pointers - the gadgets are actually freed by the handler's `userdata` holding `unique_ptr`'s to every gadget.

PR #253, opened 2016-10-28
Replace window_exists function calls with checking for window_event_result::deleted

kreatordxx:
The `window_exists` function was reported to be problematic in that another window could be created with the same pointer value as the one that was deleted, resulting in a false positive. Implement `window_event_result` return value across the entire event system and most callbacks so that a `window_event_result::deleted` can be checked for instead.

In the case of `newmenu_do2` and simpler `newmenu_do`'s, add a `track` method to the `dcx::window` class so a pointed to `bool` can be set to `false` when the window is deleted. This is in case another window is created on top without its own polling loop, so that `window_event_result::deleted` is returned for that window, not the one we're tracking. This is not expected for any other single window polling loop currently, allowing `window_event_result::deleted` to be checked for from `event_process()`.

Should this go in unification/master or a beta branch for the next version?

### DIFF
--- a/common/arch/sdl/event.cpp
+++ b/common/arch/sdl/event.cpp
@@ -116,7 +116,7 @@ int event_init()
 	return 0;
 }
 
-int call_default_handler(const d_event &event)
+window_event_result call_default_handler(const d_event &event)
 {
 	return standard_handler(event);
 }
--- a/common/arch/sdl/event.cpp
+++ b/common/arch/sdl/event.cpp
@@ -27,16 +27,16 @@
 
 namespace dcx {
 
-void event_poll()
+window_event_result event_poll()
 {
 	SDL_Event event;
 	int clean_uniframe=1;
 	window *wind = window_get_front();
-	int idle = 1;
+	window_event_result highest_result(window_event_result::ignored);
 	
 	// If the front window changes, exit this loop, otherwise unintended behavior can occur
 	// like pressing 'Return' really fast at 'Difficulty Level' causing multiple games to be started
-	while ((wind == window_get_front()) && (event = {}, SDL_PollEvent(&event)))
+	while ((highest_result != window_event_result::deleted) && (wind == window_get_front()) && (event = {}, SDL_PollEvent(&event)))
 	{
 		switch(event.type) {
 			case SDL_KEYDOWN:
@@ -44,63 +44,58 @@ void event_poll()
 				if (clean_uniframe)
 					unicode_frame_buffer = {};
 				clean_uniframe=0;
-				key_handler(&event.key);
-				idle = 0;
+				highest_result = std::max(key_handler(&event.key), highest_result);
 				break;
 			case SDL_MOUSEBUTTONDOWN:
 			case SDL_MOUSEBUTTONUP:
-                                if (CGameArg.CtlNoMouse)
-                                        break;
-				mouse_button_handler(&event.button);
-				idle = 0;
+				if (CGameArg.CtlNoMouse)
+					break;
+				highest_result = std::max(mouse_button_handler(&event.button), highest_result);
 				break;
 			case SDL_MOUSEMOTION:
-                                if (CGameArg.CtlNoMouse)
-                                        break;
-				mouse_motion_handler(&event.motion);
-				idle = 0;
+				if (CGameArg.CtlNoMouse)
+					break;
+				highest_result = std::max(mouse_motion_handler(&event.motion), highest_result);
 				break;
 			case SDL_JOYBUTTONDOWN:
 			case SDL_JOYBUTTONUP:
-                                if (CGameArg.CtlNoJoystick)
-                                        break;
-				joy_button_handler(&event.jbutton);
-				idle = 0;
+				if (CGameArg.CtlNoJoystick)
+					break;
+				highest_result = std::max(joy_button_handler(&event.jbutton), highest_result);
 				break;
 			case SDL_JOYAXISMOTION:
-                                if (CGameArg.CtlNoJoystick)
-                                        break;
-				if (joy_axis_handler(&event.jaxis))
-					idle = 0;
+				if (CGameArg.CtlNoJoystick)
+					break;
+				highest_result = std::max(joy_axis_handler(&event.jaxis), highest_result);
 				break;
 			case SDL_JOYHATMOTION:
-                                if (CGameArg.CtlNoJoystick)
-                                        break;
-				joy_hat_handler(&event.jhat);
-				idle = 0;
+				if (CGameArg.CtlNoJoystick)
+					break;
+				highest_result = std::max(joy_hat_handler(&event.jhat), highest_result);
 				break;
 			case SDL_JOYBALLMOTION:
 				break;
 			case SDL_QUIT: {
 				d_event qevent = { EVENT_QUIT };
-				call_default_handler(qevent);
-				idle = 0;
+				highest_result = std::max(call_default_handler(qevent), highest_result);
 			} break;
 		}
 	}
 
-	// Send the idle event if there were no other events
-	if (idle)
+	// Send the idle event if there were no other events (or they were ignored)
+	if (highest_result == window_event_result::ignored)
 	{
 		d_event ievent;
 		
 		ievent.type = EVENT_IDLE;
-		event_send(ievent);
+		highest_result = std::max(event_send(ievent), highest_result);
 	}
 	else
 		event_reset_idle_seconds();
 	
 	mouse_cursor_autohide();
+
+	return highest_result;
 }
 
 void event_flush()
@@ -121,7 +116,7 @@ window_event_result call_default_handler(const d_event &event)
 	return standard_handler(event);
 }
 
-void event_send(const d_event &event)
+window_event_result event_send(const d_event &event)
 {
 	window *wind;
 	window_event_result handled = window_event_result::ignored;
@@ -138,27 +133,32 @@ void event_send(const d_event &event)
 		}
 	
 	if (handled == window_event_result::ignored)
-		call_default_handler(event);
+		return call_default_handler(event);
+
+	return handled;
 }
 
 // Process the first event in queue, sending to the appropriate handler
 // This is the new object-oriented system
 // Uses the old system for now, but this may change
-void event_process(void)
+window_event_result event_process(void)
 {
 	d_event event;
 	window *wind = window_get_front();
+	window_event_result highest_result;
 
 	timer_update();
 
-	event_poll();	// send input events first
+	highest_result = event_poll();	// send input events first
 
 	cmd_queue_process();
 
 	// Doing this prevents problems when a draw event can create a newmenu,
 	// such as some network menus when they report a problem
-	if (window_get_front() != wind)
-		return;
+	// Also checking for window_event_result::deleted in case a window was created
+	// with the same pointer value as the deleted one
+	if ((highest_result == window_event_result::deleted) || (window_get_front() != wind))
+		return highest_result;
 	
 	event.type = EVENT_WINDOW_DRAW;	// then draw all visible windows
 	wind = window_get_first();
@@ -176,9 +176,13 @@ void event_process(void)
 		}
 		else
 			wind = window_get_next(*wind);
+
+		highest_result = std::max(result, highest_result);
 	}
 
 	gr_flip();
+
+	return highest_result;
 }
 
 template <bool activate_focus>
--- a/common/arch/sdl/event.cpp
+++ b/common/arch/sdl/event.cpp
@@ -131,7 +131,7 @@ void event_send(const d_event &event)
 		{
 			handled = window_send_event(*wind, event);
 
-			if (!window_exists(wind)) // break away if necessary: window_send_event() could have closed wind by now
+			if (handled == window_event_result::deleted) // break away if necessary: window_send_event() could have closed wind by now
 				break;
 			if (window_is_modal(*wind))
 				break;
--- a/common/arch/sdl/event.cpp
+++ b/common/arch/sdl/event.cpp
@@ -165,19 +165,20 @@ window_event_result event_process(void)
 	while (wind != NULL)
 	{
 		window *prev = window_get_prev(*wind);
-		window_event_result result(window_event_result::ignored);
 		if (window_is_visible(wind))
-			result = window_send_event(*wind, event);
-		if (result == window_event_result::deleted)
 		{
-			if (!prev) // well there isn't a previous window ...
-				break; // ... just bail out - we've done everything for this frame we can.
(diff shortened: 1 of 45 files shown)


Mako88:
I recommend we go ahead and push this to a beta branch. I talked to zico, and it may be a bit later than the 31st that he's able to finish implementing the new tracker, but I think we should go ahead and make the feature freeze, with just the exception of the tracker.

vLKp:
Several commits add the fragment
```c++
    while(event_process() != window_event_result::deleted) {}
```
Perhaps it would be useful to add a utility method `event_process_all` that performs this loop and returns only when the result is `deleted`, so that this fragment is not open-coded in multiple places.

kreatordxx:
Done - added as an inline function in event.h.

common/include/window.h
```diff
@@ -134,7 +135,14 @@ class window
 	{
 		return wind.prev;
 	}
-	
+
+	bool track(bool *exists)
+	{
+		if (w_exists)
```
vLKp: Since the one caller that uses this does not violate the single-assignment rule, and does not check the return value, it looks like this could be done as an assert instead:
```c++
    void track(bool *e)
    {
        assert(!w_exists);
        w_exists = e;
    }
```
That would make the failure much more obvious in debug builds, and would remove the runtime test in release builds.

common/arch/sdl/window.cpp
```diff
@@ -100,6 +100,10 @@ int window_close(window *wind)
 	return 1;
 }
 
+#if 0
```
vLKp: Why not just delete the function entirely?  It will still be in source control if we ever want the implementation back.

common/include/fwd-event.h
```diff
@@ -15,6 +15,7 @@ struct d_change_event;
 struct d_select_event;
 
 enum event_type : unsigned;
+enum class window_event_result : unsigned;
```
kreatordxx: I tried to declare it as a `uint8_t` but even when I included `pstypes.h`, above or below the `#pragma once`, it still told me that `uint8_t` is an undefined type.

similar/main/net_udp.cpp
```diff
@@ -3820,7 +3820,7 @@ int net_udp_setup_game()
 		Netgame.Tracker = 0;
 #endif
 
-	return i >= 0;
+	return (i >= 0) ? window_event_result::close : window_event_result::handled;
```
kreatordxx: OK, yeah I used parentheses a lot, I thought it would give more clarity but I can leave them out in future.

similar/main/game.cpp
```diff
@@ -1307,12 +1307,15 @@ window_event_result game_handler(window *,const d_event &event, const unused_win
...
+			return window_event_result::ignored;
 			break;
 
 		default:
 			break;
 	}
-	return window_event_result::ignored;
+	
+	// If we deleted the window, tell the event loop
+	return Game_wind ? window_event_result::ignored : window_event_result::deleted;
```
vLKp: That seems like a reasonable choice to reduce churn, but I am still confused as to what this return achieves.  Is it expected that it takes the `deleted` branch because the window already handled a `EVENT_WINDOW_CLOSE` message, set `Game_wind` to null, returned, and now another message has been delivered to that window?  If not, under what circumstances are you expecting to reach that return statement with `Game_wind` set to null?  My original comment was because, when `game_handler` is processing `EVENT_WINDOW_CLOSE`, you added a return in that case, so it will not reach the general purpose return at the bottom.

similar/main/net_udp.cpp
vLKp: I do not find it to be more clear, but it does not bother me.  I was surprised to see that, as our lead clang user, you were adding code that clang warns about. :)

kreatordxx: Ha ha yes clang doesn't seem to warn about that one. ;)

similar/main/game.cpp
kreatordxx: OK, for example there's a call to `window_close(Game_wind)` in `GameProcessFrame`, which occurs when a demo has finished playing. Within that call to `window_close(Game_wind)` it calls `game_handler` to see how to handle `window_close` (via `EVENT_WINDOW_CLOSE`), which simply does some tidying up and sets `Game_wind` to `NULL`, then returns `window_event_result::ignored`. This tells `window_close` it may delete the window itself.

Now comes the tricky part. After it has finished the call to `window_close`, it's still in `GameProcessFrame`. `GameProcessFrame` promptly returns, but it's still responding to an `EVENT_WINDOW_DRAW` in `game_handler`. If we return `event_window_result::ignored` in response to this `EVENT_WINDOW_DRAW`, then `window_send_event` will faithfully return the same to `event_process`, which will then go on to call `window_get_next` on a dangling `wind` pointer (since `Game_wind` has been freed within the handler), which **may** result in a crash (doesn't seem to on my system).

common/include/fwd-event.h
kreatordxx: Aha, I was really confused about this error because I thought the types were defined in `sys/types.h`. Apparently not for C++. Fixed, including `cstdint` at the top of `pstypes.h`.

common/include/joy.h
```diff
@@ -23,6 +23,7 @@
 namespace dcx {
 
 struct d_event;
+enum class window_event_result : unsigned;
```
kreatordxx: Fixed (see comments for common/include/fwd-event.h)

similar/main/gamecntl.cpp
```diff
@@ -1937,6 +1937,8 @@ window_event_result ReadControls(const d_event &event)
 
 	if (event.type == EVENT_KEY_COMMAND)
 	{
+		window_event_result result;
```
kreatordxx: OK, this is a habit I had since coding for good old Mac OS 9 (which @zicodxx knows all about - endless fixes to accommodate ;) ), which doesn't like variable declarations after lines of code. Since I will no longer be supporting Mac OS 9 (or PowerPC Macs for that matter), I will do it that way.

common/include/joy.h
```diff
@@ -49,19 +50,19 @@ extern int event_joystick_get_button(const d_event &event);
 namespace dcx {
 
 #if DXX_MAX_BUTTONS_PER_JOYSTICK
-extern void joy_button_handler(SDL_JoyButtonEvent *jbe);
+extern window_event_result joy_button_handler(SDL_JoyButtonEvent *jbe);
```
kreatordxx: Fixed. I can confirm that `common/arch/sdl/event.cpp` and `common/arch/sdl/joy.cpp` now compile with `scons max_joysticks=0` and `scons max_axes_per_joystick=0` (tested setting `DXX_MAX_AXES_PER_JOYSTICK`, `DXX_MAX_HATS_PER_JOYSTICK` and `DXX_MAX_BUTTONS_PER_JOYSTICK` to `0` in `dxxsconf.h` also).

However, note that unification/master won't build with `scons max_joysticks=0` or `scons max_axes_per_joystick=0` (filing a separate issue #258).

similar/main/newmenu.cpp
```diff
@@ -1663,7 +1663,7 @@ struct listbox : embed_window_pointer_t
 	const char **item;
 	int allow_abort_flag;
 	unsigned marquee_maxchars;
-	int (*listbox_callback)(listbox *lb,const d_event &event, void *userdata);
+	window_event_result (*listbox_callback)(listbox *lb,const d_event &event, void *userdata);
```
kreatordxx: Done.

common/ui/checkbox.cpp
```diff
@@ -138,9 +138,13 @@ window_event_result ui_checkbox_do( UI_DIALOG *dlg, UI_GADGET_CHECKBOX * checkbo
 		
 	if (checkbox->pressed == 1)
 	{
+		window_event_result rval;
+
 		checkbox->flag ^= 1;
-		ui_gadget_send_event(dlg, EVENT_UI_GADGET_PRESSED, checkbox);
-		return window_event_result::handled;
+		rval = ui_gadget_send_event(dlg, EVENT_UI_GADGET_PRESSED, checkbox);
```
kreatordxx: Done.

common/arch/sdl/event.cpp
```diff
@@ -165,9 +165,10 @@ void event_process(void)
 	while (wind != NULL)
 	{
 		window *prev = window_get_prev(*wind);
+		window_event_result result(window_event_result::ignored);
 		if (window_is_visible(wind))
-			window_send_event(*wind, event);
-		if (!window_exists(wind))
+			result = window_send_event(*wind, event);
+		if (result == window_event_result::deleted)
```
kreatordxx: Done.

common/arch/sdl/event.cpp
```diff
@@ -138,27 +133,32 @@ void event_send(const d_event &event)
...
 
 	cmd_queue_process();
 
 	// Doing this prevents problems when a draw event can create a newmenu,
 	// such as some network menus when they report a problem
-	if (window_get_front() != wind)
-		return;
+	// Also checking for window_event_result::deleted in case a window was created
+	// with the same pointer value as the deleted one
+	if ((highest_result == window_event_result::deleted) || (window_get_front() != wind))
```
kreatordxx: Noted for next time.

common/arch/sdl/window.cpp
kreatordxx: Done.

kreatordxx:
So to clarify ([nitpicking with terminology](https://en.wikipedia.org/wiki/Software_release_life_cycle) here), I could rename this branch to unification/pre-alpha and unification/master would be the frozen beta branch, i.e. for beta testing prior to release.

kreatordxx:
Would this be correct?

common/include/fwd-event.h
```diff
@@ -7,6 +7,8 @@
 
 #pragma once
 
+#include "pstypes.h"
```
vLKp: I would prefer to use the standard header (`<cstdint>`) here, since nothing else from `pstypes.h` is necessary.

common/include/joy.h
```diff
@@ -50,19 +50,28 @@ namespace dcx {
 #if DXX_MAX_BUTTONS_PER_JOYSTICK
 extern window_event_result joy_button_handler(SDL_JoyButtonEvent *jbe);
 #else
-#define joy_button_handler(jbe)	static_cast<void>(static_cast<SDL_JoyButtonEvent *const &>(jbe))
+static inline window_event_result joy_button_handler(void *)
```
vLKp: This should not be `void *`.  I suggest preserving the old macro definition, but with a tweak to produce the return value you need:
```c++
#define joy_button_handler(jbe)	(static_cast<SDL_JoyButtonEvent *const &>(jbe), window_event_result::ignored)
```

common/include/joy.h
```diff
@@ -50,19 +50,28 @@ namespace dcx {
...
+{
+	return window_event_result::ignored;
+}
 #endif
 
 #if DXX_MAX_HATS_PER_JOYSTICK
 extern window_event_result joy_hat_handler(SDL_JoyHatEvent *jhe);
 #else
-#define joy_hat_handler(jhe)	static_cast<void>(static_cast<SDL_JoyHatEvent *const &>(jhe))
+static inline window_event_result joy_hat_handler(void *)
```
vLKp: Same comment as for `joy_button_handler`.

common/include/joy.h
```diff
@@ -50,19 +50,28 @@ namespace dcx {
...
+{
+	return window_event_result::ignored;
+}
 #endif
 
 #if DXX_MAX_AXES_PER_JOYSTICK
 extern window_event_result joy_axis_handler(SDL_JoyAxisEvent *jae);
 #else
-#define joy_axis_handler(jae)	(static_cast<void>(static_cast<SDL_JoyAxisEvent *const &>(jae)), 1)
+static inline window_event_result joy_axis_handler(void *)
```
vLKp: Same comment as for `joy_button_handler`.

Mako88:
The way me and @zicodxx wanted to do it was to freeze unification/master (except for bugs), and have any features or other code changes pushed to a unification/alpha (or pre-alpha, or whatever you'd like to call it). Then after a release is made, the changes from unification/alpha would be merged with unification/master and development would continue on unification/master until the next freeze (at which point, features would once again shift to unification/alpha).

So to answer your question @kreatordxx, renaming this branch to unification/pre-alpha sounds good (as far as I'm concerned)!

vLKp:
I have intended for a while to deprecate the name `unification/master` and move to using simply `master`, for consistency with most other Git-based projects.  This could be a good time to do that.  Leave `unification/master` behind as a frozen branch for the 0.60 release and develop 0.61 material on `master`.  We would need to follow a different pattern when 0.61 freezes so that we could keep `master` as the development branch long term, but the enforced minimal activity during the 0.60 freeze seems like a good time to make this switch.

Mako88:
I'm definitely onboard with switching to `master` instead of `unification/master`. So @kreatordxx's branch could be renamed to `master` and development continue there (merging changes from `unification/master` after the release)?

Going forward from there, would you be agreeable to having a `pre-alpha` branch that is used whenever `master` is frozen? So `pre-alpha` would be created/rebased on `master` whenever it is frozen, then changes merged from `pre-alpha` back into `master` after a release?

vLKp:
I would rather see `unification/master` branched to `master`, then kreator's branch merged atop it, to maintain commit ancestry consistency.

I dislike the name `pre-alpha`.  I am ambivalent about the approach.  While it has a certain similarity to the kernel development model with Linus' merge window, I am inclined toward a model more like how the handling of stable kernels is done.  Branch a snapshot to `rc-v`*N.M*, let `master` continue running for major development, and cherry-pick selected bug fixes back into the rc branch.  When the release is ready, tag the tip of the rc branch and delete it (since, once released, it is no longer merely a candidate).  Alternately, if you expect to make more releases off it, name it accordingly (e.g. `stable-`*N.M* instead of `rc-v`*N.M*), tag specific releases, and defer deletion until the branch is so old that it is unsupported.

Linus' model requires discipline, either from the administrator freezing the branch or from communication to all contributors, to prevent unwanted changes from slipping into the release branch.  Additionally, it is disruptive to contributors, since whenever a release is imminent (and v0.60 has been imminent for more than a year now), everyone must switch to pushing to a different branch, then switch back after the release is posted.  It is simpler to tell everyone that changes always go into `master`, and may, depending on need, be cherry-picked *by the release maintainer* into a release branch.

common/include/fwd-event.h
kreatordxx: Done.

kreatordxx:
I agree with vLKp. Merged in **unification/master**, made new branch **master** from **no_window_exists**, closing this pull request and deleting **no_window_exists**.

I will leave it to vLKp to make **master** the default branch if/when appropriate (I think only the maintainer can do this anyway).

PR #256, opened 2016-10-30
fix a `-Wunused-parameter` warning in the DXX_USE_OGLES code path

derhass:
The `texanis` parameter is not used in OpenGL ES, but in desktop GL. I don't know what your preferred idiom for silencing this kind of warning is. You might not want to directly merge this commit, but use something else instead. However, the classical cast-to-void works well in practice.

### DIFF
--- a/similar/arch/ogl/ogl.cpp
+++ b/similar/arch/ogl/ogl.cpp
@@ -1691,7 +1691,9 @@ static int ogl_loadtexture (const uint8_t *data, int dxo, int dyo, ogl_texture &
 		default:
 		case OGL_TEXFILT_CLASSIC: // Classic - Nearest
 			gl_mag_filter_int = GL_NEAREST;
-#if !DXX_USE_OGLES
+#if DXX_USE_OGLES
+			static_cast<void>(texanis);
+#else
 			if (texanis && ogl_maxanisotropy > 1.0)
 			{
 				// looks nicer if anisotropy is applied.


vLKp:
That idiom is what I would use.  However, before merging it, I should ask: is silencing the warning the proper fix in this case?  If the warning is because `texanis` has no effect on OpenGL ES systems, is that because the underlying feature is absent?  If so, we should additionally hide from the user the ability to set `texanis`, since setting it has no effect.

derhass:
Yes, in the current GLES backend, anisotropic filter is not implemented, so removing it from the user settings might be a good idea. I should have a look on how widespread the availability of the [EXT_texture_filter_anisotropic](https://www.khronos.org/registry/gles/extensions/EXT/texture_filter_anisotropic.txtl) extension on GLES really is. I'm not sure, but it seems like the RPi is currently the only target which is actually using the GLES code path. And the last time I've checked, it didn't expose that extension.

However, after looking at the code, I've seen that the game already has code for detecting the presence of that extension at runtime (`ogl_get_verinfo`). That code is just not used in the GLES code path, but it could be. So I think the best solution would be to just unify the GLES and GL code paths related to anisotropy...

vLKp:
Unifying those paths sounds like a good idea, but for now, I merged your change to silence the warning.  We can look at replacing it if/when someone unifies the GL/GLES paths.

derhass:
I will probably have a look at this during the next week. I think we can even simplify the desktop GL code paths a bit. Some of the extension handling is done in `ogl_extensions.cpp` (which I introduced with the syncgl patches), but the aniso check predates that, and does it's own extension checking. I think centralizing the GL and GLES extension handling into a single place would be the best option.

PR #260, opened 2016-11-03
Unify OpenGL extension handling between OpenGL and OpenGL ES code paths.

derhass:
Split ogl_get_verinfo() into ogl_tune_for_current() and
ogl_extensions_init(), and consolidate all the OpenGL extension handling
into ogl_extensions.cpp. Unify the code paths for texture anisotropy and
GPU synchronization for OpenGL and OpenGL ES.

Currently, our renderer only uses GLES 1.0, so no real world implementation
will support sync objects for such an old context, but the logic is valid,
and this way, the GLES specific code paths are reduced.

This patch also fixes an issue where the old ogl_get_verinfo() did modify
the texture filtering mode if no anisotropic filter was available. This
was some leftover from the time when the anisotropic filter was a just
a specific CGameConfig.TexFilt mode, and not a separate, orthogonal setting
CGameCfg.TexAnisotropy that it is now.


common/arch/ogl/ogl_extensions.cpp
```diff
@@ -73,16 +97,33 @@ static support_mode is_supported(const char *extensions, const array<long, 2> &v
...
 	if (!version_str) {
 		con_printf(CON_URGENT, "no valid OpenGL context when querying GL extensions!");
 		return;
 	}
 	const auto version = parse_version_str(version_str);
-	const char *extension_str = reinterpret_cast<const char *>(glGetString(GL_EXTENSIONS));
+	const auto extension_str = reinterpret_cast<const char *>(glGetString(GL_EXTENSIONS));
+
+	con_printf(CON_VERBOSE, "OpenGL%s: version %ld.%ld (%s)",
+#if DXX_USE_OGLES
```
vLKp: According to `info cpp` node *Directives Within Macro Arguments*:
```
Occasionally it is convenient to use preprocessor directives within the
arguments of a macro.  The C and C++ standards declare that behavior in
these cases is undefined.
```
`con_printf` is defined as a macro to enable some checking on its usage.  All supported gcc versions handle this as you intended.  clang on Linux handles it.  I cannot test clang on OS X.  Regardless, it would produce slightly simpler runtime code to restructure this as:
```c++
#if DXX_USE_OGLES
#define DXX_OGL_STRING	" ES"
#else
#define DXX_OGL_STRING	""
	con_printf(CON_VERBOSE, "OpenGL" DXX_OGL_STRING ": version %ld.%ld (%s)", version[0], version[1], version_str);
#undef DXX_OGL_STRING
```
This ensures that string pasting, but no preprocessor directives, occur within the macro arguments.

derhass:
OK, I've added your suggests as a second commit. I wasn't even aware that `con_printf` was a macro...

If you prefer all changes in a single commit, I can also update this pull request accordingly. But since each intermediate version will compile just fine, I don't see much of a point for coalescing the commits.

vLKp:
In past releases, `con_printf` was a function.  Now, it is a macro that examines its inputs for certain common misuses and, if no fatal misuses are found, expands to a call to a function also named `con_printf`, passing through all arguments.  See `common/include/console.h` for the macro definition.

Separate commits is fine.  I will merge this as soon as we have a non-frozen branch in which to put it.

vLKp:
I pushed [Merge branch derhass:patches/GLES_GL_extension_unification into master](/dxx-rebirth/dxx-rebirth/pull/260/commits/1cb04cb64b80f2d10276edb3be0619e2524606b6) to branch `master` to merge your work with no further changes.  Due to the timing of the switchover to master, Github does not seem to automatically recognize that this was done.

PR #263, opened 2016-11-07
Fixes freeze when starting new game or editor

kreatordxx:
Fixes freeze introduced by 9511f65a901fbbe7cc3724a99fb742ef3159d177 where `event_process` wouldn't go to the next window if it's hidden.


kreatordxx:
Found a better way...

PR #264, opened 2016-11-07
Fixes freeze when starting new game or editor

kreatordxx:
Fixes freeze introduced by 9511f65a901fbbe7cc3724a99fb742ef3159d177 where `event_process` wouldn't go to the next window if it's hidden.

### DIFF
--- a/common/arch/sdl/event.cpp
+++ b/common/arch/sdl/event.cpp
@@ -161,17 +161,24 @@ window_event_result event_process(void)
 		return highest_result;
 	
 	event.type = EVENT_WINDOW_DRAW;	// then draw all visible windows
-	for (wind = window_get_first(); wind != nullptr; wind = window_get_next(*wind))
+	for (wind = window_get_first(); wind != nullptr;)
 	{
 		if (window_is_visible(wind))
 		{
 			auto prev = window_get_prev(*wind);
 			auto result = window_send_event(*wind, event);
-			if (result == window_event_result::deleted)
-				wind = prev;	// take the previous window and get the next one from that (if prev isn't nullptr)
-
 			highest_result = std::max(result, highest_result);
+			if (result == window_event_result::deleted)
+			{
+				if (!prev)
+				{
+					wind = window_get_first();
+					continue;
+				}
+				wind = prev;	// take the previous window and get the next one from that (if prev isn't nullptr)
+			}
 		}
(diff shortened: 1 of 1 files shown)


common/arch/sdl/event.cpp
```diff
@@ -161,21 +161,14 @@ window_event_result event_process(void)
...
 			auto result = window_send_event(*wind, event);
 			if (result == window_event_result::deleted)
-			{
-				if (!prev) // well there isn't a previous window ...
-					break; // ... just bail out - we've done everything for this frame we can.
-				wind = window_get_next(*prev); // the current window seemed to be closed. so take the next one from the previous which should be able to point to the one after the current closed
-			}
-			else
-				wind = window_get_next(*wind);
+				wind = prev;	// take the previous window and get the next one from that (if prev isn't nullptr)
```
vLKp: I think this is not right.  If `prev` is `nullptr`, then the next steps of the program are:

```
    wind = window_get_next(*nullptr);  // for-loop post step; causes crash
    wind != nullptr; // would end loop if previous step did not crash
```

I think the fix for that is the below, applied on top of your change:

```diff
--- a/common/arch/sdl/event.cpp
+++ b/common/arch/sdl/event.cpp
@@ -161,17 +161,24 @@ window_event_result event_process(void)
 		return highest_result;
 	
 	event.type = EVENT_WINDOW_DRAW;	// then draw all visible windows
-	for (wind = window_get_first(); wind != nullptr; wind = window_get_next(*wind))
+	for (wind = window_get_first(); wind != nullptr;)
 	{
 		if (window_is_visible(wind))
 		{
 			auto prev = window_get_prev(*wind);
 			auto result = window_send_event(*wind, event);
+			highest_result = std::max(result, highest_result);
 			if (result == window_event_result::deleted)
+			{
+				if (!prev)
+				{
+					wind = window_get_first();
+					continue;
+				}
 				wind = prev;	// take the previous window and get the next one from that (if prev isn't nullptr)
-
-			highest_result = std::max(result, highest_result);
+			}
 		}
+		wind = window_get_next(*wind);
 	}
 
 	gr_flip();
```

Moving the for loop post-step to the bottom is ugly, but necessary to let the new `continue` bypass it.

kreatordxx: Done.

I thought it always checks the test before the post-step, but probably not.

PR #266, opened 2016-11-11
General fixes for the Editor

kreatordxx:
Fixes a bunch of bugs that existed with the editor, including:
- Crash when closing a dialog
- Crash when loading Object Properties dialog with no object selected
- File load/save not working
- Crash when loading Do Wall Dialog with no wall selected
- Stability issues after closing wall or hostage dialog
- Crash when loading editor (via delete-E) when playing a demo
- Crash when any of the required files (*.pad, *.fnt and *.mnu) are missing
- Clicking in the game view box always reporting "Click on non-texture ignored"

Note this pull request depends on #264 (as that is a much higher priority).

### DIFF
--- a/common/ui/file.cpp
+++ b/common/ui/file.cpp
@@ -41,7 +41,7 @@ static PHYSFSX_counted_list file_getdirlist(const char *dir)
 {
 	ntstring<PATH_MAX - 1> path;
 	auto dlen = path.copy_if(dir);
-	if ((!dlen && strlen(dir) > 0) || !path.copy_if(dlen, "/"))
+	if ((!dlen && dir[0] != '\0') || !path.copy_if(dlen, "/"))
 		return nullptr;
 	++ dlen;
 	PHYSFSX_counted_list list{PHYSFS_enumerateFiles(dir)};
--- a/common/ui/file.cpp
+++ b/common/ui/file.cpp
@@ -41,7 +41,7 @@ static PHYSFSX_counted_list file_getdirlist(const char *dir)
 {
 	ntstring<PATH_MAX - 1> path;
 	auto dlen = path.copy_if(dir);
-	if (!dlen || !path.copy_if(dlen, "/"))
+	if ((!dlen && strlen(dir) > 0) || !path.copy_if(dlen, "/"))
 		return nullptr;
 	++ dlen;
 	PHYSFSX_counted_list list{PHYSFS_enumerateFiles(dir)};
--- a/common/ui/menubar.cpp
+++ b/common/ui/menubar.cpp
@@ -817,7 +817,6 @@ int menubar_init( const char * file )
 			if (!item.user_function)
 			{
 				con_printf(CON_URGENT, "%s:%u: unknown function \"%s\" in \"%s\"", __FILE__, __LINE__, buf1, file);
-				break;
 			}
 		}
 				
--- a/common/ui/menubar.cpp
+++ b/common/ui/menubar.cpp
@@ -735,7 +735,7 @@ static void ul_xlate(char *s)
 }
 
 
-void menubar_init( const char * file )
+int menubar_init( const char * file )
 {
 	int np;
 	char buf1[200];
@@ -751,8 +751,12 @@ void menubar_init( const char * file )
 			j.Hotkey = -1;
 	auto infile = PHYSFSX_openReadBuffered(file);
 
-	if (!infile) return;
-		
+	if (!infile)
+	{
+		Warning("Couldn't find %s\n", file);
+		return 0;
+	}
+	
 	PHYSFSX_gets_line_t<200> buffer;
 	while ( PHYSFSX_fgets( buffer, infile) != NULL )
 	{
@@ -870,6 +874,8 @@ void menubar_init( const char * file )
 
 	}
 	Menu[0].w = 700;
+
+	return 1;
 }
 
 void menubar_hide()
--- a/common/ui/menubar.cpp
+++ b/common/ui/menubar.cpp
@@ -753,7 +753,7 @@ int menubar_init( const char * file )
 
 	if (!infile)
 	{
-		Warning("Couldn't find %s\n", file);
+		Warning("Could not find %s\n", file);
 		return 0;
 	}
 	
--- a/common/ui/ui.cpp
+++ b/common/ui/ui.cpp
@@ -41,7 +41,7 @@ static int Initialized = 0;
 
 unsigned char CBLACK,CGREY,CWHITE,CBRIGHT,CRED;
 
-grs_font_ptr ui_small_font = nullptr;
+grs_font_ptr ui_small_font;
 
 int ui_init()
 {
@@ -85,8 +85,7 @@ void ui_close()
 		menubar_close();
 		
 		ui_pad_close();
-		if (ui_small_font)
-			ui_small_font.reset();
+		ui_small_font.reset();
 	}
 
 	return;
--- a/common/ui/ui.cpp
+++ b/common/ui/ui.cpp
@@ -33,6 +33,7 @@ COPYRIGHT 1993-1999 PARALLAX SOFTWARE CORPORATION.  ALL RIGHTS RESERVED.
 #include "key.h"
 #include "ui.h"
 #include "mouse.h"
+#include "dxxerror.h"
 
 namespace dcx {
 
@@ -40,17 +41,20 @@ static int Initialized = 0;
 
 unsigned char CBLACK,CGREY,CWHITE,CBRIGHT,CRED;
 
-grs_font_ptr ui_small_font;
+grs_font_ptr ui_small_font = nullptr;
 
-void ui_init()
+int ui_init()
 {
 
-	if (Initialized) return;
-
-	Initialized = 1;
+	if (Initialized) return 1;
 
 	const grs_font *org_font = grd_curcanv->cv_font;
 	ui_small_font = gr_init_font( "pc6x8.fnt" );
+	if (!ui_small_font)
+	{
+		Warning("Could not find pc6x8.fnt");
+		return 0;
+	}
 	grd_curcanv->cv_font =org_font;
 
 	CBLACK = gr_find_closest_color( 1, 1, 1 );
@@ -67,6 +71,9 @@ void ui_init()
 	
 	atexit(ui_close );
 
+	Initialized = 1;
+
+	return 1;
 }
 
 void ui_close()
@@ -78,7 +85,8 @@ void ui_close()
 		menubar_close();
 		
 		ui_pad_close();
-		ui_small_font.reset();
+		if (ui_small_font)
+			ui_small_font.reset();
 	}
 
 	return;
--- a/similar/editor/med.cpp
+++ b/similar/editor/med.cpp
@@ -329,7 +329,7 @@ static void close_editor();
 namespace dsx {
 void init_editor()
 {
-	const char *const pads[] = {
+	static const char pads[][13] = {
 		"segmove.pad",
 		"segsize.pad",
 		"curve.pad",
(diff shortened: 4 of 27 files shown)


similar/editor/medrobot.cpp
```diff
@@ -767,6 +767,10 @@ void object_close_window()
 //-------------------------------------------------------------------------
 int do_object_dialog()
 {
+	if (Cur_object_index == object_none)
+		Cur_object_index = object_first;
+	Assert(Cur_object_index != object_guidebot_cannot_reach);	// not supposed to be this one
```
vLKp: This assert is harmless, but unnecessary.  The call to `vobjptr` will throw if the assert would have failed.

common/ui/file.cpp
```diff
@@ -41,7 +41,7 @@ static PHYSFSX_counted_list file_getdirlist(const char *dir)
 {
 	ntstring<PATH_MAX - 1> path;
 	auto dlen = path.copy_if(dir);
-	if (!dlen || !path.copy_if(dlen, "/"))
+	if ((!dlen && strlen(dir) > 0) || !path.copy_if(dlen, "/"))
```
vLKp: Please do not use `strlen` just to test if a string is empty.  Instead, check whether the first byte is zero.

similar/editor/medwall.cpp
```diff
@@ -410,14 +410,15 @@ window_event_result wall_dialog_handler(UI_DIALOG *dlg,const d_event &event, wal
 	//------------------------------------------------------------
 	ui_button_any_drawn = 0;
 
-	const auto &&w = vwallptr(Cursegp->sides[Curside].wall_num);
+	const auto wn = Cursegp->sides[Curside].wall_num;
```
vLKp: This looks like a good use case for `wallptridx` (not `vwallptridx`).  You can construct a `wallptridx` from an index that is equal to None or is a valid index; any invalid value other than None triggers an exception.  You can treat a `wallptridx` like a pointer in most cases, and it spares you keeping a separate explicit copy of the index in `wn`, because you can also treat it like an index when needed.

common/ui/ui.cpp
```diff
@@ -33,24 +33,28 @@ COPYRIGHT 1993-1999 PARALLAX SOFTWARE CORPORATION.  ALL RIGHTS RESERVED.
...
+#include "dxxerror.h"
 
 namespace dcx {
 
 static int Initialized = 0;
 
 unsigned char CBLACK,CGREY,CWHITE,CBRIGHT,CRED;
 
-grs_font_ptr ui_small_font;
+grs_font_ptr ui_small_font = nullptr;
```
vLKp: This should not be necessary.  The default constructor should initialize it to `nullptr` for you.

common/ui/ui.cpp
```diff
@@ -78,7 +85,8 @@ void ui_close()
 		menubar_close();
 		
 		ui_pad_close();
-		ui_small_font.reset();
+		if (ui_small_font)
```
vLKp: This test is unnecessary.  `std::unique_ptr` (which is used for the implementation of `grs_font_ptr`) already has a test internally to detect `nullptr` and skip freeing it.

similar/editor/med.cpp
```diff
@@ -116,7 +116,7 @@ grs_canvas *Canv_editor_game=&_canv_editor_game; //the game on the editor screen
 
 window *Pad_info;		// Keypad text
 
-grs_font_ptr editor_font;
+grs_font_ptr editor_font = nullptr;
```
vLKp: This should not be necessary.  The default constructor should initialize it to `nullptr` for you.

similar/editor/med.cpp
```diff
@@ -319,38 +319,62 @@ static int padnum=0;
 static void init_editor_screen();
 static void gamestate_restore_check();
 static window_event_result editor_handler(UI_DIALOG *dlg,const d_event &event, unused_ui_userdata_t *data);
+static void close_editor();
 
 namespace dsx {
 void init_editor()
 {
+	const char *const pads[] = {
```
vLKp: Can this be made `static`?  Given the consistency of size, you might get better packing using `static const char pads[][13]` (or `[16]` for quick indexing) than using pointers to anonymous strings.  On x64 systems, the pointers will cost 8 bytes each, so if you waste an average of fewer than 8 bytes per string, storing it in a single contiguous array should be smaller.

common/ui/menubar.cpp
```diff
@@ -751,8 +751,12 @@ void menubar_init( const char * file )
 			j.Hotkey = -1;
 	auto infile = PHYSFSX_openReadBuffered(file);
 
-	if (!infile) return;
-		
+	if (!infile)
+	{
+		Warning("Couldn't find %s\n", file);
```
vLKp: This commit adds several `Warning` calls of this form.  Some of them use `Couldn't`; others use `Could not`.  It would be nice to be consistent.

similar/editor/med.cpp
```diff
@@ -856,7 +880,8 @@ static void close_editor()
 
 	ui_close();
 
-	editor_font.reset();
+	if (editor_font)
```
vLKp: This test is unnecessary.  `std::unique_ptr` (which is used for the implementation of `grs_font_ptr`) already has a test internally to detect `nullptr` and skip freeing it.

similar/editor/med.cpp
```diff
@@ -878,10 +903,13 @@ static void close_editor()
 			break;
 
 		case 3:
-			set_screen_mode(SCREEN_GAME);		//put up game screen
-			Game_mode = GM_EDITOR;
-			editor_reset_stuff_on_level();
-			N_players = 1;
+			if (!Game_wind)	// if we're already playing a game, don't touch!
```
vLKp: Keep this as-is (unless you rebase to undo the hunk entirely), but note for next time that it is often cleaner and reduces churn to use flow control explicitly.  In this case, you could have added

```c++
    if (Game_wind)
        break;
```

immediately before the call to `set_screen_mode` to get the same effect, but avoid reindenting all subsequent lines.

similar/main/gamecntl.cpp
```diff
@@ -1304,7 +1304,10 @@ static window_event_result HandleTestKey(int key)
 		case KEY_E + KEY_DEBUGGED:
 			window_set_visible(Game_wind, 0);	// don't let the game do anything while we set the editor up
 			init_editor();
-			return window_event_result::close;
+			// If editor failed to load, carry on playing
+			if (!EditorWindow)
+				window_set_visible(Game_wind, 1);
+			return EditorWindow ? window_event_result::close : window_event_result::handled;
```
vLKp: The compiler will likely assume that `window_set_visible` could modify `EditorWindow`.  You could improve the generated code by leaving the original `return` as-is and writing the recovery path as:

```c++
    if (!EditorWindow)
    {
        window_set_visible(Game_wind, 1);
        return window_event_result::handled;
    }
```

This would discourage the compiler from rereading `EditorWindow` after `window_set_visible` returns.

kreatordxx:
No worries :) 

I can apply them, no problems.

similar/editor/med.cpp
kreatordxx: Noted. Later on I want to make and use a proper saved game when loading the editor from within a game, so will change it then.

similar/editor/medrobot.cpp
kreatordxx: Removed check.

common/ui/file.cpp
kreatordxx: Done.

similar/editor/medwall.cpp
kreatordxx: Done.

common/ui/ui.cpp
kreatordxx: Removed initialisation.

common/ui/ui.cpp
kreatordxx: Removed check.

similar/editor/med.cpp
kreatordxx: Removed initialisation.

similar/editor/med.cpp
kreatordxx: Done.

similar/editor/med.cpp
kreatordxx: Removed check.

common/ui/menubar.cpp
kreatordxx: Changed **all** occurrences of `Couldn't` to `Could not` (except the one in text.h, which reads from the hog file).

similar/main/gamecntl.cpp
kreatordxx: Done.

similar/main/state.cpp
```diff
@@ -1431,10 +1431,19 @@ int state_restore_all_sub(const char *filename, const secret_restore secret)
 // Read the mission info...
 	PHYSFS_read(fp, mission, sizeof(char) * 9, 1);
 
-	if (!load_mission_by_name( mission ))	{
+	if (!load_mission_by_name( mission )
```
vLKp: This will attempt to `load_mission_by_name` on `new_miss` before it notices that this is a special case.  This could lead to surprising results if someone has a loadable mission with that name.  Is that intended?

vLKp:
I did not intend that you had to normalize **all** uses of *Couldn't* vs *Could not*.  I just wanted consistency in editor warnings. ;)  A global cleanup is fine by me, though.

kreatordxx:
Ha ha yes I know, I felt like doing it completely ;)

For some strange reason GitHub will let me edit your comments...

similar/main/mission.cpp
```diff
@@ -1088,13 +1088,87 @@ int select_mission(int anarchy_mode, const char *message, window_event_result (*
 }
 
 #if DXX_USE_EDITOR
+int write_mission(void)
```
kreatordxx: Defined as `static` for now. Noted, will use `()` in future.

similar/main/mission.cpp
```diff
@@ -1088,13 +1088,87 @@ int select_mission(int anarchy_mode, const char *message, window_event_result (*
...
+{
+	auto &msn_extension =
+#if defined(DXX_BUILD_DESCENT_II)
+	(Current_mission->descent_version == Mission::descent_version_type::descent2) ? MISSION_EXTENSION_DESCENT_II :
```
kreatordxx: Done.

similar/main/mission.cpp
```diff
@@ -1088,13 +1088,87 @@ int select_mission(int anarchy_mode, const char *message, window_event_result (*
...
+	auto &&mfile = PHYSFSX_openWriteBuffered(mission_filename.data());
+	if (!mfile)
+	{
+		PHYSFS_mkdir(MISSION_DIR);	//try making directory - in *write* path
+		mfile = PHYSFSX_openWriteBuffered(mission_filename.data());
+		if (!mfile)
+			return 0;
+	}
+
+#if defined(DXX_BUILD_DESCENT_II)
```
kreatordxx: Done, yes I thought I may as well support any mission type for future use.

similar/main/mission.cpp
```diff
@@ -1088,13 +1088,87 @@ int select_mission(int anarchy_mode, const char *message, window_event_result (*
...
+		for (int i = 0; i < N_secret_levels; i++)
+			PHYSFSX_printf(mfile, "%s,%i\n", static_cast<const char *>(Secret_level_names[i]), Secret_level_table[i]);
+	}
+
+#if defined(DXX_BUILD_DESCENT_II)
+	if (Current_mission->alternate_ham_file)
+		PHYSFSX_printf(mfile, "ham = %s\n", static_cast<const char *>(*Current_mission->alternate_ham_file.get()));
+#endif
+
+	mfile.reset();
```
kreatordxx: Removed.

vLKp: I think in `load_mission` it is explicitly reset because the non-automatic version closed `mfile` before running some other statements.  In that case, I added the explicit reset to maintain the ordering, although it is probably not necessary.  In your commit, it is effectively the last statement, so the reset happens in the same place whether or not it is explicit.

vLKp:
This merge has a trivial conflict due to this pull request including an early version of the fix for freezing on invisible windows, but the version of that fix that was merged to master had additional work done.  The resolution is obvious.  I will fix it up and push this to master soon.

kreatordxx:
Thanks vLKp!

PR #272, opened 2016-11-21
Used saved game loading/saving when switching between editor and a game

kreatordxx:
When going to editor mode via Delete-E, save the game to _Players/gamesave.sge_ (_sge_ short for _saved game editor_). If switching back into the game via **File->Play in 320x200** without saving the level, load that saved game. The saved game can also be loaded back into the editor if the editor was subsequently launched another way (or the level was saved) with **File->Restore Game State**.

This is primarily to provide more continuity when switching between the game and the editor. The last set of changes to the editor made the player's stats reset to starting a new game when choosing **File->Play in 320x200** (as it was calling `StartNewGame`). This also maintains the game state more accurately, preserving weapon fire and various other info, as stored in the saved game file.

May also be useful for testing, especially for finding the object number of any given object, including temporary objects. (And cheating! ;) But not for multiplayer - editor mode is disallowed!!)

### DIFF
--- a/common/include/editor/editor.h
+++ b/common/include/editor/editor.h
@@ -155,6 +155,13 @@ struct editor_view
 	fix ev_zoom;				//zoom for this window
 };
 
+enum class editor_gamestate : uint8_t
+{
+	none,		// editing a level, not a gamestate
+	unsaved,	// just pressed delete-e from a game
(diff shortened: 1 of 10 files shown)


kreatordxx:
Thanks @vLKp!

PR #281, opened 2016-12-05
Call newmenu_free_background at exit

kreatordxx:
Suppresses '1 blocks were left allocated' warning at exit, which is an alert box on non-Linux builds.

Only applicable when built with `debug=1` or `memdebug=1`.

### DIFF
--- a/similar/main/inferno.cpp
+++ b/similar/main/inferno.cpp
@@ -502,6 +502,7 @@ static int main(int argc, char *argv[])
 	show_titles();
 
 	set_screen_mode(SCREEN_MENU);
+	atexit(newmenu_free_background);	// suppresses 'blocks were left allocated' warning at exit
 
 	con_printf( CON_DEBUG, "\nDoing gamedata_init..." );
 	gamedata_init();


vLKp:
NAK for several reasons.

1. `newmenu_free_background` is safe (but useless) to call multiple times, so it could be called directly instead of using an `atexit` hook.
1. The memory is not leaked; it is freed by a global destructor.  There is no need to explicitly free it in non-memdebug builds.
1. This is a bug in the memdebug code.  It incorrectly assumes that memory freed by global destructors is not freed, because memdebug runs the sweep code before global destructors run.  The ideal fix would be to run the memdebug sweep at the right time, but there is no standard-approved way of running code after all global destructors run.  The second-best fix would be never to rely on a global destructor to free allocated resources, but to instead store those resources at more narrow scope.

Does this patch work for you instead?  It avoids the use of `atexit` and only applies when memdebug is enabled.

```diff
--- a/similar/main/inferno.cpp
+++ b/similar/main/inferno.cpp
@@ -502,6 +502,16 @@ static int main(int argc, char *argv[])
 	show_titles();
 
 	set_screen_mode(SCREEN_MENU);
+#ifdef DEBUG_MEMORY_ALLOCATIONS
+	struct hack_free_global_backgrounds
+	{
+		~hack_free_global_backgrounds()
+		{
+			newmenu_free_background();
+		}
+	};
+	hack_free_global_backgrounds hack_free_global_background;
+#endif
 
 	con_printf( CON_DEBUG, "\nDoing gamedata_init..." );
 	gamedata_init();


```

kreatordxx:
Yes, that works fine.

I followed the steps recommended for finding abandoned memory and memory leaks using Apple's Instruments program (Allocations and Leaks tools). The number of blocks not freed according to memdebug remained stable (1 after just playing games and an extra 326 after loading the editor) no matter how many times I played a game or ran the editor before returning to the main menu. Memory usage increased by around 30-50 kB each time with no resolution switches but I assumed this was entirely in libraries after searching some of the reported variables (enormous list). Switching resolutions (to 800x600 and 1024x768 in one iteration) resulted in a whopping 20 MB of memory gain each time, I did a search on this and apparently [SDL is responsible](http://stackoverflow.com/questions/4917661/sdl-memory-leaks-detected-using-valgrind).

I know the memdebug code is a simplistic way of finding memory leaks and this isn't a real memory leak. I just wanted to suppress this warning. In terms of suppressing the warning, memdebug only checks blocks created with `mem_malloc` - not `new`. IMHO I think the memdebug code is worth keeping as a simple memory leak check. I plan to have all the editor memory blocks freed when the editor is closed, since this appears to be the intention of `close_editor`. This would also allow modification of the editor interface files without having to exit.

vLKp:
I pushed a change containing the patch I proposed above, along with a block comment explaining why it is guarded by a `#ifdef`.

kreatordxx:
Thanks! It still works.

vLKp:
I do not find `memdebug` to be useful.  I prefer dedicated tools that can handle global destructors properly.  However, since it is useful to you, we may as well fix the easy problems.  If you see any other reports from memdebug, even bogus ones like this, please fix them or report them, at your preference.

Issue #296, opened 2017-01-08
Crash when Demo Recording Runs Out of Disk Space


kreatordxx:
Steps to reproduce:
1. Create a small disk image (e.g. 5 MB or smaller) and load it
2. Rename 'demos' directory to 'demos2' (or something else) - this is in:
	Same directory as program for Windows
	(your home directory)/Library/Preferences/D?X Rebirth/ for Mac OS X
	(your home directory)/.d?x-rebirth/ for other *NIX
Where ? is the descent version number.
3. Create a symlink (*NIX - Mac OS X aliases won't work) or a shortcut (Windows) in the same place called 'demos', pointing to someplace in the disk image
4. Run D?X-Rebirth
5. Record a demo, playing until it runs out of disk space (took me about 10 minutes for 10 MB - can reduce available space by stuffing with other files to make it happen faster)

The warning dialog comes up saying your demo is probably corrupt, enter a filename and it either crashes (for release build) or fails an assert (for debug build).

I looked into this because there's a call to `newdemo_stop_playback()` in `newdemo_record_start_frame`, in a check for `nd_record_v_no_space`. `newdemo_stop_playback()` closes the game window, which I want to replace with returning `window_event_result::close` as per #227.

PR #297, opened 2017-01-08
Fix crash when ran out of disk space when writing demo

kreatordxx:
Check for `nd_record_v_no_space` in `_newdemo_write`. Fixes issue #296.

### DIFF
--- a/similar/main/newdemo.cpp
+++ b/similar/main/newdemo.cpp
@@ -317,13 +317,16 @@ static int _newdemo_write(const void *buffer, int elsize, int nelem )
 {
 	int num_written, total_size;
 
+	if (unlikely(nd_record_v_no_space))
+		return -1;
+
 	total_size = elsize * nelem;
 	nd_record_v_framebytes_written += total_size;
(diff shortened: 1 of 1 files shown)


similar/main/newdemo.cpp
```diff
@@ -317,13 +317,16 @@ static int _newdemo_write(const void *buffer, int elsize, int nelem )
...
+		return -1;
+
 	total_size = elsize * nelem;
 	nd_record_v_framebytes_written += total_size;
 	Newdemo_num_written += total_size;
 	Assert(outfile);
 	num_written = (PHYSFS_write)(outfile, buffer, elsize, nelem);
 
-	if (num_written == nelem && !nd_record_v_no_space)
+	if (likely(num_written == nelem && !nd_record_v_no_space))
```
vLKp: The addition of the test for `nd_record_v_no_space` above seems to make it unnecessary to retest it here.  This variable is local to `newdemo.cpp`, so it cannot be changed by a call to `PHYSFS_write` (but the compiler cannot know that).

Issue #280, opened 2016-12-04
Crash On Exit While Recording a Demo


Mako88:
Tested in D2 (didn't test D1), if you are recording a demo and then hit abort game the game will freeze (not crash). I sent a sigint and got a backtrace:

```
#0  0x00000000004b644c in std::__detail::_Mod_range_hashing::operator() (
    this=0x7fffffffc390, __den=23, __num=8)
    at /usr/include/c++/6.2.1/bits/hashtable_policy.h:446
#1  std::__detail::_Hash_code_base<unsigned short, std::pair<unsigned short const, render_state_t::per_segment_state_t>, std::__detail::_Select1st, std::hash<unsigned short>, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, false>::_M_bucket_index (this=0x7fffffffc390, __n=23, __c=8)
    at /usr/include/c++/6.2.1/bits/hashtable_policy.h:1171
#2  std::_Hashtable<unsigned short, std::pair<unsigned short const, render_state_t::per_segment_state_t>, std::allocator<std::pair<unsigned short const, render_state_t::per_segment_state_t> >, std::__detail::_Select1st, std::equal_to<unsigned short>, std::hash<unsigned short>, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<false, false, true> >::_M_bucket_index (
    this=0x7fffffffc390, __k=@0x7fffffff5a5a: 8, __c=8)
    at /usr/include/c++/6.2.1/bits/hashtable.h:623
#3  std::__detail::_Map_base<unsigned short, std::pair<unsigned short const, render_state_t::per_segment_state_t>, std::allocator<std::pair<unsigned short const, render_state_t::per_segment_state_t> >, std::__detail::_Select1st, std::equal_to<unsigned short>, std::hash<unsigned short>, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<false, false, true>, true>::operator[] (
    this=0x7fffffffc390, this@entry=0x7fffffffc398, __k=@0x7fffffff5a5a: 8)
---Type <return> to continue, or q <return> to quit---
    at /usr/include/c++/6.2.1/bits/hashtable_policy.h:590
[...]

Issue #227, opened 2016-09-27
Make dcx::window into a class and use C++ features


kreatordxx:
When I originally wrote the d_window struct and corresponding system it was in C, with the intention of making the different views/panes in DXX-Rebirth more object oriented (and IMO cleaner code). Due to the limitations of C this meant maintaining up to 3 pointers (d_window, UI_DIALOG and robot_dialog for example). The use of unique_ptr's makes this potentially even more complicated.  C++ adds features such as subclassing and polymorphism and I would be keen to make use of these features for dcx::window, in particular so that dcx::window is a subclass of grs_canvas; UI_DIALOG, newmenu and others are subclasses of dcx::window and the specific uses are further subclasses. The trigger for this is continuing issues with the editor, which I would like to work on further. That said I also like this sort of 'tidy up'.

I would do this in a separate branch, but I am posting this idea first before I start, to get some feedback.

vLKp:
This sounds plausible, but I am wary of using polymorphism for this.  Compilers implement it correctly, but it has a runtime cost (particularly with regard to branch prediction), so I am cautious about using it without a good reason, especially if it would be on a hot path.  What areas do you think would benefit from polymorphism relative to the current design?

kreatordxx:
The number one thing I would like to achieve is combine the data structures using inheritance, to reduce pointer and heap maintenance. This appears to be where the editor is tripping up currently. I would also like to do this to move some gadget access global variables in the editor (maybe more) into those data structures.

I was also thinking of making all the window manipulation/accessing functions in window.h into class methods (mainly to remove the `window_` prefix).

Another thing I was considering was adding virtual draw and event_handler methods to the window class, since I feel it makes the code cleaner than passing a function as a function parameter. If this causes an unacceptable performance hit then I will stick to the current system for this.

All of the above changes would affect the code broadly (everywhere a dcx::window and all inherited structs like newmenu are used), at least that was my intention.

kreatordxx:
Combining the data structures using inheritance would mean use of C++ constructors and destructors, rather than the explicitly defined and called creator and closing functions.

vLKp:
I think the performance hit would be minimal in areas where the code already uses function pointers.  Virtual function calls typically require one extra dereference, which should be quite cheap compared to the cost of a call by pointer.  Most compilers implement vtable handling such that these two generate roughly equivalent code:
```c++
struct A
{
	virtual void f();
};

struct B
{
	struct VTable {
[...]

PR #298, opened 2017-01-08
Remove calls to window_close(Game_wind)

kreatordxx:
Remove calls to `window_close(Game_wind)` as discussed in #227, replacing with returning `window_event_result::close` all the way back to `game_handler`.

There have been a number of issues picked up relating to these offending calls (e.g. #280), particularly within functions ultimately called by `game_handler`. This is an attempt to remove all such calls.

There are still three outstanding cases, but I thought I would get feedback first before going further.

### DIFF
--- a/similar/main/endlevel.cpp
+++ b/similar/main/endlevel.cpp
@@ -690,8 +690,7 @@ window_event_result do_endlevel_frame()
 					                    CT_NONE,MT_NONE,RT_NONE);
 
 					if (objnum == object_none) { //can't get object, so abort
-						result = std::max(stop_endlevel_sequence(), result);
-						return result;
+						return std::max(stop_endlevel_sequence(), result);
 					}
 
 					Viewer = endlevel_camera = objnum;
--- a/similar/main/endlevel.cpp
+++ b/similar/main/endlevel.cpp
@@ -281,9 +281,8 @@ vms_matrix surface_orient;
 static int endlevel_data_loaded;
 
 namespace dsx {
-void start_endlevel_sequence()
+window_event_result start_endlevel_sequence()
 {
-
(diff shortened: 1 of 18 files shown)


similar/main/endlevel.cpp
```diff
@@ -692,8 +690,8 @@ void do_endlevel_frame()
 					                    CT_NONE,MT_NONE,RT_NONE);
 
 					if (objnum == object_none) { //can't get object, so abort
-						stop_endlevel_sequence();
-						return;
+						result = std::max(stop_endlevel_sequence(), result);
```
vLKp: This could be done without writing to `result` first:
```c++
return std::max(stop_endlevel_sequence(), result);
```

similar/main/switch.cpp
```diff
@@ -293,21 +299,26 @@ int check_trigger_sub(object &plrobj, const trgnum_t trigger_num, int pnum,int s
...
+		if (trigger.flags & TRIGGER_EXIT)
+		{
+			result = start_endlevel_sequence();
+			if (result == window_event_result::handled)
+				result = window_event_result::ignored;	// call multi_send_trigger, or end game anyway
 		}
 
 		if (trigger.flags & TRIGGER_SECRET_EXIT) {
+			if (trigger.flags & TRIGGER_EXIT)
+				LevelError("Trigger %u is both a regular and secret exit! Not recommended", trigger_num);
```
vLKp: This will look a bit odd after string concatenation.  `LevelError` always appends to the text `"  Please report this to the level author, not to the Rebirth maintainers."`, so the bare `Not recommended` will be formatted oddly.  It might look better to say `  This is not a supported combination.` or `This combination is deprecated.`

PR #299, opened 2017-01-10
Remove remaining calls to window_close(Game_wind)

kreatordxx:
Remove calls to `window_close(Game_wind)` as discussed in #227, replacing with returning `window_event_result::close` all the way back to `game_handler`.

There have been a number of issues picked up relating to these offending calls (e.g. #280), particularly within functions ultimately called by `game_handler`.

This removes the three cases that were left. The somewhat contentious commit is the final one 0c5a000328058e18791aabc22d055cd27b6d2f61, as potentially a lot of code can be run between receiving a network event telling us to end the game and the game actually ending, possibly leading to unintended side effects. This is the change I was talking about in #227 @zicodxx.

### DIFF
--- a/similar/main/multi.cpp
+++ b/similar/main/multi.cpp
@@ -983,7 +983,7 @@ void multi_do_protocol_frame(int force, int listen)
 	}
 }
 
-void multi_do_frame(void)
+window_event_result multi_do_frame()
 {
 	static int lasttime=0;
 	static fix64 last_gmode_time = 0, last_inventory_time = 0, last_repo_time = 0;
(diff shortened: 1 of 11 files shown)


similar/main/multi.cpp
```diff
@@ -1041,12 +1041,7 @@ void multi_do_frame(void)
 
 	multi_do_protocol_frame(0, 1);
 
-	if (multi_quit_game)
-	{
-		multi_quit_game = 0;
-		if (Game_wind)
-			window_close(Game_wind);
-	}
+	return multi_quit_game ? window_event_result::close : window_event_result::handled;
```
vLKp: Previously, `multi_quit_game` was reset immediately.  Now, it is not.  Based on the commit message, I think you intended this change, so that `multi_quit_game` remains set until the new game clears it.  Is that correct?

kreatordxx: Yes, I intended this change. This is in case `multi_quit_game` is first caught other than in the call to `multi_do_frame()` in `GameProcessFrame()`.

kreatordxx:
I did a quick test with two instances of D2X-Rebirth and spotted issue #303, but this occurs before these changes. I didn't find any game usage differences before and after these changes, as intended.

PR #304, opened 2017-01-12
Fix Clang warnings

kreatordxx:
Fix three warnings when compiling using Clang (e.g. Apple's Clang).

### DIFF
--- a/similar/main/multi.cpp
+++ b/similar/main/multi.cpp
@@ -303,7 +303,7 @@ void powerup_cap_state::cap_secondary_ammo(powerup_type_t type, uint8_t &player_
 		return;
 	const auto capped = static_cast<uint8_t>(current_in_mine < maximum_allowed ? maximum_allowed - current_in_mine : 0);
 	player_ammo = capped;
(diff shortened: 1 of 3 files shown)


similar/main/multi.cpp
```diff
@@ -303,7 +303,7 @@ void powerup_cap_state::cap_secondary_ammo(powerup_type_t type, uint8_t &player_
 		return;
 	const auto capped = static_cast<uint8_t>(current_in_mine < maximum_allowed ? maximum_allowed - current_in_mine : 0);
 	player_ammo = capped;
-	con_printf(CON_VERBOSE, "Capping secondary %u due to powerup cap: current=%u max=%u was=%u now=%hu", idx, get_current(idx), get_max(idx), static_cast<unsigned>(current_on_player), capped);
+	con_printf(CON_VERBOSE, "Capping secondary %u due to powerup cap: current=%u max=%u was=%u now=%hc", idx, get_current(idx), get_max(idx), static_cast<unsigned>(current_on_player), capped);
```
vLKp: This is the wrong fix.  `%c` prints the supplied data as a character, not as a very short number.  For example, with your change, a `capped` value of 8 will try to render a tab character, rather than print the number 8.

The correct fix is either `%hhu` or, if that is not portable, to use `%u` and cast the supplied value to an `unsigned`.

PR #305, opened 2017-01-12
Fix Clang warnings

kreatordxx:
Fix 3 warnings when compiling with Clang, e.g. Apple's Clang.

### DIFF
--- a/common/arch/ogl/ogl_sync.cpp
+++ b/common/arch/ogl/ogl_sync.cpp
@@ -33,9 +33,9 @@ ogl_sync::~ogl_sync()
 		con_printf(CON_URGENT, "fence sync object was never destroyed!");
 }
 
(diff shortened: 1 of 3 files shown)


vLKp:
Do you normally build test Descent 1?  There is currently a warning in the gcc builds for Descent 1 due to a recent change in your `window_event_result` series.  I was a bit surprised to see that this series does not address that, since it is fatal in `-Werror` builds.  I fixed it just now in [Fix D1 build](/dxx-rebirth/dxx-rebirth/commit/5201218e20293472224df5dc59a99d41b4e43b72).

kreatordxx:
I was occasionally testing D1, I'll do it more often.

vLKp:
This looks good.  Merged.

Issue #303, opened 2017-01-11
Debugger breakpoint hit when client dropped from a multiplayer game


kreatordxx:
Steps to reproduce (**master**):
1. Launch one instance of D2X-Rebirth
2. Host a multiplayer game (choose defaults)
3. Launch a second instance of D2X-Rebirth (e.g. launch with debugger in Xcode)
4. Manually join multiplayer game using 42425 or (some other number) for 'my port'
5. Host starts game
6. Host kicks other player (press F8, type "/kick: <player name>" and hit enter)
7. Second instance clears the 'you got kicked' message

Immediately I get the debugger breaking at this line multiple times:
`draw_weapon_info_sub(player_info, info_index, box, pic_x, pic_y, PRIMARY_WEAPON_NAMES_SHORT(weapon_num), text_x, text_y, multires_gauge_graphic);`
in similar/main/gauges.cpp.

The value of `weapon_num` is the poisoned value of `0xfd`.

Exactly the same thing happens if the second player is dropped due to timeout.

Note this occurs before commit 0c5a000328058e18791aabc22d055cd27b6d2f61 (not merged into **master** at time of posting).

vLKp:
The debugger trap at step 7 is probably the `d_debugbreak` triggered by trying to name a weapon with an invalid type code.  I suggest printing the entire structure of the affected player when this happens.  I expect you will find that parts of it are garbage bytes.

kreatordxx:
Whoops, it looks like this is actually two bugs, with one of them much simpler than I first thought. Rewriting the first description and posting a second bug report.

Issue #306, opened 2017-01-13
Failed Assert when non-host player in a multiplayer game hits a wall


kreatordxx:
Steps to reproduce (**master**):

1. Launch one instance of D2X-Rebirth
2. Host a multiplayer game (choose defaults) and start it
3. Launch a second instance of D2X-Rebirth (e.g. launch with debugger in Xcode)
4. Manually join multiplayer game using 42425 or (some other number) for 'my port'
5. Second instance: fly into a wall

It immediately fails this assert:
`Assert(!(obj->mtype.phys_info.flags & PF_STICK && obj->mtype.phys_info.flags & PF_BOUNCE));	//can't be bounce and stick`

Note this occurs before commit 0c5a000328058e18791aabc22d055cd27b6d2f61, where I changed the handling of when a network event causes the game to abort.

kreatordxx:
I'm going to keep investigating this to see if I can find the root cause. Commit e76c8cf264d5cc105434e85bddab459511df1c70 fixes it per se, but there might be something else going on. I just found out `ConsoleObject` is set in `multi_prep_level_player()`, so had been setting a watchpoint on the wrong memory address (setting a watchpoint on an expression doesn't seem to be working for me).

kreatordxx:
OK, it looks like I've found the root cause. This sequence happens:
1. Host starts the game
2. Host sends a packet on all the start positions. Because no-one has joined yet, these are all `OBJ_GHOST` (except for the host's ship).
3. Client receives a packet from the host while the client is still in the menus on the player positions
4. Because the client's start position has been sent as `OBJ_GHOST`, it reads it as one, NOT filling out the physics info (because ghosts don't move) and leaving it as poisoned memory.
5. The client calls `multi_reset_player_object` in `multi_prep_level_player()`, which fails to clear the poisoned memory as it uses the `|=` operator instead of `=` when setting `objp->mtype.phys_info.flags`. **This is risky.**
6. The client sets a start position as `ConsoleObject` in `multi_prep_level_player()`.
7. The client calls `reset_player_object()` multiple times, but every call fails to initialise `ConsoleObject->mtype.phys_info.flags` for the same reason as step 5.
8. Whenever the client hits a wall, the assert is failed.

Therefore, sending another commit to address the shortcoming of `multi_reset_player_object`.

PR #307, opened 2017-01-13
Fix debugger break at draw_weapon_info_sub when multi player dumped

kreatordxx:
Fixes bug #303.

Also happened when a client to a multiplayer game dropped out due to some network error. Delay call of `multi_leave_game()` until responding to `EVENT_WINDOW_CLOSE`, so the game isn't in an unstable state between handling the network event and the game closing.

### DIFF
--- a/similar/main/game.cpp
+++ b/similar/main/game.cpp
@@ -1274,7 +1274,6 @@ window_event_result game_handler(window *,const d_event &event, const unused_win
 			break;
 
 		case EVENT_WINDOW_CLOSE:
-			Game_wind = nullptr;
 			digi_stop_digi_sounds();
 
 			if ( (Newdemo_state == ND_STATE_RECORDING) || (Newdemo_state == ND_STATE_PAUSED) )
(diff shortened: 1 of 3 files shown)


vLKp:
Some of the early commits pushed for this branch look questionable, but the latest iteration looks fine.  I do not understand from the diffs alone why it works, but I see nothing that worries me.  I have no objection to merging it.

kreatordxx:
Yes, the earlier ones unintentionally included some stuff that was meant to be temporary as I was debugging for #306.

The idea is that rather than calling `multi_leave_game()` when the network event occurs, all of the closing functions are called in response to `EVENT_WINDOW_CLOSE`, so the game isn't in an unstable state between receiving the event and actually closing. I'm not sure about the specifics, but `multi_leave_game()` must reset the player struct, where the player's weapon number variable is poisoned. This was a semi-educated 'guess and check' and I got lucky :D Not so with #306 (yet anyway)...

PR #308, opened 2017-01-15
Fix bug #306 where a client in a multi game hitting a wall fails an assert

kreatordxx:
Initialise ConsoleObject->mtype.phys_info.flags in reset_player_object() instead of adding flags. The only place this was initialised properly was when reading the player object from disk.

### DIFF
--- a/similar/main/multi.cpp
+++ b/similar/main/multi.cpp
@@ -2415,7 +2415,7 @@ void multi_reset_player_object(const vobjptr_t objp)
 	objp->mtype.phys_info.mass = Player_ship->mass;
 	objp->mtype.phys_info.drag = Player_ship->drag;
 	if (objp->type == OBJ_PLAYER)
-		objp->mtype.phys_info.flags |= PF_TURNROLL | PF_WIGGLE;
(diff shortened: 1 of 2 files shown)


vLKp:
This looks like a reasonable workaround, but I am curious why the flags end up incorrect in the first place.  If they are correct on disk, but wrong in memory, that says they were overwritten somewhere else with garbage data or that there is a path where loading players from disk skips setting an initial value.

kreatordxx:
I strongly suspect the only reason it worked before is because the structs were initialised with zero bytes via `memset`. I see in the editor that when a player start is created, `place_object` calls `obj_create`, which properly initialises `obj->mtype.phys_info` if the object's `movement_type == MT_PHYSICS`. In the case of a multiplayer game, the client actually receives the player ship struct from the host, which resets the whole player struct but never initialises `obj->mtype.phys_info.flags` properly, instead using the or operator `|=` in `reset_player_object()`. When object packet receiving was initialised with poisoned bytes `0xfd`, it revealed this vulnerability.

kreatordxx:
See bug #306 for an update - found the root cause.

Issue #293, opened 2016-12-30
D2X Sidearm bot does not drop children in debug build. 


zicodxx:
Another issue caused by uninitialized variables as I can see. The Sidearm bot 
![sidearm](https://cloud.githubusercontent.com/assets/3537029/21565887/53536a04-ce9d-11e6-92b5-e7d05a9c9592.jpg)
should normally drop a few children bots. But in the debug build it doesn't. Compiling with poison=valgrind, it spews quite a few warnings - not all related to this issue (I'll try to hunt them down later) - but the one related to the issue seems to be that one:

```
==20156==  Uninitialised value was created by a client request
==20156==    at 0x4A7188: DXX_MAKE_MEM_UNDEFINED<unsigned char> (poison.h:12)
==20156==    by 0x4A7188: DXX_POISON_MEMORY<unsigned char*, unsigned char> (poison.h:53)
==20156==    by 0x4A7188: DXX_POISON_VAR<d2x::object> (poison.h:66)
==20156==    by 0x4A7188: d2x::obj_create(dcx::object_type_t, unsigned char, valptridx<d2x::segment>::basic_ptridx<valptridx<d2x::segment>::vm>, dcx::vms_vector const&, dcx::vms_matrix const*, int, unsigned char, unsigned char, unsigned char) (object.cpp:1120)
==20156==    by 0x444360: d2x::drop_powerup(int, int, int, dcx::vms_vector const&, dcx::vms_vector const&, valptridx<d2x::segment>::basic_ptridx<valptridx<d2x::segment>::vm>, bool) (fireball.cpp:934)
==20156==    by 0x444C50: d2x::object_create_robot_egg(int, int, int, dcx::vms_vector const&, dcx::vms_vector const&, valptridx<d2x::segment>::basic_ptridx<valptridx<d2x::segment>::vm>) (fireball.cpp:1033)
==20156==    by 0x444DC8: d2x::object_create_robot_egg(valptridx<d2x::object>::basic_ptr<valptridx<d2x::object>::vm, 0u>) (fireball.cpp:1051)
==20156==    by 0x44570D: d2x::do_explosion_sequence(valptridx<d2x::object>::basic_ptr<valptridx<d2x::object>::vm, 0u>) (fireball.cpp:1234)
==20156==    by 0x4AA09F: object_move_one (object.cpp:1684)
==20156==    by 0x4AA09F: d2x::object_move_all() (object.cpp:1881)
[...]

PR #311, opened 2017-01-16
Fix bug #293, where robots dropped by robots would be ghost-like then…

kreatordxx:
… explode.

Initialise the ai info for spawned robots with init_ai_object. Doesn't appear to have ever been done.

### DIFF
--- a/similar/main/fireball.cpp
+++ b/similar/main/fireball.cpp
@@ -964,7 +964,7 @@ objptridx_t drop_powerup(int type, int id, int num, const vms_vector &init_vel, 
 
 				obj->shields = Robot_info[get_robot_id(obj)].strength;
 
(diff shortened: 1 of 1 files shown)


kreatordxx:
Yes I think I'll completely revise this and fix it how you suggested in bug #293 to @zicodxx (reading that properly would have saved me a lot of time!)

vLKp:
I pushed [Use `init_ai_object` for dropped robots](/dxx-rebirth/dxx-rebirth/commit/cc885f17a3a0443f58317f168cfd4fefef039d56) instead of merging this.  That commit was created by taking your pull request, then moving the `init_ai_object` call up so that the `drop_powerup` assignment of `obj->mtype.phys_info.velocity` would override the zeroing in `init_ai_object`.

PR #312, opened 2017-01-21
Move calls to init_ai_object into obj_create and fix robot creation bug with editor

kreatordxx:
Move calls to init_ai_object into obj_create as discussed in #293 and fix a bug with the editor, where creating a level with a robot then playing it would throw an exception.


kreatordxx:
Done - created `static inline` function `robot_create`.

vLKp:
:+1:

Issue #309, opened 2017-01-15
D2X: Ship status not carried over from secret levels


zicodxx:
As posted here: http://www.dxx-rebirth.com/frm/index.php/topic,2150.0.html

If a player returns from a Descent 2 secret level ALIVE, whatever they gathered in there should be preserved. If they DIE, they start as a new ship in the base level. The latter seems to work as intended but should the player survive the level, in current master, their inventory is reverted to the state before they entered the level. 
This is related to a lot of player status variables being moved to player_info in the object struct which is unconditionally restored upon the player leaving the secret level.

Now this change was done over multiple commits and I have not yet fully documented all the variables that are unintentionally reset. It's probably the easiest to compare player_info between current master and 0.58.1 to get the proper list. But it also makes this a bit cumbersome to fix easily since some variables in player_info sure should be reset while others probably should not. 

I would personally go on and save the persistent variables and restore them individually after reading the player object in case of secret_restore::survived - similar to how it is done in for "Read player info", using the dummy_player. This could be done when the game resets the player position and inventory in case of death in secret level which happens after the object reading anyway. Or do you have a better idea, @vLKp ?

vLKp:
That sounds reasonable.  Restoring them should be easy since the save is loaded into a local `player_info` that is later copied to the live one.  Fields which should not be restored from the savegame could be copied from the live player into the local variable inside the `secret_restore == survived` block, so that the later bulk copy leaves them intact.

Although it is far too late to fix this, it seems rather awkward to use a selectively reloaded savegame as the mechanism for tracking secret levels.

zicodxx:
Technically, it creates a savegame to store the exact state of the normal level while the player is visiting the secret sub-level. Since these savestates are only intended to be temporal (as one cannot save inside secret levels) I suppose we could just store the base level in RAM, too. Nowadays that shouldn't be that big of a deal. But I am not sure if that would be worth the effort.

vLKp:
I think, as currently implemented, it would not be worth it.

kreatordxx:
> It's probably the easiest to compare player_info between current master and 0.58.1 to get the proper list
That's easy @zicodxx, I had a look at the history and the entire `player_info` struct that's part of the `ctype` union was created by @vLKp. He probably intended to move the entire `player` struct into there, which would have been nice - the limitation was probably keeping the size of the union the same (i.e. `player_info` is not larger than any of the other union members).

vLKp:
[...]

PR #313, opened 2017-01-24
Player keeps loot if going between secret level and normal level

kreatordxx:
Fix bug #309. Also keep omega cannon charge. If the player dies, keep any keys collected on the secret level - or the game will become un-winnable.

### DIFF
--- a/similar/main/state.cpp
+++ b/similar/main/state.cpp
@@ -1504,6 +1504,9 @@ int state_restore_all_sub(const char *filename, const secret_restore secret)
 	{
 	player_info pl_info;
 	fix pl_shields;
+#if defined(DXX_BUILD_DESCENT_II)
+	player_info ret_pl_info;
+#endif
 	{
 #if DXX_USE_EDITOR
 		// Don't bother with the other game sequence stuff if loading saved game in editor
@@ -1518,6 +1521,8 @@ int state_restore_all_sub(const char *filename, const secret_restore secret)
 			player	dummy_player;
 			state_read_player(fp, dummy_player, swap, pl_info, pl_shields);
 			if (secret == secret_restore::survived) {		//	This means he didn't die, so he keeps what he got in the secret level.
+				ret_pl_info = plrobj.ctype.player_info;
+				pl_shields = plrobj.shields;
 				get_local_player().level = dummy_player.level;
 				get_local_player().time_level = dummy_player.time_level;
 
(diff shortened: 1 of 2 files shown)


similar/main/state.cpp
```diff
@@ -1536,6 +1536,12 @@ int state_restore_all_sub(const char *filename, const secret_restore secret)
 				do_cloak_invul_secret_stuff(old_gametime, ret_pl_info);
 			} else {
 				get_local_player() = dummy_player;
+				// Keep keys even if they died on secret level (otherwise game becomes impossible)
+				// Example: Cameron 'Stryker' Fultz's Area 51
+				pl_info.powerup_flags |= (plrobj.ctype.player_info.powerup_flags &
```
vLKp: This seems like a semantic change from 0.58.1.  This is in the `!= secret_restore::survived` path, so it seems like we ought to remove the keys obtained on the secret level and use only the keys from the save game.  In 0.58.1, this branch overwrote the active player with `dummy_player` and did nothing else; in that version, key flags were stored in `struct player`, so player keys would have reverted to whatever was in `dummy_player`, which had just been read from the secret save.  Did 0.58.1 really allow the player to keep keys from a secret level after dying there?  If not, why are there levels that require that behavior in order to let the player win?

kreatordxx: You're right, I had a copy of 0.57.3 on hand and it doesn't keep any keys I got on a secret level if I die on the secret level. However, it does keep the keys if I die on a normal level.

I would argue that this was overlooked by the original programmers simply because none of the secret levels in Descent II: Counterstrike have keys in them. It is possible to win Area 51, you just have to make sure you don't die between getting a key on the secret level and returning to the base level. However, if you do die in that situation, even if it's not Game Over, any keys you picked up in the secret level in that session of playing it are permanently lost, making the mission unwinnable. Quite frankly, I think this was overlooked by the mission creator. He made a note of getting all the keys before blowing up the secret level, but not of this possibility I described.

kreatordxx:
Unfortunately when the objects are read (fortunately for debugging purposes), `state_object_rw_to_object` initialises the whole object struct to `0` (or `0xfd` if poisoning is enabled).

similar/main/state.cpp
vLKp: As mentioned in the separate ticket (#314), I have no objection to changing the semantics here.  I flagged it because I thought this pull was solely about fixing regressions caused by the use of `player_info`, and this is not a regression.  Entering a secret level does not clear any keys that the player gained in the base level, so for consistency, it makes sense that returning from the secret level should not clear any keys the player gained in the secret level.

:+1: for this hunk.

vLKp:
I forgot that it loads objects after loading this block.  I thought that the live player's `player_info` was available as stable scratch space at the point where you use `ret_pl_info`, but it is not.  As you say, it is overwritten by `state_object_rw_to_object` and that happens **after** this section.  While it might be possible to avoid an extra variable, it's definitely not as easy as I thought above.  I see no reason to hold this trying to be stingy with variables.

:+1: for this pull.

Issue #315, opened 2017-01-26
Bald guy easter egg crashes D1X-Rebirth (applicable to Mac D1 full version data)


kreatordxx:
Playing D1X-Rebirth with the full version Mac data, I hit ALT-B during a briefing screen. Instead of showing Dravis wearing a silly hat, it crashes D1X-Rebirth.

This was reported by @roncli.

vLKp:
Is this a regression relative to 0.58.1?  Briefing memory management has been cleaned up a bit, but briefings are comparatively lightly touched.

kreatordxx:
Actually, I wonder if this ever worked with D1X-Rebirth. When I try with 0.57.3 it crashes or skips the rest of the briefing. 0.55 was the first one to have this feature, but I can't get the source to compile.

PR #317, opened 2017-01-29
Get 'BALDGUY' easter egg to work for D1X-Rebirth (Mac data)

kreatordxx:
Pressing ALT-B while in the briefing screens will show Dravis wearing a silly hat, for D1X-Rebirth with Mac data. Also add back the 'BALDGUY' cheat code so the next briefing featuring Dravis will show the same. Fixes bug #315.


vLKp:
I have not tested this, but it looks fine to me.

:+1:

PR #318, opened 2017-01-29
Add back homing weapons cheat

kreatordxx:
Add back homing weapons cheat for D2X-Rebirth, only this time, they all home! Also fix a typo with 'ORALGROOVE' cheat.


kreatordxx:
Good idea with using the second bit of `homing_flag`. I had thought of adding a `homing_flag` to every weapon object but thought that would add too much extra maintenance for little gain.

The **cheats2** branch works fine for me.

Issue #104, opened 2015-06-11
Only flush fire inputs on respawn


Mako88:
Instead of flushing ALL inputs, we should only flush fire inputs (primary, secondary, flare, bomb).

This will make sure people don't cheat by firing blindly during respawn...but it will allow you to start moving immediately without having to press a button twice.

vLKp:
I believe the rationale for clearing the fire input was not a concern about cheating, but rather that users who pressed "fire weapon" in reaction to the respawn "press any key" instruction would begin firing upon respawn, which could be bad in team or cooperative games.  Even in games where no one is hit by the errant fire, players might dislike the wasted energy, the potential for accidentally opening doors, etc.

zicodxx:
Yeah. I recall way back when I adopted D1X, the flushing was disabled (and you could rotate the view when dead) and it caused me to fire each time I respawned. I guess we might implement a game_flush_inputs() variant that only clears the firing inputs (pri, sec, flare, etc) while letting the movement inputs untouched so a player could accelerate immediately after spawn by holding the button down while still in the dead-player-view. I think that is the primary concern. Right now you have to let go of each button and press them again after the spawn.

vLKp:
Modifying the code not to flush navigation inputs in `HandleDeathInput` is easy, but they are flushed a second time by the activation of the game window.

```
#1  0x00005555555cee19 in game_flush_inputs () at similar/main/game.cpp:374
#2  game_handler (event=...) at similar/main/game.cpp:1114
#3  0x0000555555576113 in window_send_event (wind=..., event=...) at common/arch/sdl/window.cpp:208
#4  0x000055555557627b in WINDOW_SEND_EVENT (file=0x5555556dc126 "common/arch/sdl/window.cpp", e=0x5555556dc141 "EVENT_WINDOW_ACTIVATED", line=179, event=..., w=...) at common/include/window.h:111
#5  window_set_visible (w=..., visible=visible@entry=1) at common/arch/sdl/window.cpp:179
#6  0x00005555555e90f5 in window_set_visible (visible=1, wind=<optimized out>) at common/include/window.h:90
#7  DoPlayerDead () at similar/main/gameseq.cpp:1479
#8  0x000055555563ed65 in dead_player_frame () at similar/main/object.cpp:1486
#9  0x00005555555cf4df in GameProcessFrame () at similar/main/game.cpp:1329
#10 game_handler (event=...) at similar/main/game.cpp:1157
```

[...]

PR #320, opened 2017-02-07
Allow player to press a movement key to respawn then move with the same keypress

kreatordxx:
Adds enhancement discussed in #104.


vLKp:
:+1:

I have not tested this, but I see nothing concerning in it.  If it does what you want, I say merge it.

PR #324, opened 2017-02-20
Fix compiler error for Clang in maximum_value() method

kreatordxx:
Fixes this compiler error:
```
common/include/vecmat.h:166:10: error: call to constructor of 'dcx::vm_distance_squared' is ambiguous
        return vm_distance_squared{0x7fffffffffffffff};
              ^                  ~~~~~~~~~~~~~~~~~~~~

common/include/vecmat.h:130:2: note: candidate constructor has been explicitly deleted
    vm_distance_squared(const fix &) = delete;
    ^

common/include/vecmat.h:131:21: note: candidate constructor
    constexpr explicit vm_distance_squared(const fix64 &f2) :
                       ^

common/include/vecmat.h:126:7: note: candidate is the implicit move constructor
class vm_distance_squared
      ^
common/include/vecmat.h:126:7: note: candidate is the implicit copy constructor
```

### DIFF
--- a/common/include/vecmat.h
+++ b/common/include/vecmat.h
@@ -163,7 +163,7 @@ public:
 	}
 	static constexpr vm_distance_squared maximum_value()
 	{
-		return vm_distance_squared{0x7fffffffffffffff};
+		return vm_distance_squared{static_cast<fix64>(0x7fffffffffffffff)};
 	}
 	static constexpr vm_distance_squared minimum_value()
 	{


common/include/vecmat.h
```diff
@@ -163,7 +163,7 @@ class vm_distance_squared
 	}
 	static constexpr vm_distance_squared maximum_value()
 	{
-		return vm_distance_squared{0x7fffffffffffffff};
+		return vm_distance_squared{static_cast<fix64>(0x7fffffffffffffff)};
```
vLKp: NAK.  `0x7fffffffffffffff` is too big to fit in an `int`, so it is already `long`.  On x86_64-pc-linux-gnu, `fix64` is `long`, so this change adds a useless cast, which causes gcc to warn (due to `-Wuseless-cast`).  It looks wrong to me that clang would even consider the deleted constructor, since it should not try to bind `const fix &` to a value that is too big to fit in an `int`.

Does the below patch work for you?  `INT64_MAX` is defined in `stdint.h` to include a platform-appropriate suffix to request the right type.  For me, this code change has no effect on the generated code.

```diff
--- a/common/include/vecmat.h
+++ b/common/include/vecmat.h
@@ -163,7 +163,7 @@ public:
 	}
 	static constexpr vm_distance_squared maximum_value()
 	{
-		return vm_distance_squared{0x7fffffffffffffff};
+		return vm_distance_squared{INT64_MAX};
 	}
 	static constexpr vm_distance_squared minimum_value()
 	{
```

kreatordxx: Yes that works fine.

It probably isn't using the deleted constructor, just noting that it has been deleted. The other constructor being considered would probably be the implicit move or implicit copy constructor, which would make Clang think it's ambiguous (even though `vm_distance_squared` is a wrapped `fix64`). I don't really know, Clang has some pretty zealous warnings.

vLKp: [Use stdint constants for some INT*_MAX](/dxx-rebirth/dxx-rebirth/commit/08e4a6e6203bec1b44326e6e7d1081a21b8f140a) changes this site as shown above and cleans up a few other places that integer literals had been hardcoded.  Based on your prior feedback, I expect this should resolve the clang failure you saw without introducing a gcc failure.

kreatordxx: Great, this works fine. Closing.

Issue #326, opened 2017-02-24
Boss Looping Sound Doesn't Play when Saved Game Loaded


kreatordxx:
When loading a saved game for a level with a boss, the boss has to teleport again before the looping sound is played again. This happened with the original too.

It seems the intention was for the boss to play the looping sound after it has teleported at least once. I would argue this should be the case for if the boss teleported before the loaded state was saved. However, since this would change game semantics slightly, posting an issue.

vLKp:
I agree with your argument.  In my opinion, saving and restoring should have no observable effects on the game play.  Any place where the results differ if the player saves, quits, and later loads are likely an indication of a bug.

PR #327, opened 2017-02-24
Fix 2 boss teleporting related issues

kreatordxx:
If boss has teleported before, even before a loaded state was saved, play the boss looping sound immediately when loading the saved game (if near the boss). Resolves issue #326.

Also make the boss teleport even if it was generated by a robot maker or released by another robot. Fixes issue #328. Tacking onto this pull request because it depends on the first commit.

### DIFF
--- a/common/main/ai.h
+++ b/common/main/ai.h
@@ -99,7 +99,7 @@ namespace dsx {
 void create_awareness_event(vobjptr_t objp, player_awareness_type_t type);         // object *objp can create awareness of player, amount based on "type"
 ai_mode ai_behavior_to_mode(ai_behavior behavior);
 void do_ai_robot_hit(vobjptridx_t robot, player_awareness_type_t type);
-void init_ai_object(object &objp, ai_behavior initial_mode, segidx_t hide_segment);
+void init_ai_object(vobjptridx_t objp, ai_behavior initial_mode, segidx_t hide_segment);
 }
 
 namespace dcx {
@@ -182,6 +182,7 @@ static inline void force_dump_ai_objects_all(const char *msg)
 namespace dsx {
 void start_boss_death_sequence(vobjptr_t objp);
 extern void ai_init_boss_for_ship(void);
+extern void boss_init_all_segments(const vobjptridx_t boss_objnum);
 
 #if defined(DXX_BUILD_DESCENT_II)
 extern vms_vector Last_fired_upon_player_pos;
--- a/similar/editor/medrobot.cpp
+++ b/similar/editor/medrobot.cpp
@@ -87,7 +87,7 @@ namespace dsx {
 static window_event_result robot_dialog_handler(UI_DIALOG *dlg,const d_event &event, robot_dialog *r);
 
 }
-static void call_init_ai_object(object &objp, ai_behavior behavior)
+static void call_init_ai_object(vobjptridx_t objp, ai_behavior behavior)
 {
 	segnum_t	hide_segment;
 
@@ -110,7 +110,7 @@ static void call_init_ai_object(object &objp, ai_behavior behavior)
 static int RobotNextType()
 {
 	if (Cur_object_index != object_none )	{
-		const auto &&obj = vobjptr(Cur_object_index);
+		const auto &&obj = vobjptridx(Cur_object_index);
 		if (obj->type == OBJ_ROBOT)
 		{
 			obj->id++;
@@ -139,7 +139,7 @@ static int RobotNextType()
 static int RobotPrevType()
 {
 	if (Cur_object_index != object_none )	{
-		const auto &&obj = vobjptr(Cur_object_index);
+		const auto &&obj = vobjptridx(Cur_object_index);
 		if (obj->type == OBJ_ROBOT)
 		{
 			if (obj->id == 0 ) 
@@ -606,7 +606,7 @@ window_event_result robot_dialog_handler(UI_DIALOG *dlg,const d_event &event, ro
 		if (GADGET_PRESSED(r->initialMode[i].get()))
 		{
 			const auto b = static_cast<ai_behavior>(MIN_BEHAVIOR + i);
-			const auto &&objp = vobjptr(Cur_object_index);
+			const auto &&objp = vobjptridx(Cur_object_index);
 			auto &behavior = objp->ctype.ai_info.behavior;
 			if (behavior != b) {
 				behavior = b;		// Set the ai_state to the cooresponding radio button
--- a/similar/main/ai.cpp
+++ b/similar/main/ai.cpp
@@ -2137,12 +2137,10 @@ static void teleport_boss(const vobjptridx_t objp)
 	//	After a teleport, boss can fire right away.
 	ai_local		*ailp = &objp->ctype.ai_info.ail;
 	ailp->next_fire = 0;
-#if defined(DXX_BUILD_DESCENT_I)
-	digi_link_sound_to_object2(SOUND_BOSS_SHARE_SEE, objp, 1, F1_0, vm_distance{F1_0*512});	//	F1_0*512 means play twice as loud
-#elif defined(DXX_BUILD_DESCENT_II)
+#if defined(DXX_BUILD_DESCENT_II)
 	ailp->next_fire2 = 0;
-	digi_link_sound_to_object2(Robot_info[get_robot_id(objp)].see_sound, objp, 1, F1_0, vm_distance{F1_0*512});	//	F1_0*512 means play twice as loud
 #endif
+	boss_link_see_sound(objp);
 
 }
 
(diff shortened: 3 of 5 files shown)


common/main/ai.h
```diff
@@ -99,7 +99,7 @@ namespace dsx {
 void create_awareness_event(vobjptr_t objp, player_awareness_type_t type);         // object *objp can create awareness of player, amount based on "type"
 ai_mode ai_behavior_to_mode(ai_behavior behavior);
 void do_ai_robot_hit(vobjptridx_t robot, player_awareness_type_t type);
-void init_ai_object(object &objp, ai_behavior initial_mode, segidx_t hide_segment);
+void init_ai_object(objptridx_t objp, ai_behavior initial_mode, segidx_t hide_segment);
```
vLKp: Since `init_ai_object` always dereferences this without checking, the type should be `vobjptridx_t`, to prevent someone trying to pass `nullptr` or `object_none`.

common/main/ai.h
```diff
@@ -182,6 +182,7 @@ static inline void force_dump_ai_objects_all(const char *msg)
 namespace dsx {
 void start_boss_death_sequence(vobjptr_t objp);
 extern void ai_init_boss_for_ship(void);
+extern void boss_init_all_segments(objptridx_t boss_objnum);
```
vLKp: `extern` is unnecessary, but harmless.

vLKp: Both callers of `boss_init_all_segments` pass a `vobjptridx_t`.  While it can be transparently converted to `objptridx_t`, retaining its original type would document that it must be a valid object.  That would then remove the need for one branch of the new `Assert` inside `init_boss_segments`.

similar/editor/medrobot.cpp
```diff
@@ -87,7 +87,7 @@ namespace dsx {
 static window_event_result robot_dialog_handler(UI_DIALOG *dlg,const d_event &event, robot_dialog *r);
 
 }
-static void call_init_ai_object(object &objp, ai_behavior behavior)
+static void call_init_ai_object(objptridx_t objp, ai_behavior behavior)
```
vLKp: All callers pass a `vobjptridx_t`.  While it can be transparently converted to `objptridx_t`, retaining its original type would document that it must be a valid object.

similar/main/ai.cpp
```diff
@@ -89,7 +89,7 @@ using std::min;
 #define	BABY_SPIDER_ID	14
 
 namespace dsx {
-static void init_boss_segments(boss_special_segment_array_t &segptr, int size_check, int one_wall_hack);
+static void init_boss_segments(objptridx_t boss_objnum, boss_special_segment_array_t &segptr, int size_check, int one_wall_hack);
```
vLKp: Since the first step in the called function is to reject `object_none`, you could use `vobjptridx_t` to enforce that no caller can pass `object_none`.

similar/main/ai.cpp
```diff
@@ -481,11 +481,25 @@ void ai_init_boss_for_ship(void)
 #endif
 }
 
+void boss_init_all_segments(objptridx_t boss_objnum)
```
vLKp: Where possible, I prefer to have the definition use a `const` argument, e.g. `void boss_init_all_segments(const vobjptridx_t boss_objnum)` to document that the argument is never redirected to a new object.  In this context, the `const` applies only to the pointer/integer in the `objptridx_t`, but not the contained object, so it is still possible to write to the boss robot data.  If you do not need to write to it, use the `c` variant: `const vcobjptridx_t`, a constant pointer/integer that references constant data.  In this case, one of the called functions does write to the underlying `object`, so the `c` variant is not appropriate here.

similar/main/ai.cpp
```diff
@@ -481,11 +481,25 @@ void ai_init_boss_for_ship(void)
...
+{
+	if (Boss_teleport_segs.count())
```
vLKp: Although not relevant here (since the count is tracked as a separate field that can be easily retrieved and tested), it is generally preferable to test for `!C.empty()` rather than `C.count()`, since some containers need O(N) time to count their elements, but only O(1) to distinguish between empty and non-empty.  This container works the same regardless of method, so there is no need to change this line.

similar/main/ai.cpp
```diff
@@ -3066,6 +3075,67 @@ void do_ai_frame(const vobjptridx_t obj)
...
+	switch (robptr->boss_flag) {
+		case 0:
+			break;
+			
+		case BOSS_D1:
+			if (aip->GOAL_STATE == AIS_FLIN)
+				aip->GOAL_STATE = AIS_FIRE;
+			if (aip->CURRENT_STATE == AIS_FLIN)
+				aip->CURRENT_STATE = AIS_FIRE;
+			dist_to_player /= 4;
```
vLKp: I see that this whole block is merely moved from somewhere else, so you might not be able to answer, but I will ask anyway: Why is `dist_to_player` divided here, then reverted after the call to `do_boss_stuff`?  Since it is a local, `do_boss_stuff` cannot see its value.

similar/main/multibot.cpp
```diff
@@ -1033,7 +1033,7 @@ void multi_do_boss_teleport(const playernum_t pnum, const ubyte *buf)
 
 	digi_link_sound_to_pos( Vclip[VCLIP_MORPHING_ROBOT].sound_num, teleport_segnum, 0, boss_obj->pos, 0 , F1_0);
 	digi_kill_sound_linked_to_object( boss_obj);
-	digi_link_sound_to_object2(SOUND_BOSS_SHARE_SEE, boss_obj, 1, F1_0, vm_distance{F1_0*512});	//	F1_0*512 means play twice as loud
+	boss_link_see_sound(boss_obj);
```
vLKp: This is a minor semantic change in Descent 2.  Previously, the code always used `SOUND_BOSS_SHARE_SEE` (defined as `183`); now it will use the sound specified in the robot_info structure.  This is probably a change for the better, but I felt I should point out that it is a change.

If I did the analysis correctly, the boss robot info specifies these see sounds:

Descent 1: First Strike, Level 7: 170
Descent 1: First Strike, Final: 170
Descent 2: Counterstrike, Level 4: 183
Descent 2: Counterstrike, Level 8: 225
Descent 2: Counterstrike, Level 12: 186
Descent 2: Counterstrike, Level 16: 192
Descent 2: Counterstrike, Level 20: 237
Descent 2: Counterstrike, Level 24: 195
Descent 2: Vertigo, first boss: 195
Descent 2: Vertigo, second boss: 47

common/main/ai.h
kreatordxx: Good idea, I was ignorant of the difference between `vobjptridx_t` and `objptridx_t`.

similar/main/ai.cpp
kreatordxx: Ah yes, I think I used the `c` variant with another unrelated change. I probably tried it with these changes at some point but the compiler complained, because it was being written to.

similar/main/ai.cpp
kreatordxx: OK, noted.

common/main/ai.h
vLKp: Instances of the base name wrap a value that is either None or a valid target.  The `c` prefix denies writing to the target, but still permits changing where the *X*__ptridx__ points.  The `v` prefix denies using None.  The `c` prefix is always compile-time checked, since it causes the contained pointer to be `const T *` instead of `T *`, so the compiler enforces the const-ness.  The `v` prefix is compile-time checked where possible (e.g. by having a `static_assert` in the implementation), but runtime-checked where necessary.  The `v` and `c` prefixes can be combined if the target is immutable and guaranteed not to be None.

```c++
    auto a = objptridx(i); // checks that i is -1 or a valid index, throws if neither
    vobjptridx_t b = a; // checks that a is a valid index, throws if not; if i is -1, this line throws
```

similar/main/ai.cpp
kreatordxx: I think this was simply carrying on from the same thinking with the other boss cases, without it having been thought through. Removed.

similar/main/multibot.cpp
kreatordxx: Yeah, I was a bit confused by why it didn't use the Descent 2 boss sounds. I'll leave this change, but noted.

kreatordxx:
Since I think they're a good idea to integrate into the changes I made, I made the changes and used `git commit -a --amend` then `git push -f` (which doesn't wipe out your feedback like `git reset HEAD~1` then `git commit` does!).

similar/main/multibot.cpp
vLKp: `git blame` says it was `SOUND_BOSS_SHARE_SEE` when btb imported the original sources into d2x.  It is probably an accidental holdover caused by Parallax copying the Descent source to use as a base for Descent 2, extending the bosses, and never revisiting some of their earlier assumptions.

Issue #329, opened 2017-02-26
If boss created in segment it can't fit in, it can still teleport to it


kreatordxx:
This applies to Jonadab's Domain Level 5 (see issue #328 for download), blue access area. The robot maker generates smart missile bosses, where the boss doesn't fit entirely in the robot maker - yet can sometimes teleport back into the robot maker.

I argue the boss should move out and into the large prism at the end where it can fit, and not bother teleporting back into the robot maker - it wouldn't fit and fire at the player properly (instead just blocking the player). This would be particularly useful for bosses generated by a robot maker or dropped by another robot (admittedly uncommon cases at current).

vLKp:
Are there any further changes required before closing this ticket?

kreatordxx:
I found out the boss can stay 'stuck' in the matcen, but only because there are another 2 or more bosses. This is related to the fact there is just 1 set of globals controlling boss teleporting and cloaking. I figured this wouldn't be worth fixing - it would require the cloaking/timing variables to be stored in the object struct or have a separate array with an index in the object struct. Then there's the added complication of only being able to save the timers for one boss, unless we change the saved game format.

vLKp:
Agreed.  The engine is not at all suited to multi-boss levels.  The semantics when a boss robot is killed are also odd.  I am not opposed to making more of the boss specials per-object, although, as you say, without savegame format changes, that would be of relatively little value.  Robot data is currently the largest branch of the union, so `struct object` would grow if boss data is moved into `ai_static`.  That in turn would cause `Objects` to grow, and becomes more noticeable as `MAX_OBJECTS` rises.

kreatordxx:
Yes the boss semantics are very odd. If I destroy one boss, they all die. And they don't die at the same time - they blow up one at a time, and each time resets the countdown timer to rush to the exit.

I used the odd semantics when making my (unpublished) mission with the boss releasing two copies of itself when it dies. So they keep blowing up and multiplying, resetting the countdown timer each time. When I was experimenting ages ago, using original Descent II, it actually crashed the computer when I had a matcen generating bosses.

PR #330, opened 2017-02-26
Don't teleport to original segment if boss won't fit

kreatordxx:
Unless there are no other segments. Fixes issue #329.


vLKp:
:+1:

I have not tested this, but I see nothing concerning in it.  If it does what you want, I say merge it.

Issue #291, opened 2016-12-27
Can't load current savegames in 0.58.1


zicodxx:
It has been reported that savegames created with current master can't be loaded in DXX-Rebirth v0.58.1. I can confirm this for both D1X and D2X. Here are the backtraces I got from newly compiled 0.58.1 builds:

```
d1x-rebirth: main/object.c:927: obj_link: Assertion `segnum>=0 && segnum<=Highest_segment_index' failed.
Thread 1 "d1x-rebirth" received signal SIGABRT, Aborted.
0x00007ffff694104f in raise () from /usr/lib/libc.so.6
(gdb) bt
#0  0x00007ffff694104f in raise () from /usr/lib/libc.so.6
#1  0x00007ffff694247a in abort () from /usr/lib/libc.so.6
#2  0x00007ffff6939ea7 in __assert_fail_base () from /usr/lib/libc.so.6
#3  0x00007ffff6939f52 in __assert_fail () from /usr/lib/libc.so.6
#4  0x00000000004a88a8 in obj_link (objnum=5, segnum=1536) at main/object.c:927
#5  0x00000000004c5bcd in state_restore_all_sub (filename=0x7fffffffd250 "Players/12345678.sg2") at main/state.c:1296
#6  0x00000000004c5384 in state_restore_all (in_game=0) at main/state.c:1109
#7  0x000000000047bafc in do_option (select=5) at main/menu.c:547
#8  0x000000000047b36e in main_menu_handler (menu=0x1a5f050, event=0x7fffffffe400, menu_choice=0x1afe440) at main/menu.c:449
#9  0x00000000004a230d in newmenu_key_command (wind=0x1955030, event=0x7fffffffe400, menu=0x1a5f050) at main/newmenu.c:1019
#10 0x00000000004a45db in newmenu_handler (wind=0x1955030, event=0x7fffffffe400, menu=0x1a5f050) at main/newmenu.c:1522
#11 0x00000000004158ea in window_send_event (wind=0x1955030, event=0x7fffffffe400) at arch/sdl/window.c:211
#12 0x00000000004128ee in event_send (event=0x7fffffffe400) at arch/sdl/event.c:130
#13 0x0000000000413c0b in key_handler (kevent=0x7fffffffe450) at arch/sdl/key.c:427
#14 0x0000000000412750 in event_poll () at arch/sdl/event.c:43
#15 0x000000000041296e in event_process () at arch/sdl/event.c:152
[...]

PR #331, opened 2017-02-26
Save secondary weapons flags for 0.58.1 compatibility

kreatordxx:
Set secondary weapon flags to 0xff when saving game, so that if it's loaded with 0.58.1 or earlier the player can actually select all secondary weapons they have at least one of. Fixes issue referred to in  https://github.com/dxx-rebirth/dxx-rebirth/issues/291#issuecomment-282528356

### DIFF
--- a/similar/main/state.cpp
+++ b/similar/main/state.cpp
@@ -511,7 +511,15 @@ static void state_player_to_player_rw(const fix pl_shields, const player *pl, pl
 	pl_rw->starting_level            = pl->starting_level;
 	pl_rw->killer_objnum             = pl_info.killer_objnum;
 	pl_rw->primary_weapon_flags      = pl_info.primary_weapon_flags;
-	pl_rw->secondary_weapon_flags    = 0;
+#if defined(DXX_BUILD_DESCENT_I)
+	// make sure no side effects for Mac demo
+	pl_rw->secondary_weapon_flags    = 0x0f | (pl_info.secondary_ammo[MEGA_INDEX] > 0) << MEGA_INDEX;
+#elif defined(DXX_BUILD_DESCENT_II)
+	// make sure no side effects for PC demo
+	pl_rw->secondary_weapon_flags    = 0xef | (pl_info.secondary_ammo[MEGA_INDEX] > 0) << MEGA_INDEX
+											| (pl_info.secondary_ammo[SMISSILE4_INDEX] > 0) << SMISSILE4_INDEX	// mercury missile
+											| (pl_info.secondary_ammo[SMISSILE5_INDEX] > 0) << SMISSILE5_INDEX;	// earthshaker missile
+#endif
 	pl_rw->laser_ammo = 0;
 	pl_rw->vulcan_ammo   = pl_info.vulcan_ammo;
 	pl_rw->obsolete_primary_ammo = {};


vLKp:
Will this cause side effects for shareware data?  As I read the code, 0.58.1 will only allow you to select the weapon if you have the flag set and `HAVE_AMMO_FLAG` is set.  `HAVE_AMMO_FLAG` is set when `Weapon_info[weapon_index].ammo_usage <= Players[Player_num].secondary_ammo[weapon_num]`, which is true if you have ammo or if `ammo_usage <= 0`.  I seem to recall a report from derhass where shareware data set `ammo_usage` to `0` for weapons that did not exist in shareware (smart missiles, mega missiles).  If that applies here, then setting the flag bits for retail-only weapons will allow shareware users to select (but not fire) retail-only weapons.

kreatordxx:
For Descent 1, this isn't an issue - it still won't let me select smart missiles or mega missiles, even if I have some of each. For Descent 2, I don't know how I would test this - if I load a saved game using demo data, it crashes.

Unfortunately this fix isn't working for mercury missiles or earthshaker missiles - I still can't select them in 0.58.1 with full version data.

kreatordxx:
Fixed for all Descent 2 secondary weapons.

vLKp:
I was mainly concerned with a user who installs shareware data, starts a new game with that data using a build from tip, saves the game, quits the program, starts 0.58.1, and loads the game saved by tip.  Such a user never uses the retail data, so he has no opportunity to acquire retail-only missiles.  My concern was that, despite not having any such missiles, he would be permitted to select them.  I read your first response to be that this is not what happens in Descent 1 0.58.1 using shareware data.

I am not familiar with what weapons are enabled in the various Descent 2 shareware offerings.

Assuming it is appropriate to set all flags, even those for missiles the player lacks, you could avoid an explicit `#ifdef` by using a compile-time arithmetic expression:

```diff
diff --git a/similar/main/state.cpp b/similar/main/state.cpp
index 2b836f2..d3df33d 100644
--- a/similar/main/state.cpp
+++ b/similar/main/state.cpp
@@ -511,7 +511,7 @@ static void state_player_to_player_rw(const fix pl_shields, const player *pl, pl
 	pl_rw->starting_level            = pl->starting_level;
 	pl_rw->killer_objnum             = pl_info.killer_objnum;
 	pl_rw->primary_weapon_flags      = pl_info.primary_weapon_flags;
-	pl_rw->secondary_weapon_flags    = 0;
+	pl_rw->secondary_weapon_flags    = (1 << MAX_SECONDARY_WEAPONS) - 1;
 	pl_rw->laser_ammo = 0;
 	pl_rw->vulcan_ammo   = pl_info.vulcan_ammo;
 	pl_rw->obsolete_primary_ammo = {};
```

This expression will produce `0x1f` for Descent 1 (`(1 << 5) - 1 = 32 - 1 = 31 = 0x1f`) and `0x3ff` for Descent 2 (`(1 << 10) - 1 = 1024 - 1 = 1023 = 0x3ff`).  In both games, the result will cover all valid missile types.

kreatordxx:
Oh right, I tried that and it does actually let me select mercury missiles, mega missiles and earthshaker missiles in 0.57.1, but it shows a big ugly white square covering some of the status bar as well. It says I have zero of each missile. Strange, I did a lot of saved game hacking ages ago and never had this happen.

I'll use the `&` operator for those three missiles, checking count. Because of that, I'll need to leave the `#ifdef`.

kreatordxx:
I found a similar issue with the Mac demo data for Descent 1 as well - it lets me select the mega missile even though I don't have any, showing a blue square with a red cross. Changes revised.

PR #351, opened 2017-10-17
RPI: use new vendor graphics library names

psyke83:
Necessary for compatibility with recent firmwares included in Raspbian stretch.

If building against VC4 Mesa driver, build without "raspberrypi=1" and use
"opengles=1" for generic GLES support without dispmanx.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -2806,6 +2806,7 @@ class DXXCommon(LazyObjectConstructor):
 		# Paths for the Videocore libs/includes on the Raspberry Pi
 		RPI_DEFAULT_VC_PATH='/opt/vc'
 		default_OGLES_LIB = 'GLES_CM'
+		default_EGL_LIB = 'EGL'
 		_default_prefix = '/usr/local'
 		__stdout_is_not_a_tty = None
 		__has_git_dir = None
@@ -2872,8 +2873,12 @@ class DXXCommon(LazyObjectConstructor):
 			return self.raspberrypi
 		def selected_OGLES_LIB(self):
 			if self.raspberrypi:
-				return 'GLESv2'
+				return 'brcmGLESv2'
 			return self.default_OGLES_LIB
+		def selected_EGL_LIB(self):
+			if self.raspberrypi:
+				return 'brcmEGL'
+			return self.default_EGL_LIB
 		def __default_DATA_DIR(self):
 			return '%s/share/games/%s' % (self.prefix, self._program.target)
 		def _generic_variable(key,help,default):
@@ -2943,6 +2948,7 @@ class DXXCommon(LazyObjectConstructor):
 				'arguments': (
 					('rpi_vc_path', self.RPI_DEFAULT_VC_PATH, 'directory for RPi VideoCore libraries'),
 					('opengles_lib', self.selected_OGLES_LIB, 'name of the OpenGL ES library to link against'),
+					('egl_lib', self.selected_EGL_LIB, 'name of the OpenGL ES Graphics Library to link against'),
 					('prefix', self._default_prefix, 'installation prefix directory (Linux only)'),
 					('sharepath', self.__default_DATA_DIR, 'directory for shared game data (Linux only)'),
 				),
@@ -3170,7 +3176,7 @@ class DXXCommon(LazyObjectConstructor):
 		@property
 		def ogllibs(self):
 			user_settings = self.user_settings
-			return (user_settings.opengles_lib, 'EGL') if user_settings.opengles else ('GL', 'GLU')
+			return (user_settings.opengles_lib, user_settings.egl_lib) if user_settings.opengles else ('GL', 'GLU')
 		@staticmethod
 		def get_platform_objects(_empty=()):
 			return _empty


vLKp:
I see nothing in this patch that looks particularly concerning, but I have no Raspberry Pi based devices on which to test, nor any experience with them.

@psyke83: as I interpret your commit message, this improves support for recent Pi devices at the expense of older Pi devices, as the user must now use a different procedure to build for those devices.  Is that interpretation correct?  If so, it may be better to change the `raspberrypi` special option to take a family designation and pick appropriate libraries based on the family specified.  For backward compatibility, `1` could be made an alias to the family that is presently supported by that option.

@derhass: since you have worked on the Pi-related Rebirth code before, I would appreciate your opinion on the proposed changes.

psyke83:
@vLKp 

Not exactly... let me explain.

In Pi firmwares older than 1.20160921-1, the vendor graphics libraries in /opt/vc used the standard Mesa names: libEGL, libGLESv2, etc. Due to conflicts with the standard Mesa libraries, the Foundation opted to rename the conflicting libraries to libbrcmEGL, libbrcmGLESv2, etc., but in order to ease the transition, they provided both versions of the libraries (libEGL & libbrcmEGL, etc) for a time. In the most recent firmware (for Raspbian stretch) they've decided that the transition phase is over, and have dropped the duplicate libraries, leaving just the libbrcmXXX versions. Other projects have already started to switch to the new libraries (including SDL 2.0.6, RetroArch, etc).

In summary:
* Firmwares older than 1.20160921-1 only provide libEGL, libGLESv2, etc. They will break with this PR.
* Firmwares from 1.20160921-1 to 1.20170703-2 provide both copies. They will work with or without this PR.
* Firmware 1.20170811-1 and later only provide the libbrcmXXX version of the libraries. They require this PR or will fail to build due to the old library names not being present for the linker.

As long as users have a recent firmware image, this PR should not break anything or require any change to the build flags (it still expects raspberrypi=1).

The Mesa VC4 driver is separate to the above, and is under active development with the intent of completely replacing the vendor library (but isn't yet feature complete enough to replace it). RPI-specific code such as dispmanx won't be useful for this driver, so for your project, "raspberrypi=0 opengles=1" would be the appropriate way to target a build against the VC4 Mesa driver. But that's only for people who want to test the driver, and can be seen as future proofing.

Perhaps raspberrypi flag could be renamed to "videocore" (the vendor library) to make the intent more obvious? I'd be happy to modify the PR with whatever you suggest, of course.

vLKp:
Thank you for the elaboration.  Since large projects are converting to the new names, and since the `raspberrypi` flag is a shortcut, rather than a requirement for use on the Pi, merging this seems appropriate.  Users who use older Pi firmware will be troubled by those other projects as well, and for Rebirth, would retain the option of setting by hand the options that are currently set automatically by `raspberrypi=1`.

Although the existing flag name may (or may not) be a great choice, I am reluctant to change it now, since that would require users to update their invocation of SCons.

I will sit on this until the weekend to give any silent readers an opportunity to voice disagreement.  Barring such objection, I will merge this pull request as-is over the weekend.

psyke83:
No hurry - thank you. Here's a comment confirming what I stated from someone actually affiliated with the Foundation: https://github.com/anholt/mesa/issues/24#issuecomment-231067341

Pi users will need to be using firmware 1.20160921-1 or later to avoid issues with this PR, so a 100% smooth transition can't be guaranteed, unfortunately. But I think the onus should be on such users to upgrade to a more recent firmware, even if only for reasons of security.

derhass:
I do not have objections with the patch. It will break building on devices with the older firmware installed, but the new firmware should be available for all older devices. And we do support setting the file names for this libs directly, so supporting older firmwares shouldn't be a big deal (we could put a hint into the `README.RPi` file). Also, there are issues with rebirth and gcc-4.x anyway, so switching to a recent raspbian stretch and a recent firmware is quite reasonable for building rebirth.

I do have some remarks, though:

The `rasperrypi` switch currently does not only enable the use of the proprietary dispmanx/VideoCore APIs, it also sets `WORDS_NEED_ALIGNMENT`. Without it, you will get more or less random `SIGBUS` crashes in release builds (which use the VFP floating point unit of the ARM, and require aligned operands).

Also, we conceptually have 3 options:
a) using the proprietary dispman/videocore APIs as before
b) using the mesa VC4 driver with GLES via EGL
c) using the mesa VC4 driver with desktop GL via glX

Option c) doesn't make too much sense, because the driver is emulating lots of stuff that the HW can't do, but I did try it and it actually worked quite OK (since dxx-rebirth is also not really using much features which aren't in GLES anyway). 

So IMO, it is probably best to overhaul the raspberrypi switches to support all 3 options. Maybe something like @vLKp suggested, but just with the options `proprietary-gles`, `gles` and `gl` (though I don't like the `proprietary-gles` term, any better ideas?)

psyke83:
How about leaving raspberrypi=1 with the current behaviour to avoid issues, and "raspberrypi=open" to simply set words_need_alignment, and perhaps selecting gles by default, but allowing the user to specify gles or gl via the already existing options?

vLKp:
> Also, there are issues with rebirth and gcc-4.x anyway

gcc-4.9.x is still a supported compiler.  If you have problems with it, please report them.  I think gcc-4.8.x is now unsupported due to it rejecting a construct used in some of the Rebirth code (but even there, it should fail an SConf test, rather than start building and then reject a file).

> The `rasperrypi` switch currently `...` sets `WORDS_NEED_ALIGNMENT`.

True, but that can be set separately via SCons parameter `words_need_alignment=1`.  I take your point though.  The `raspberrypi` option sets multiple defaults useful to Pi users, not just the library names.  Some of those defaults apply to all Pi iterations, and some are specific to software choices.

> How about leaving raspberrypi=1 with the current behaviour to avoid issues

I am amenable to that.

> Edit: I'm not too familiar with scons arguments. If they can only be specified as booleans, then separate options will be necessary, sorry.

SCons supports several argument types.  Presently, `raspberrypi` is classified as a boolean argument, so that it can coerce several common truth representations into a Python boolean.  It could be specified as one of several other types:

* Integer value.
* Enumerated value.  SCons would provide a list and the user must choose any one value from that list.  Python code would then inspect the value chosen and act accordingly.
* String value.  The user provides an arbitrary string, which the Python code then inspects.  No argument-time constraints are applied.

For this purpose, an enumerated value (or set of values, if there are groups of independent choices) should work.  I'm happy to receive change requests in the form of ready-to-merge pull requests, but if you're uncomfortable using the SCons facilities, I can map your requirements into SCons Python code.  Tell me:

* How many enumerated values you want.
* What to name them.  These names will be visible to and used by people building Rebirth for the Pi, so they should be chosen such that someone familiar with the Pi, but not necessarily familiar with Rebirth, will guess the right choice.  Naming them after the common names of the firmware they target would be one way to encourage users to guess the right value.
* What SCons options are implied by each one.  For example, all of the Pi-related ones probably need to imply `words_need_alignment=1` since that is due to a hardware constraint.  Some need to imply the new library names.  One probably needs to imply the old library names.  Etc.

psyke83:
What about this? https://github.com/dxx-rebirth/dxx-rebirth/compare/unification/master...psyke83:rpi_test

It changes raspberrypi into a generic variable, and changes the logic so that it will build against the (new) vendor libraries if the variable is set to '1'. For any other value set (but the description says "open" to keep things simple), it only selects the alignment option and defaults to GLES.

I'm not sure if this is an acceptable solution, since "raspberrypi=true" will no longer work as expected. If I'm wrong, let me know and I'll update this PR with the change.

derhass:
@vLKp : 

> gcc-4.9.x is still a supported compiler. If you have problems with it, please report them.

Well, I was referring to your own recommendation in https://github.com/dxx-rebirth/dxx-rebirth/issues/289#issuecomment-274292838 , since raspbian jessie seems to have the same gcc bug debian has (no surprise there). I'm not aware of any particular issue with 4.9.2 builds, though.

@psyke83 
> What about this? 
https://github.com/dxx-rebirth/dxx-rebirth/compare/unification/master...psyke83:rpi_test

I do not really have a preference for which the alternatives for the `raspberrypi` option we should choose, so your suggestion is fine for me. All I would ask for is that we update `README.RPi` accordingly, documenting all the options there are now, including the command line required to build this on an old firmware.

psyke83:
I'm not happy with the rpi_test branch; it seems that an enumerated value would be preferable, as @vLKp suggested, but I didn't manage to get it working. When I specify "raspberrypi" as an EnumVariable (in the same array as the host_endian and host_platform), I get this error:

```
scons: Reading SConscript files ...
===== D1X-Rebirth v0.59.100 0.59.100-3-g6f20cc17efd6*+ =====
===== D2X-Rebirth v0.59.100 0.59.100-3-g6f20cc17efd6*+ =====
AttributeError: UserSettings instance has no attribute 'raspberrypi':
  File "/mnt/usb-dodgy/tmp/me/dxx-rebirth/SConstruct", line 4215:
    main(register_program)
  File "/mnt/usb-dodgy/tmp/me/dxx-rebirth/SConstruct", line 4190:
    ''.join(['%s:\n%s' % (d.program_message_prefix, d.init(substenv)) for d in dxx])
  File "/mnt/usb-dodgy/tmp/me/dxx-rebirth/SConstruct", line 3903:
    self.user_settings.read_variables(self.variables, substenv)
  File "/mnt/usb-dodgy/tmp/me/dxx-rebirth/SConstruct", line 3093:
    value = value()
  File "/mnt/usb-dodgy/tmp/me/dxx-rebirth/SConstruct", line 2874:
    if self.raspberrypi:
```

I'm not sure what I need to do in order to ensure that enumerated values also get parsed in the same way as booleans or generic variables in the required parts of the script, but given more time I hope to figure it out.

Once I've got that sorted, I'll update the PR to expand upon the build instructions (how to target builds for the old library names when using an old firmware, and for the VC4 Mesa driver).

vLKp:
@derhass:

> Well, I was referring to your own recommendation in https://github.com/dxx-rebirth/dxx-rebirth/issues/289#issuecomment-274292838 

Ah, right.  In that comment, I was mainly hoping people would update Debian to gcc 4.9.4 for all the Debian targets (x86, x86_64, arm, etc.), particularly since the fix was already quite old by the time that ticket came to me.  I haven't dropped support for 4.9.2, but it is not something I test, either.  If users want to use gcc 4.9.2, and have a problem with it, I'll try to help and will consider adding workarounds (as in that ticket).

@psyke83: would you mind pushing a temporary branch containing your unsuccessful attempt with the EnumVariable?  I don't mind writing the change for you since you've provided good details on what you want, but if you would prefer to do it yourself, I'm willing to help you learn how.  Based on a quick draft I attempted, I think I see why it failed for you.  The variable parsing code has an underdocumented (but previously known to me, sorry) quirk that makes ordering important.  UserSettings instances acquire the relevant attributes in the order they appear in the tuple that lists all the options.  On its own, this implementation detail is irrelevant.  Combine it with the way that default values are set by calling methods (such as `selected_EGL_lib(self)`) which can in turn inspect other attributes on the UserSettings and it becomes important.  This is hinted at in a comment later on, but on re-reading it, I wouldn't expect someone new to the code to recognize the significance:

https://github.com/dxx-rebirth/dxx-rebirth/blob/71cfd9bc6ca230ae3893e14e8c3ee617cacf348f/SConstruct#L3402-L3404

In short: any function which needs to inspect the state of the `raspberrypi` attribute must not be called until after that attribute is set.  Functions that supply default values are called when their corresponding attribute is set.  Therefore, the `raspberrypi` attribute must appear in the `_options` tuple above any lines which derive their default value from a function that inspects that variable.  The only `EnumVariable` uses presently in the `_options` tuple are near the end, well after all the functions that inspect the `raspberrypi` attribute, so grouping the enum `raspberrypi` together with other enum variables would fail in the manner you showed.  Again, my apologies for not documenting this quirk more clearly.

If you'd like, I can write up an enum based implementation and publish it to a temporary branch for your review (to verify it produces proper library names, uses names meaningful to Raspberry Pi users, etc.).

psyke83:
@vLKp

Many thanks for the explanation, it explains my difficulties. I'm stubborn, so I gave it a shot myself: https://github.com/dxx-rebirth/dxx-rebirth/compare/unification/master...psyke83:rpi_test2

I was able to change "raspberrypi" into an EnumVariable thanks to your explanation of the importance of the ordering. I added three possible values - "yes", "no" and "mesa", but added some maps so that 1/true or 0/false will map onto the appropriate values. I also updated the documentation to explain how to build against mesa or on a very old firmware.

Are you happy with this change? If so, I'll modify it this existing PR to match the rpi_test2 branch.

psyke83:
Also, to make explicit the intended difference between "yes" vs "mesa":

raspberrypi=yes
* select GLES with "brcm" names
* set alignment
* add /opt/vc include & lib directories, link bcm_host & set RPI flags (to enable dispmanx code paths)

raspberrypi=mesa
* select GLES (no name change necessary)
* set alignment
* Don't expose /opt/vc, link bcm_host or set RPI flags, as dispmanx is not wanted

It makes sense to auto-select GLES for VC4, but it can be overridden. I've included that info in the commit description as well as the README.RPi.

vLKp:
That change looks fine, and can be merged (almost) as-is.  I added one commit on top of it to [address a potential user confusion with the `ignorecase` option](/dxx-rebirth/dxx-rebirth/commit/1eb29c6e711777001d3387d9a3918d5260b5c42d), but otherwise merged it as-is from the tip of your [rpi_test2](/psyke83/dxx-rebirth/tree/rpi_test2) branch.

psyke83:
Much appreciated - thank you, and apologies for overlooking the ignorecase issue. One question before I close this PR: I proposed merging the fix to the "unification/master" branch due to [RetroPie's default branch selection](https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/ports/dxx-rebirth.sh#L25).

Would it be prudent for me to encourage RetroPie to switch to the master branch, or is the unificiation/master branch more appropriate for a Pi for any reason (e.g. better stability or performance compared to master)? Regardless, I'll target a build against master to ensure it's working OK on my Pi 3 ASAP. Thanks.

psyke83:
Update: the master branch seems to run fine with the shareware data for both Descent 1 & 2 on Raspbian stretch. I'll discuss the issue with @joolswills, no need to reply. Thanks.

vLKp:
Branch `unification/master` was considered for use backporting fixes for the upcoming release, but due to other issues, tagging a release is so far behind schedule that it is likely it will need to come from `master`.  I suggest switching to branch `master`.

PR #359, opened 2017-12-23
account for OpenBSD

thfrwn:
d1x-rebirth builds nicely on OpenBSD. Plz consider this pull request which only adjusts ifdefs and SConstruct platforms to account for OpenBSD as a target.


similar/editor/med.cpp
```diff
@@ -891,7 +891,7 @@ static void close_editor()
 {
 	//	_MARK_("end of editor");//Nuked to compile -KRB
 	
-#ifndef __linux__
+#if !defined (__linux__) || !defined(__OpenBSD__)
```
vLKp: This seems wrong to me.  Linux is not OpenBSD, so `(!linux || !openbsd)` -> `!(linux && openbsd)` -> `!true` -> `false`.  What is your goal here?  Do you want OpenBSD to use the same path as Linux, a path which calls `clear_warn_func`?

similar/main/inferno.cpp
```diff
@@ -95,7 +95,7 @@ char copyright[] = "DESCENT II  COPYRIGHT (C) 1994-1996 PARALLAX SOFTWARE CORPOR
 #endif
 #include "event.h"
 #include "rbaudio.h"
-#ifndef __linux__
+#if !defined (__linux__) || !defined (__OpenBSD__)
```
vLKp: Both of these hunks have the same concern as in `med.cpp`.

SConstruct
```diff
@@ -3398,7 +3398,7 @@ class DXXCommon(LazyObjectConstructor):
 				'variable': EnumVariable,
 				'arguments': (
 					('host_endian', None, 'endianness of host platform', {'allowed_values' : ('little', 'big')}),
-					('host_platform', 'linux' if sys.platform == 'linux2' else sys.platform, 'cross-compile to specified platform', {'allowed_values' : ('win32', 'darwin', 'linux')}),
+					('host_platform', 'linux' if sys.platform == 'linux2' else sys.platform, 'cross-compile to specified platform', {'allowed_values' : ('win32', 'darwin', 'linux', 'openbsd6')}),
```
vLKp: At first look, this change seems useless, since you added no handling of `host_platform` anywhere else, so all later code treats it as unknown.

By chance, all the uses you likely care about test for `!windows` and/or `!darwin`, leaving any unrecognized value as `linux` (which in practice is "Linux and things similar to Linux", hence why it worked for you).

I suspect you only did this because `sys.platform` is `openbsd6` for you, so the default value was not on the `allowed_values` list, causing an SCons error.  Is my interpretation correct?

thfrwn: Yes, you are correct. This change got me past the configure error that the platform was not on the list. So far this is what helps get this to build on OpenBSD - sounds like there would be a better way, but I don't understand the configuration well enough for that.

similar/editor/med.cpp
thfrwn: Yes, that was my goal. I guess I messed up the logic and this should be !linux && !openbsd

similar/main/inferno.cpp
thfrwn: same as above - should be &&, not ||

SConstruct
vLKp: For now, all that is missing is a comment explaining it.  I can add that.  My concern is that you are relying on `LinuxPlatformSettings` producing the right results for a non-Linux platform, so without such a comment, or possibly even with it, continued success depends on `LinuxPlatformSettings` not doing anything that's wrong for OpenBSD6.  On the other hand, since those results are right for you, for now, it's minimally disruptive to allow that quirk.

similar/editor/med.cpp
thfrwn: Build failed without the change because of msgbox_warning which I think is Windows specific?? I fixed the logic locally, stay tuned for updated pull request.

vLKp:
Sorry about your wasted work writing a40ae65.  I had already written my counterpart (e7aedff) before I refreshed this view and saw your new commit.

thfrwn:
No problem. Only wrote it because it could have saved you time if you hadn't written it yourself yet.

PR #376, opened 2018-04-16
Update SDLMain.m

MaddTheSane:
10.9 changed how applications were launched, making the old method of detecting a Finder launch fail.

### DIFF
--- a/common/arch/cocoa/SDLMain.m
+++ b/common/arch/cocoa/SDLMain.m
@@ -351,12 +351,43 @@ static void CustomApplicationMain (int argc, char **argv)
 #endif
 
 
+static int IsRootCwd()
+{
+    char buf[MAXPATHLEN];
+    char *cwd = getcwd(buf, sizeof (buf));
+    return (cwd && (strcmp(cwd, "/") == 0));
+}
+
+static int IsTenPointNineOrLater()
+{
+    /* Gestalt() is deprecated in 10.8, but I don't care. Stop using SDL 1.2. */
+    SInt32 major, minor;
+    Gestalt(gestaltSystemVersionMajor, &major);
+    Gestalt(gestaltSystemVersionMinor, &minor);
+    return ( ((major << 16) | minor) >= ((10 << 16) | 9) );
+}
+
+static int IsFinderLaunch(const int argc, char **argv)
+{
+    const int bIsNewerOS = IsTenPointNineOrLater();
+    /* -psn_XXX is passed if we are launched from Finder in 10.8 and earlier */
+    if ( (!bIsNewerOS) && (argc >= 2) && (strncmp(argv[1], "-psn", 4) == 0) ) {
+        return 1;
+    } else if ((bIsNewerOS) && (argc == 1) && IsRootCwd()) {
+        /* we might still be launched from the Finder; on 10.9+, you might not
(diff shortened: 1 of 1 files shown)


vLKp:
@MaddTheSane @kreatordxx : what are the user-visible consequences of this detection failing?

@kreatordxx : any objection to merging this?

common/arch/cocoa/SDLMain.m
```diff
@@ -351,12 +351,43 @@ - (NSString *)stringByReplacingRange:(NSRange)aRange with:(NSString *)aString
...
+    return (cwd && (strcmp(cwd, "/") == 0));
+}
+
+static int IsTenPointNineOrLater()
+{
+    /* Gestalt() is deprecated in 10.8, but I don't care. Stop using SDL 1.2. */
+    SInt32 major, minor;
+    Gestalt(gestaltSystemVersionMajor, &major);
+    Gestalt(gestaltSystemVersionMinor, &minor);
+    return ( ((major << 16) | minor) >= ((10 << 16) | 9) );
```
vLKp: This seems like a strange way to test the version.  It'd probably generate simpler code, and definitely be more readable, to test the version components individually.

kreatordxx: I think the only way of doing that would be to use a more recent function that tells the OS version. Unfortunately the methods listed on [a StackExchange article](https://stackoverflow.com/questions/11072804/how-do-i-determine-the-os-version-at-runtime-in-os-x-or-ios-without-using-gesta) look to be even more complicated.

Currently DXX-Rebirth is supported on Mac OS X 10.6 and higher - so
`[[NSProcessInfo processInfo] operatingSystemVersion]`
is out.

kreatordxx:
I have no objections to merging as-is, I tested and it builds and runs fine.

MaddTheSane:
The user-facing issues would be that games launched in place with the data won't find the data. Worst case, it'll think the current directory is `/`.

common/arch/cocoa/SDLMain.m
MaddTheSane: This code was ripped straight out of SDL 1.2's SDLMain library.

kreatordxx:
Yes, I tested with d2x.ini disabled and could reproduce this error. It was failing to add the app binary directory to the search path, causing it to abort with the 'hog file not found' error message.

common/arch/cocoa/SDLMain.m
vLKp: If this is a backport of upstream's fix, that is an excellent reason to leave it as-is.  I'm inclined to track their code absent a compelling reason to deviate (e.g. if it caused a regression on a supported platform - which does not apply here).

vLKp:
Yes, it will affect the [commit tagged 0.60.0-beta2](/dxx-rebirth/dxx-rebirth/commit/8a16296eb4aa50ff086eacd7b30cff7a81be456f), but the [fix is backported](/dxx-rebirth/dxx-rebirth/commit/b5edc4fd8f20bacfa51a91866f45de7f673faae3) to [stable-0.60.x](/dxx-rebirth/dxx-rebirth/tree/stable-0.60.x), so the next release from the 0.60.x branch will include it.

PR #377, opened 2018-04-20
Fix Blocky Filtered option.

roncli:
I initially found there was a problem with Blocky Filtered in this line of code:

```gl_mag_filter_int = GL_LINEAR_MIPMAP_LINEAR;```

The problem with this code is that this is an invalid value for magnifying filters, because mipmaps don't make sense for magnifying filters, only minifying filters.  In fact, this is probably creating an OpenGL error, but Rebirth does not appear to catch any OpenGL errors, so the program will just skip over setting that value and either use the default, `GL_LINEAR`, or whatever was last used, which most likely will be `GL_NEAREST`.

The solution was not obvious... was this a typo or a copy/paste error from the patch?  So I went to the source, the original patch found at https://forum.dxx-rebirth.com/showthread.php?tid=874.  To quote forum user beware:

> a feature i call "blocky filtered". textures are upscaled 4x, and then rendered trilinear/anisotropic. so the textures have square pixels, but the image has no aliasing artefacts, regardless of if textures are nearby or far away or at a steep angle.

Based on this, it appears the correct course of action is to use the same filters for Blocky Filtered as are used for Trilinear.  In the code, this essentially means to use the same code block for the two rendering modes, and that is the solution used for the PR.

### DIFF
--- a/similar/arch/ogl/ogl.cpp
+++ b/similar/arch/ogl/ogl.cpp
@@ -1713,10 +1713,6 @@ static int ogl_loadtexture(const palette_array_t &pal, const uint8_t *data, cons
 			}
 			break;
 		case OGL_TEXFILT_UPSCALE: // Upscaled - i.e. Blocky Filtered (Bilinear)
-			gl_mag_filter_int = GL_LINEAR_MIPMAP_LINEAR;
-			gl_min_filter_int = GL_LINEAR_MIPMAP_LINEAR;
-			buildmipmap = true;
-			break;
 		case OGL_TEXFILT_TRLINEAR: // Smooth - Trilinear
 			gl_mag_filter_int = GL_LINEAR;
 			gl_min_filter_int = GL_LINEAR_MIPMAP_LINEAR;


vLKp:
What are the user-visible consequences of this problem?  Any suggestions for how, if at all, to document this fix so that impacted users can recognize that this will fix their issue?

The change looks fine.  I will try to merge it soon.

derhass:
Since the default magnification filter is `GL_LINEAR` anyway, and the attempt to change it to `GL_LINEAR_MIPMAP_LINEAR` will just result in a GL error with no other effect whatsoever, there will be no user-visible consequences of this patch at all. All it does is cleaning up an invalid use of the GL. I don't think this needs to be documented for end users.

roncli:
True, but `GL_NEAREST` is used on Blocky Filtered for large textures, and I don't know if OpenGL would continue to use that for future textures or not.  If there is a visual change, it can be simply be documented as "corrected implementation of Blocky Filtered", as that's pretty much all this fix does. :)

derhass:
> and I don't know if OpenGL would continue to use that for future textures or not.

No, the filter modes are state which are managed per texture object in the GL (like everything else set with `glTexParameter`). In the code in question,  a new texture object is created directly before (`glGenTextures` + `glBindTexture`), so it is guaranteed to be in `GL_LINEAR` state at this point in time, and it does not influence any other texture objects.

roncli:
Good to know, so no visible change, just that the PR will no longer contain an OpenGL error.

vLKp:
@Mako88: OK to put in 0.60?  Per comments from roncli and derhass, this has been silently wrong up to now.

Mako88:
@vLKp Yep! Go ahead an put it in!

vLKp:
Pushed to master as e9d36a29c1bd6a6a00e90f54b4650c7ded72b664.

Cherry-picked to stable-0.60.x as 419f4f5f6b3699d88dc57d14a554463faa97885d.

Nothing further to do here, I think.  Closing.  Thanks to both of you for the combination of report, solution, and analysis of user impact.

PR #381, opened 2018-04-22
change SYNC_GL_AUTO mode to disable GL sync if VSync is turned off

derhass:
Users with disabled VSync might not expect any forced waiting on the GPU,and the GL sync methods were intented to fix issues with VSync only, so the new heuristic for `SYNC_GL_AUTO` is to enable GL sync only if VSync is enabled, too.

Users can still request to use a specific GL sync method via the `-gl_syncmethod` switch, independent of the VSync setting.

### DIFF
--- a/common/arch/ogl/ogl_sync.cpp
+++ b/common/arch/ogl/ogl_sync.cpp
@@ -14,6 +14,7 @@
 #include <SDL.h>
 
 #include "args.h"
+#include "config.h"
 #include "console.h"
 #include "maths.h"
 #include "ogl_sync.h"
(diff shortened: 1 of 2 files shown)


vLKp:
Merged as [change SYNC_GL_AUTO mode to disable GL sync if VSync is turned off](/dxx-rebirth/dxx-rebirth/commit/38815ba10bc60b06b20edb76086636f46fbba1ae) with a small whitespace modification to reduce number of changed lines.

vLKp:
@Mako88: do you want this in 0.60?  It looks safe to me.  It could be argued to be a bugfix, and it's definitely intended to reduce user surprise.

PR #382, opened 2018-04-22
handle multiplayer network packets during SYNC_GL_FENCE_SLEEP wait loop

derhass:
This mirrors the logic from `calc_frame_time()`. When VSync is enabled, waiting for the previous frame to complete might induce long wait periods, up to the complete frame time, so we should also handle multiplayer there.

For the other sync modes, the GL calls will simply block, so we can't do anything about that (except going multithreaded, but that's a totally different can of worms). Note that even without sync, the `SwapBuffers() `call in `gr_flip()` may also block if VSync is on (and enough frames are queued up), so the issue is not the GL sync code itself.

`SYNC_GL_FENCE_SLEEP` is now probably the nicest mode for multiplayer with VSsync, as it is the only one which has the potential to continue handling multiplayer packets during the wait for VBlank time.

### DIFF
--- a/common/arch/ogl/ogl_sync.cpp
+++ b/common/arch/ogl/ogl_sync.cpp
@@ -15,7 +15,9 @@
 
 #include "args.h"
 #include "console.h"
+#include "game.h"
 #include "maths.h"
+#include "multi.h"
 #include "ogl_sync.h"
 #include "timer.h"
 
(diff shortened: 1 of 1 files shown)


vLKp:
@Mako88: do you want this in 0.60?  It looks safe, and the rationale is good.

derhass:
Just my two cents:

> do you want this in 0.60?

Since the complete "handle multiplayer packets during longer waits" scheme is new in 0.60 (compared to latest previous release 0.58.1), I'd argue that the above change is just a bug-fix, since this particular wait spot was missed before. However, while the changes look benign, I did not do extended playtesting, so I would not insist in including this into 0.60. YMMV.

Mako88:
@vLKp I think it should be included in 0.60 since as @derhass says, it really is a bugfix. :)

PR #383, opened 2018-04-23
Fix FPS.

### LINKED ISSUE #378: Beta 2: FPS issues
On Windows, I am experiencing an FPS hovering between 185 and 190, dependent on how much is getting rendered.  It almost seems like something happened with the FPS counter that is preventing it from going the full 200.

It also sounds like I'm not alone, I was talking to @Mako88 earlier about it and he mentioned there were people reporting this as well.

I haven't tried a whole lot of things to see what happens, please let me know if you want me to try anything.

roncli:
This fixes #378.

The issue is that `last_timer_value` was getting set to the `timer_value` at the conclusion of every frame.  This made the assumption that the frame time was going to run exactly at the desired value for the target FPS, and in fact that assumption was true up until a04aa340c50456630cd8ba52774702c8c751e94f.

This fixes the issue by not making that assumption, and instead incrementing `last_timer_value` by the `bound`, which is our target frame time.  In the case where our frame actually runs over the target frame time, we revert to previous behavior of just setting the `last_timer_value` to `timer_value` since our FPS will be off anyway.

### DIFF
--- a/similar/main/game.cpp
+++ b/similar/main/game.cpp
@@ -124,6 +124,7 @@ COPYRIGHT 1993-1999 PARALLAX SOFTWARE CORPORATION.  ALL RIGHTS RESERVED.
 #include "segiter.h"
 
 static fix64 last_timer_value=0;
+static fix64 sync_timer_value=0;
 fix ThisLevelTime=0;
 
 grs_canvas	Screen_3d_window;							// The rectangle for rendering the mine to
@@ -477,14 +478,13 @@ void calc_frame_time()
 	{
 		const auto timer_value = timer_update();
 		FrameTime = timer_value - last_timer_value;
-		if (FrameTime >= bound)
+		if (timer_value - sync_timer_value >= bound)
 		{
-			// For constant target FPS rates, we increase the last timer value by the bound.
-			last_timer_value += bound;
+			last_timer_value = timer_value;
 
-			// If this frame ran long, we're not going to meet our target FPS, so just set the last timer value to the current timer value.
-			if (last_timer_value < timer_value) {
-				last_timer_value = timer_value;
+			sync_timer_value += bound;
+			if (sync_timer_value + bound < timer_value) {
+				sync_timer_value = timer_value;
 			}
 			break;
 		}
--- a/similar/main/game.cpp
+++ b/similar/main/game.cpp
@@ -478,7 +478,7 @@ void calc_frame_time()
 	{
 		const auto timer_value = timer_update();
 		FrameTime = timer_value - last_timer_value;
-		if (timer_value - sync_timer_value >= bound)
+		if (FrameTime > 0 && timer_value - sync_timer_value >= bound)
 		{
 			last_timer_value = timer_value;
 
--- a/similar/main/game.cpp
+++ b/similar/main/game.cpp
@@ -479,7 +479,13 @@ void calc_frame_time()
 		FrameTime = timer_value - last_timer_value;
 		if (FrameTime >= bound)
 		{
-			last_timer_value = timer_value;
+			// For constant target FPS rates, we increase the last timer value by the bound.
+			last_timer_value += bound;
+
+			// If this frame ran long, we're not going to meet our target FPS, so just set the last timer value to the current timer value.
+			if (last_timer_value < timer_value) {
+				last_timer_value = timer_value;
+			}
 			break;
 		}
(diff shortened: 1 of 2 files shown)


Mako88:
@vLKp If this looks good to you, I'd say include it in 0.60 as well, since I've had several people point out that the FPS drop is weird.

derhass:
I have to admit that I don't understand this patch (maybe I'm missing something), but since `FrameTime timer_value - last_timer_value`, the `if` condition will be equivalent to `if (timer_value - last_timer_value >= bound)`. So there are two cases now:

1. equality: we end up with increasing `last_timer_value` by `bound`, which is the same as setting `last_timer_value=timer_value` in this case, or
2. greater than: we first increase `last_timer_value` by `bound`, and then taking the inner if and setting `last_timer_value=timer_value`.

So to me, this appears to be just a  more complicated way of writing the original `last_timer_value=timer_value`. As I said, I might miss something here, but  when I actually try out the code, I also see no observable difference in the fps counter (which is at least consistent with my understanding of the patch).

roncli:
Apologies for the first commit on this PR.  I tested it, but improperly.  Thanks for pointing the issue out, this time I was sure to test it better. :)

In rewriting the PR, I noticed the FPS counter was wrong, so I tried to fix that, too.  It can still be wonky and oscillate between 199, 200, 201, 200, etc, but at least it's not incorrectly showing a constant 201.

vLKp:
@derhass: thanks.  I wrote a similar critique, then apparently signed off without pressing the button to start the review, so no one could see my questions.

@roncli: thanks for working on this.  It may be purely cosmetic, but I'd be surprised if a release that lacked your fix didn't promptly get someone inquiring about the change.  I'd like to take a few minutes to play with it before merging, so this probably will not go in today, but I'll try to get it in soon.

vLKp:
The first thing that hit me when I tried this was a fatal assert that `FrameTime > 0`.  Prior to this change, that could only happen if `bound <= 0`.  With this change, it can happen when `timer_value - sync_timer_value >= bound` and `timer_value - last_timer_value <= 0`.  I applied the below diff to eliminate that problem by ensuring that non-timing events that change `last_timer_value` also update `sync_timer_value`, but I am unsure whether it is the right fix.  It made the assertion cease.

Since you worked in this area, any feedback would be appreciated.  If you agree that this is the right fix, I will include it in the merge.

```diff
diff --git a/similar/main/game.cpp b/similar/main/game.cpp
index 7f9e1ac62..148f758bc 100644
--- a/similar/main/game.cpp
+++ b/similar/main/game.cpp
@@ -380,6 +380,7 @@ void game_world_time_paused::increase_pause_count()
 		if (last_timer_value < 0) {
 			last_timer_value = 0;
 		}
+		sync_timer_value = last_timer_value;
 	}
 	time_paused++;
 }
@@ -391,6 +392,7 @@ void game_world_time_paused::decrease_pause_count()
 	if (time_paused==0) {
 		const fix64 time = timer_update();
 		last_timer_value = time - last_timer_value;
+		sync_timer_value = last_timer_value;
 	}
 }
 
@@ -463,6 +465,7 @@ void calc_d_tick()
 void reset_time()
 {
 	last_timer_value = timer_update();
+	sync_timer_value = last_timer_value;
 }
 
 }
```

You may find the below useful for dumping out the state of various FPS related variables.  Be aware that once in game, if you do not crash, it is *very* noisy.

```diff
@@ -478,6 +481,7 @@ void calc_frame_time()
 	{
 		const auto timer_value = timer_update();
 		FrameTime = timer_value - last_timer_value;
+		con_printf(CON_URGENT, "FrameTime=%u timer_value=%lu last_timer_value=%lu sync_timer_value=%lu bound=%u", FrameTime, timer_value, last_timer_value, sync_timer_value, bound);
 		if (timer_value - sync_timer_value >= bound)
 		{
 			last_timer_value = timer_value;
```

roncli:
Good catch on the pause, although I have not run into this one yet.  I think I am not getting it because of how the Windows build compiles.  I'll sidebar that topic with AFP on Discord since he put the compile instructions together.

I believe the easiest way to deal with this is to just add `FrameTime > 0` to the check rather than try to catch everywhere `last_timer_value` got updated.

vLKp:
I think I did get all the places it was updated.  However, testing FrameTime is fairly cheap and likely more robust in the presence of future maintenance, so I will merge your change instead of using the add-on patch I posted above.

PR #400, opened 2018-08-21
Gobble up -psn command line options

### LINKED ISSUE #387: Crash: Unhandled argument "-psn_0_1831359" while processing <command line>
### Environment
Commit c4417aafb1fff5ab5b703a340a24d741649bc389.

* Mac OS X
  * [x] Mac OS X 10.13

#### CPU environment

* [x] x86\_64 (64-bit Intel/AMD)

### Description
When DXX-Rebirth is double clicked in the Finder, I get the error message "similar/misc/args.cpp:478: Unhandled argument "-psn_0_1831359" while processing \<command line\>".

#### Regression status

After running `git bisect` the commit c4417aafb1fff5ab5b703a340a24d741649bc389 was identified to be the cause. This is pull request #376.

### Steps to Reproduce
1. Launch D1X-Rebirth or D2X-Rebirth from the Finder.

* Engines affected:
  * [x] D1X-Rebirth
  * [x] D2X-Rebirth

What would be the best way to handle this argument?

MaddTheSane:
Should fix #387.

### DIFF
--- a/similar/misc/args.cpp
+++ b/similar/misc/args.cpp
@@ -400,6 +400,10 @@ static void ReadCmdArgs(Inilist &ini, Arglist &Args)
 				throw nesting_depth_exceeded();
 			ReadIniArgs(ini);
 		}
+		else if (!strncmp(p, "-psn", 4))
+		{
+			//do nothing/gobble it up
+		}
 		else
 			throw unhandled_argument(std::move(*pp));
 	}
--- a/similar/misc/args.cpp
+++ b/similar/misc/args.cpp
@@ -400,10 +400,12 @@ static void ReadCmdArgs(Inilist &ini, Arglist &Args)
 				throw nesting_depth_exceeded();
 			ReadIniArgs(ini);
 		}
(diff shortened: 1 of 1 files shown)


vLKp:
Although #387 veered off into discussing sharepath, I stand by my original statement that this option should never reach the common code, so the common code should not deal with it.  If it cannot be easily suppressed in the OS X specific code, then it should at least be wrapped in a preprocessor guard to compile out the code on non-OS X platforms.

MaddTheSane:
So find out why the `-psn` argc is getting passed to `InitArgs`. Got it.

vLKp:
> So find out why ...

At the time, I said that, and I would still like to understand the cause.  However, after I said that, I learned that the file in question is mostly or exactly copied from SDL, and I am reluctant to change a bundled file without very good cause.  As such, I went ahead and merged your commits from this pull as-is.

If you want to research why the argument list isn't reset, I'd like to hear your findings.  If you don't want to deal with it, that's fine too.

PR #414, opened 2018-12-25
Fix build on FreeBSD

pkubaj:
FreeBSD needs those patches to build dxx-rebirth.

### DIFF
--- a/similar/editor/med.cpp
+++ b/similar/editor/med.cpp
@@ -890,7 +890,7 @@ static void close_editor()
 {
 	//	_MARK_("end of editor");//Nuked to compile -KRB
 	
-#if !defined(__linux__) && !defined(__OpenBSD__)
+#if !defined(__linux__) && !defined(__OpenBSD__) && !defined(__FreeBSD__)
 	set_warn_func(msgbox_warning);
 #else
 	clear_warn_func();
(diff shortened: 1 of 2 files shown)


vLKp:
Could you show how the build fails without these patches?  Is this really the best solution, since it seems like you need the same special-case logic as OpenBSD?  Wouldn't it be better to test for `is-any-BSD` instead of testing `is-OpenBSD || is-FreeBSD`?  The BSD variants probably all need the same rules, and I'd rather not keep revisiting this every time a new BSD variant comes up.  Wikipedia lists DragonFly, Free, Net, and Open as major BSDs, and a long list of minor variants off those.  I would rather test for a single preprocessor symbol that they all define, instead of calling them out individually.

vLKp:
Sorry, I was unclear.  I wanted your input on how to handle all the BSDs without enumerating them individually.  As far as I can tell from my reading, there is no good way for the C preprocessor to identify all BSDs as a group.

The link failure is due to a mismatch in how the function was declared/used/defined.  I reworked that to synchronize those cases, which should fix the failure.  Please sync to tip and retest.

vLKp:
I believe my changes will address the problem your patch was intended to resolve, and will avoid the need to keep extending this blacklist.  Please reopen or file an issue if tip still does not work for you.

PR #422, opened 2019-05-24
Create README.md

fskirschbaum:
When reading [announcement](https://www.dxx-rebirth.com/news/dxx-rebirth-rebooted/) it referenced the README file for reporting issues but there isn't a root README...  and the ISSUE_TEMPLATE.md wasn't referenced by the README files within the sub-folders... and generally it felt like there should be a README

... every project should have an up front README file.  (This is one.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ DXX-Rebirth is based on the late D1X and D2X source ports (which, in turn, were 
 
 It is the goal of DXX-Rebirth to keep Descent 1 & 2 alive and well, updating them for modern PCs while also keeping them the same games you remember playing back in 1995!
 
-If you have any questions, comments, or suggestions, contact zico [at] dxx-rebirth [dot] com, or come to our forum: http://dxx-rebirth.com/frm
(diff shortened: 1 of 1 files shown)


vLKp:
`ISSUE_TEMPLATE.md` doesn't need to be referenced by the README, because it is primarily for the benefit of the Github Issue Tracker to pre-fill the issue text.  This project's README files migrated down into subdirectories when the source repositories were unified.

- The forum link changed after the README files you used as a reference.
- Github's markdown parser seems to want to do the wrong thing with the inline \* in one of the lists.

Thanks for reminding me that this needs to be done.  I'll clean up the minor points above and get this merged.

fskirschbaum:
Happy to get the ball rolling.

For what it's worth, the `ISSUE_TEMPLATE.md` is old style so an update to the template structure might be worth looking into as well: https://help.github.com/en/articles/about-issue-and-pull-request-templates

vLKp:
Thank you for the link.  I wasn't aware that the format used here was deprecated, nor that Github had a proposed successor.  I've split the template accordingly.

PR #426, opened 2019-05-27
Fix build on FreeBSD

pkubaj:


similar/main/digiobj.cpp
```diff
@@ -578,7 +578,7 @@ void digi_sync_sounds()
                                 &s.volume, &s.pan, s.max_distance );
 
 			} else if ( s.flags & SOF_LINK_TO_OBJ )	{
-				const auto objp = [&vcobjptr, &s]{
+				const auto objp = [&s]{
```
vLKp: NAK.  This line is required by <gcc-8.  If this change is merged, <gcc-8 fail to build this file.

I will push a configure test to suppress the clang warning in this case, since it actively encourages the reader to break other compilers.

similar/main/object.cpp
```diff
@@ -1057,7 +1057,7 @@ static void free_object_slots(uint_fast32_t num_used)
 
 	// Capture before num_to_free modified
 	const auto &&r = partial_const_range(obj_list, num_to_free);
-	auto l = [&vmobjptr, &r, &num_to_free](bool (*predicate)(const vcobjptr_t)) -> bool {
+	auto l = [&r, &num_to_free](bool (*predicate)(const vcobjptr_t)) -> bool {
```
vLKp: NAK, same problem as in the other commit.  This capture is required for <gcc-8.

vLKp:
Thank you for the report.  Several of the proposed changes cause problems elsewhere, so I resolved those differently.  Please fetch and report whether the FreeBSD build now works for you without modification.

vLKp:
I believe the changes I made above will address the problems that motivated this pull request.  No further response from original poster.

@pkubaj : please reopen if the FreeBSD build does not work for you with the changes I pushed.

PR #439, opened 2019-07-27
Dim the screen when cloak is picked up

ziplantil:
When picking up a cloaking device in Descent 1 or 2 on DOS, the screen dims briefly. This was not present on Rebirth after conversion to OpenGL because of how the palette effect is applied; any negative values, if even given, would simply get clamped by GL, preventing them from having any effect.

Testing this can be done on any Descent level with a cloaking device, such as level 5 in Descent 1 (a cloak is present in the reactor room). On the DOS version, the screen noticeably dims when it is picked up, while this does not happen on Rebirth, giving the player fewer visual cues that they have activated a cloaking device.

This PR re-introduces the effect by modifying the OpenGL code. First, gr_palette_step_up is modified to accept negative values and pass them to GL via ogl_do_palfx, which is itself modified to handle the cloak special case when all of the overlaid RGB values are negative. This requires changing the GL filter to GL_ONE_MINUS_SRC_COLOR, which effectively subtracts the color from the screen instead of adding it in.

This did have one drawback - the dimming effect was considerably fainter than on the DOS version, so I added a magic number by multiplying the values by 2.5, which makes the intensity of the effect now match the DOS version.

Note that getting damage also calls gr_palette_step_up with negative values, but not solely negative values, in which case the code changes included by this PR should have no effect. Only positive values, such as those applied by most powerups, should also be unaffected.

### DIFF
--- a/similar/arch/ogl/gr.cpp
+++ b/similar/arch/ogl/gr.cpp
@@ -75,7 +75,9 @@
 
 #include "compiler-make_unique.h"
 
+using std::min;
 using std::max;
+using std::abs;
 
 namespace dcx {
 
@@ -1042,11 +1044,22 @@ void ogl_ulinec(grs_canvas &canvas, const int left, const int top, const int rig
 }
 
 static GLfloat last_r, last_g, last_b;
+static GLfloat alast_r, alast_g, alast_b;
 static int do_pal_step;
 
 void ogl_do_palfx(void)
 {
-	GLfloat color_array[] = { last_r, last_g, last_b, 1.0, last_r, last_g, last_b, 1.0, last_r, last_g, last_b, 1.0, last_r, last_g, last_b, 1.0 };
+	// scale negative effect by 2.5 to match D1/D2 on GL
+	alast_r = abs(last_r) * (last_r < 0 ? 2.5 : 1.0);
+	alast_g = abs(last_g) * (last_g < 0 ? 2.5 : 1.0);
+	alast_b = abs(last_b) * (last_b < 0 ? 2.5 : 1.0);
+
+	GLfloat color_array[] = { 
+		alast_r, alast_g, alast_b, 1.0,
+		alast_r, alast_g, alast_b, 1.0,
+		alast_r, alast_g, alast_b, 1.0,
+		alast_r, alast_g, alast_b, 1.0
+	};
 
 	OGL_DISABLE(TEXTURE_2D);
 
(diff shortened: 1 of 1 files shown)


similar/arch/ogl/gr.cpp
```diff
@@ -1042,11 +1044,22 @@ void ogl_ulinec(grs_canvas &canvas, const int left, const int top, const int rig
 }
 
 static GLfloat last_r, last_g, last_b;
+static GLfloat alast_r, alast_g, alast_b;
```
vLKp: I think these could be function-local, rather than `static`.

similar/arch/ogl/gr.cpp
```diff
@@ -1082,9 +1098,20 @@ void gr_palette_step_up(int r, int g, int b)
 	old_b_g = ogl_brightness_g;
 	old_b_b = ogl_brightness_b;
 
-	ogl_brightness_r = max(r + gr_palette_gamma, 0);
-	ogl_brightness_g = max(g + gr_palette_gamma, 0);
-	ogl_brightness_b = max(b + gr_palette_gamma, 0);
+	if (r >= 0)
```
vLKp: This block looks like it would benefit from an inline function to reduce the repetition.

similar/arch/ogl/gr.cpp
vLKp: I think these could be function-local, rather than `static`.

similar/arch/ogl/gr.cpp
```diff
@@ -1049,30 +1049,38 @@ static int do_pal_step;
...
 	glEnableClientState(GL_COLOR_ARRAY);
  
-	if (do_pal_step)
+	if (do_pal_step && ((last_r <= 0) & (last_g <= 0) & (last_b <= 0))) {
+		// scale negative effect by 2.5 to match D1/D2 on GL
+		alast_r = abs(last_r) * 2.5;
+		alast_g = abs(last_g) * 2.5;
+		alast_b = abs(last_b) * 2.5;
+
+		color_array[ 0] = color_array[ 4] = alast_r;
```
vLKp: This does the right thing, but looks odd at first.  I'd prefer to order the logic so that the array is initialized with the right values, rather than patching it later.

zicodxx:
@vLKp Not at all. Granted I am a bit rusty with the code but I don't see any obvious red flags, and @ziplantil posted a comparison video and it looks really good. Me having never looked into this is just because - in all honesty - I straight out forgot that this effect ever existed.

PR #440, opened 2019-07-28
Add black shape to cover exit hatch space

ziplantil:
In the original DOS version, the mine keeps rendering even after the large explosion that "destroys" the exit hatch. For all well-lit mines (including all mines in Descent 1) that have a completely dark end of the exit tunnel, the level will simply show up as black.

In Rebirth however, the mine stops rendering once the large explosion happens. Therefore, currently, without this change, once the exit hatch is destroyed, the terrain texture will instead show through the hatch entrance, which looks fairly awkward. 

The mine stops rendering because the `Render_depth` is gradually decreased under `render_external_scene`. Disabling this would cause the mine to be rendered, but it would appear on top of the terrain regardless of whether it should for a given pixel. This could be addressed with more precise rendering, particularly to figure out a good Z order, but since all well-lit mines would display darkness anyway, the following solution is easier and produces a very similar result.

The solution implemented by this PR is to add a black quad which is rendered close to the exit hatch entrance instead. It is only rendered once `mine_destroyed` is true, or when the exit hatch will appear to be destroyed. I specifically set the proportions so that it would not poke past the hatch visually, but also covers its entire visual area.

While rendering a black quad is a less ideal solution than managing to make the mine render and not obstruct the terrain except for the exit hatch, this is still more ideal than how the hatch currently appears after the explosion.

### DIFF
--- a/similar/main/endlevel.cpp
+++ b/similar/main/endlevel.cpp
@@ -139,6 +139,7 @@ static void draw_stars(grs_canvas &, const d_unique_endlevel_state::starfield_ty
 static int find_exit_side(const object_base &obj);
 static void generate_starfield(d_unique_endlevel_state::starfield_type &stars);
 static void start_endlevel_flythrough(flythrough_data *flydata,const vmobjptr_t obj,fix speed);
+static void draw_mine_exit_cover(grs_canvas &);
 
 #if defined(DXX_BUILD_DESCENT_II)
 constexpr array<const char, 24> movie_table{{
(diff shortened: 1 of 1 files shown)


ziplantil:
Before the change:
exithatchfix_before

After the change:
exithatchfix_after

zicodxx:
I like this. Considering how tricky the sorting of this scene is otherwise, this seems to be like a really faithful solution.

vLKp:
I'll merge this next time I work on the code.

PR #441, opened 2019-08-13
Reload sounds when loading D2 mission

ziplantil:
This change causes game sounds to be reloaded and read from disk again every time a new mission is loaded when playing Descent 2 (D2X), allowing missions to have custom .s11/.s22 files within the HOG and therefore do sound changes. Granted, the s11/s22 files have to be named descent2.s11 and descent2.s22, but this is better than nothing before where the only way to change the sounds would be to use a dxa or replace the game files manually.

This change should have no effect on D1X.

Notes:
* This could be optimized by reloading the sounds only if we either detect a custom sound file or the last mission had one (so that sounds would reset when going back to the builtin D2 mission, for instance)
* I also added a comment to read_hamfile to specify that it is _meant_ to read the .HAM from the .HOG, since I assume this is the case.

### DIFF
--- a/common/main/piggy.h
+++ b/common/main/piggy.h
@@ -165,6 +165,7 @@ extern void remove_char( char * s, char c );	// in piggy.c
 #define REMOVE_COMMENTS(s)	remove_char((s),';')
 #define REMOVE_DOTS(s)  	remove_char((s),'.')
 
+extern int Piggy_hamfile_version;
 extern unsigned Num_bitmap_files;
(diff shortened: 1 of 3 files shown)


vLKp:
The sound loading code looks like it could use some other optimizations, too.  It looks like it closes and reopens the raw sound file, which is wasteful.  I merged your change, with a couple of minor cleanups committed on top of it.  Optimizations can be done later.

PR #446, opened 2019-08-22
Read previous_best_levelnum even if rotating

ziplantil:
This is a simple fix for set_highest_level that should hopefully fix an elusive issue regarding the highest level info being set to wrong values. I haven't been able to come up with a reproducible list of steps to get this to occur on Rebirth as it is right now (but I have encountered it while testing my mission in progress), so I chose to instead make this simple solution for what seems to be the issue.

### DIFF
--- a/similar/main/playsave.cpp
+++ b/similar/main/playsave.cpp
@@ -1258,22 +1258,31 @@ void set_highest_level(int levelnum)
 			PlayerCfg.NHighestLevels++;
 		ii->Shortname.back() = 0;
 		strncpy(ii->Shortname.data(), Current_mission_filename, ii->Shortname.size() - 1);
+		// make sure we overwrite it in the following code
+		ii->LevelNum = 0;
 	}
 	else if (ii != ie - 1)
 	{
 		/* If this mission is not the most recently used, reorder the
 		 * list so that it becomes the most recently used.
 		 */
-		previous_best_levelnum = ii->LevelNum;
 		std::rotate(ii, std::next(ii), ie);
 		ii = ie - 1;
 	}
 	else
 		previous_best_levelnum = ii->LevelNum;
 
+  // only use previous_best_levelnum to determine if we need to save again
 	if (previous_best_levelnum < levelnum)
 	{
-		ii->LevelNum = levelnum;
+		/*
+		 * make sure we don't actually send the player back from the
+		 * previously highest level. in case of rotation, the previous
+		 * level number is never checked, which can cause the player to not
+		 * be able to choose as late into the mission as they should.
+		 */
+	 
+		ii->LevelNum = std::max(ii->LevelNum, static_cast<uint8_t>(levelnum));
 		write_player_file();
 	}
 }
--- a/similar/main/playsave.cpp
+++ b/similar/main/playsave.cpp
@@ -1264,6 +1264,7 @@ void set_highest_level(int levelnum)
 		/* If this mission is not the most recently used, reorder the
 		 * list so that it becomes the most recently used.
 		 */
+		previous_best_levelnum = ii->LevelNum;
 		std::rotate(ii, std::next(ii), ie);
 		ii = ie - 1;
 	}


vLKp:
Could you explain the motivating problem in more detail?  As is, I don't see why this change would improve the situation.  As I interpret this code path:

- https://github.com/dxx-rebirth/dxx-rebirth/blob/211a172c1dc5f939a3ed387bea57b1d0ee679192/similar/main/playsave.cpp#L1243
  Set `previous_best_levelnum`
- https://github.com/dxx-rebirth/dxx-rebirth/blob/211a172c1dc5f939a3ed387bea57b1d0ee679192/similar/main/playsave.cpp#L1244
  Assume untaken, since we are discussing the path that uses `std::rotate`.
- https://github.com/dxx-rebirth/dxx-rebirth/blob/211a172c1dc5f939a3ed387bea57b1d0ee679192/similar/main/playsave.cpp#L1262
- https://github.com/dxx-rebirth/dxx-rebirth/blob/211a172c1dc5f939a3ed387bea57b1d0ee679192/similar/main/playsave.cpp#L1267
  According to https://en.cppreference.com/w/cpp/algorithm/rotate, this use of `std::rotate`:
```
Performs a left rotation on a range of elements.
Specifically, std::rotate swaps the elements in the range [first, last) in such a way that the element n_first becomes the first element of the new range and n_first - 1 becomes the last element.
```
  Applying that to this call, the element (mission) that was leftmost becomes rightmost.  Then:
- https://github.com/dxx-rebirth/dxx-rebirth/blob/211a172c1dc5f939a3ed387bea57b1d0ee679192/similar/main/playsave.cpp#L1268
  The rightmost element is marked to be overwritten.
- https://github.com/dxx-rebirth/dxx-rebirth/blob/211a172c1dc5f939a3ed387bea57b1d0ee679192/similar/main/playsave.cpp#L1273
  At this point, `previous_best_levelnum` is still 0, so it is guaranteed that this branch is taken, which should cause the change to be saved.

---
[Apologies for any formatting errors.  Github's Preview feature is broken at the moment.]

ziplantil:
Let us assume we have two custom missions in our missions folder called `alpha` and `beta`. 

* We've beaten `alpha` until level 10, so when we select a level, we can select any level up to 10.
* We have a save file for `alpha` in level 6. There was something interesting about the way the level ended, so you want to check that again at some point.
* We have not played `beta` at all yet.

Therefore, the highest level info table is at ''(possible builtin missions)'', `alpha`: 10.

We now open `beta` and start playing it, beating level 1 and reaching level 2.
Thus the highest level info table is at ''(possible builtin missions)'', `alpha`: 10, `beta`: 2.

It's now that we decide to load that savefile for level 6 and beat it again to reach level 7.
The code will now notice that upon reaching level 7, the `alpha` mission is not the rightmost mission in the list,
so we take the code path `ii != ie - 1` and starts pushing it to the end of the list.

We rotate the list, and finally, `alpha` is at the end. It seems we're still on level 0 and we've reached
level 7 now, so we'll store that 7 is the highest level we've reached for `alpha`.

Wait... what about that level 10 we had earlier? The rotate code now never checks the existing highest level
we have reached; this is only done if the mission happens to already be last.

vLKp:
`previous_best_levelnum` is only used to decide whether the player file needs to be rewritten.  It is not the value that is stored into the completed levels table.  Leaving it set to 0 ensures that the final test will be true and the computed level table rewritten, even if, as in your example, the player has not reached a new high level.  I did this so that the rotation would be recorded.  With your change, `previous_best_levelnum` would be `10`, so the final branch would not execute when the new best is `7`.  Thus, the game would not record the rotation, and your choice to play this level would not push the level back to a status of "most recently played."  In turn, that would allow it to age off the end of the list sooner than the levels that are actually least recently played.

ziplantil:
The game would not save the rotation, yes, but it also wouldn't reset the highest level reached back to 7 when it was 10 before that, which is a regression from how the code worked previously. Then again, if the rotation must be saved, there is a better approach to this.

vLKp:
Good point.  I focused on the new regression (not saving on rotation), and missed the existing regression (reducing the player's maximum level).  I fixed this a bit differently in dd0f639176a667e82e318ffdc63acdeb436023b8, but I think it matches what you proposed.

PR #452, opened 2019-09-01
Added axis buttons

Daivuk:
Each axis can act as two buttons in both ways.
For example, a player might map slide left and slide right to J1 -A1 and J1 +A1 as button presses instead of the slide L/R axis.

This is mostly to fix XBox 360 Controller Left and Right triggers. But it can work on every axis if the player wishes to bind them.

### DIFF
--- a/common/arch/sdl/joy.cpp
+++ b/common/arch/sdl/joy.cpp
@@ -129,6 +129,7 @@ class d_physical_joystick
 	VERB(button_map)	\
 	VERB(axis_map)	\
 	VERB(axis_value)	\
+	VERB(axis_button_map)	\
 
 #if DXX_USE_SIZE_SORTED_TUPLE
 	template <typename... Ts>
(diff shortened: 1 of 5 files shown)


similar/main/kconfig.cpp
```diff
@@ -1114,7 +1114,7 @@ void kconfig_read_controls(const d_event &event, int automap_flag)
 				}
 			}
 			break;
-#if DXX_MAX_BUTTONS_PER_JOYSTICK || DXX_MAX_HATS_PER_JOYSTICK
+#if DXX_MAX_BUTTONS_PER_JOYSTICK || DXX_MAX_HATS_PER_JOYSTICK || DXX_MAX_AXES_PER_JOYSTICK
```
Daivuk: Hum I'll remove this || here its not going through that case.

common/arch/sdl/joy.cpp
```diff
@@ -247,6 +250,78 @@ window_event_result joy_hat_handler(SDL_JoyHatEvent *jhe)
 #endif
 
 #if DXX_MAX_AXES_PER_JOYSTICK
+window_event_result joy_axisbutton_handler(SDL_JoyAxisEvent *jae)
+{
+	auto &js = SDL_Joysticks[jae->which];
+	auto axis_value = js.axis_value()[jae->axis];
+    window_event_result highest_result(window_event_result::ignored);
```
Daivuk: will fix spaces for tabs here to be tabs like original code

common/arch/sdl/joy.cpp
```diff
@@ -247,6 +250,78 @@ window_event_result joy_hat_handler(SDL_JoyHatEvent *jhe)
...
+	// We could add another deadzone slider called "axis button deadzone".
+	// I think it's safe to assume a 30% deadzone on analog button presses for now.
+	const decltype(axis_value) deadzone = 38;
+	auto prev_value = axis_value;
+	auto new_value = jae->value/256;
+
+	if (prev_value > deadzone)
+		prev_value = ((prev_value - deadzone) * 128) / (128 - deadzone);
+	else if (prev_value < -deadzone)
+		prev_value = ((prev_value - deadzone) * 128) / (128 - deadzone);
```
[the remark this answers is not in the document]
Daivuk: You are right, it should be `((prev_value + deadzone)...`

This code is directly taken from `kconfig.cpp`
```
if (value > joy_null_value)
  value = ((value - joy_null_value) * 128) / (128 - joy_null_value);
else if (value < -joy_null_value)
  value = ((value + joy_null_value) * 128) / (128 - joy_null_value);
else
  value = 0;
```

Copy paste error.

Daivuk:
> Review summary:
> 
> * The deadzone handling for the negative case looks wrong to me.  If you think it is correct as-is, please explain the current logic.
> * Other than that, I don't see any correctness issues/concerns, just some minor refactoring/code cleanup that I can add on top of it.

requested change completed

PR #453, opened 2019-09-17
SConstruct: remove equals sign from RPI -isystem definitions

psyke83:
When targeting Raspberry Pi with legacy drivers, the following error
is observed on Raspbian buster with gcc 8.3.0:

cc1plus: error: =/opt/vc/include: No such file or directory [-Werror=missing-include-dirs]
cc1plus: error: =/opt/vc/include/interface/vcos/pthreads: No such file or directory [-Werror=missing-include-dirs]
cc1plus: error: =/opt/vc/include/interface/vmcs_host/linux: No such file or directory [-Werror=missing-include-dirs]

The error can be resolved by removing the equals sign from these definitions.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -4362,9 +4362,9 @@ class DXXCommon(LazyObjectConstructor):
 			# are not very clean and would trigger some warnings we usually consider as
 			# errors. Using them as system headers will make gcc ignoring any warnings.
 				CPPFLAGS = [
-				'-isystem=%s/include' % rpi_vc_path,
-				'-isystem=%s/include/interface/vcos/pthreads' % rpi_vc_path,
-				'-isystem=%s/include/interface/vmcs_host/linux' % rpi_vc_path,
+				'-isystem%s/include' % rpi_vc_path,
+				'-isystem%s/include/interface/vcos/pthreads' % rpi_vc_path,
+				'-isystem%s/include/interface/vmcs_host/linux' % rpi_vc_path,
 			],
 				LIBPATH = '%s/lib' % rpi_vc_path,
 				LIBS = ['bcm_host'],


psyke83:
In case more detail is needed:
Buster gcc version: `Raspbian 8.3.0-6+rpi1`
Stretch gcc version: `Raspbian 6.3.0-18+rpi1+deb9u1`

The build is targeting the legacy graphics stack for Raspberry Pi 3B via `raspberrypi=1`. This patch isn't necessary for Stretch, but it doesn't break the build for the older gcc version, so it should hopefully be acceptable.

vLKp:
According to `info gcc`:

> If DIR begins with \`=\', then the \`=\' is replaced by the sysroot prefix; see \`--sysroot\' and \`-isysroot\'.

Thus, I think the `=` would matter in cross-compilation cases.  The documentation shows a space separating the `-isystem` and `DIR`, and does not state whether that space is mandatory for proper operation.  I wonder if the version you used considers it mandatory, and so the `=` is not expanded to `sysroot`.

If you use the below patch instead, does that resolve the error for you?  This retains the `=`, but passes the path as a separate argument from the literal `-isystem`.

```diff
diff --git a/SConstruct b/SConstruct
index 653279f17..745714f79 100644
--- a/SConstruct
+++ b/SConstruct
@@ -4362,9 +4362,9 @@ class DXXCommon(LazyObjectConstructor):
 			# are not very clean and would trigger some warnings we usually consider as
 			# errors. Using them as system headers will make gcc ignoring any warnings.
 				CPPFLAGS = [
-				'-isystem=%s/include' % rpi_vc_path,
-				'-isystem=%s/include/interface/vcos/pthreads' % rpi_vc_path,
-				'-isystem=%s/include/interface/vmcs_host/linux' % rpi_vc_path,
+				'-isystem', '=%s/include' % rpi_vc_path,
+				'-isystem', '=%s/include/interface/vcos/pthreads' % rpi_vc_path,
+				'-isystem', '=%s/include/interface/vmcs_host/linux' % rpi_vc_path,
 			],
 				LIBPATH = '%s/lib' % rpi_vc_path,
 				LIBS = ['bcm_host'],
```

psyke83:
Thanks for the suggestion; unfortunately, it doesn't resolve the issue:
```
g++ -c -g -O2 -ftabstop=4 -Wall -Werror=extra -Werror=format=2 -Werror=missing-braces -Werror=missing-include-dirs -Werror=uninitialized -Werror=undef -Werror=pointer-arith -Werror=cast-qual -Werror=missing-declarations -Werror=redundant-decls -Werror=vla -pthread -funsigned-char -Werror -Wno-sign-compare -isystem =/opt/vc/include -isystem =/opt/vc/include/interface/vcos/pthreads -isystem =/opt/vc/include/interface/vmcs_host/linux -DPHYSFS_DEPRECATED= -DRPI -Icommon/include -Icommon/main -I. .sconf_temp/conftest_0.cpp -o .sconf_temp/conftest_0.o
cc1plus: error: =/opt/vc/include: No such file or directory [-Werror=missing-include-dirs]
cc1plus: error: =/opt/vc/include/interface/vcos/pthreads: No such file or directory [-Werror=missing-include-dirs]
cc1plus: error: =/opt/vc/include/interface/vmcs_host/linux: No such file or directory [-Werror=missing-include-dirs]
cc1plus: all warnings being treated as errors
```

This appears to resolve the errors:
```
diff --git a/SConstruct b/SConstruct
index bd8d095e8..485a7ea8e 100644
--- a/SConstruct
+++ b/SConstruct
@@ -4369,6 +4369,7 @@ class DXXCommon(LazyObjectConstructor):
                                '-isystem=%s/include' % rpi_vc_path,
                                '-isystem=%s/include/interface/vcos/pthreads' % rpi_vc_path,
                                '-isystem=%s/include/interface/vmcs_host/linux' % rpi_vc_path,
+                               '--sysroot /',
                        ],
                                LIBPATH = '%s/lib' % rpi_vc_path,
                                LIBS = ['bcm_host'],
```
I presume this is not the proper fix either, as hardcoding the root path in this fashion may also break cross-compile or chroot builds. I'm unsure how to proceed.

vLKp:
Yes, I'd rather not force a sysroot.  I could add a new SConf option for the user to set a sysroot, and allow that to be passed through, then use the `=` only when a sysroot is set.  When sysroot is unset, use `-isystem` without an `=`.  Another option is to go with your approach, accept that it is not compatible with cross-compiling to a Raspberry Pi, and ask that people who care about cross-compilation not use the `raspberrypi` convenience switch, and instead explicitly set the include and library directories they need.  Rebirth can be built for the Pi without using the `raspberrypi` option; it's just more convenient to use that option because it sets various options to good defaults.

----

I'm not inclined to remove `raspberrypi` as a configuration option, but I am now doubtful that having a special switch for one target is the right way to handle this.  It would have been better to have a `target=FOO`, where `raspberrypi` is one of several valid choices for `FOO`.  This would avoid polluting the top level configuration space to support platforms that need special help to find their headers/libraries.

----

Untested, but a demonstration of what I was thinking for the sysroot SConf option:
```diff
diff --git a/SConstruct b/SConstruct
index bd8d095e8..d3d6a2d6b 100644
--- a/SConstruct
+++ b/SConstruct
@@ -3672,6 +3672,7 @@ class DXXCommon(LazyObjectConstructor):
 			{
 				'variable': generic_variable,
 				'arguments': (
+					('sysroot', None, None),
 					('rpi_vc_path', self.RPI_DEFAULT_VC_PATH, 'directory for RPi VideoCore libraries'),
 					('opengles_lib', self.selected_OGLES_LIB, 'name of the OpenGL ES library to link against'),
 					('egl_lib', self.selected_EGL_LIB, 'name of the OpenGL ES Graphics Library to link against'),
@@ -4360,15 +4361,20 @@ class DXXCommon(LazyObjectConstructor):
 		if user_settings.raspberrypi == 'yes':
 			rpi_vc_path = user_settings.rpi_vc_path
 			message(self, "Raspberry Pi: using VideoCore libs in %r" % rpi_vc_path)
+			rpi_inc = ('=' if user_settings.sysroot else '', rpi_vc_path)
+			if user_settings.sysroot:
+				env.Append(CPPFLAGS = [
+					'--sysroot=%s' % user_settings.sysroot
+					])
 			env.Append(
 				CPPDEFINES = ['RPI'],
 			# use CPPFLAGS -isystem instead of CPPPATH because these those header files
 			# are not very clean and would trigger some warnings we usually consider as
 			# errors. Using them as system headers will make gcc ignoring any warnings.
 				CPPFLAGS = [
-				'-isystem=%s/include' % rpi_vc_path,
-				'-isystem=%s/include/interface/vcos/pthreads' % rpi_vc_path,
-				'-isystem=%s/include/interface/vmcs_host/linux' % rpi_vc_path,
+					'-isystem', '%s%s/include' % rpi_inc,
+					'-isystem', '%s%s/include/interface/vcos/pthreads' % rpi_inc,
+					'-isystem', '%s%s/include/interface/vmcs_host/linux' % rpi_inc,
 			],
 				LIBPATH = '%s/lib' % rpi_vc_path,
 				LIBS = ['bcm_host'],
```

vLKp:
I was unable to find a gcc where the original form (with unconditional `=dir`) worked at all, so I merged your proposal as is, without adding any new configuration knobs.

psyke83:
@vLKp 

Thanks - and apologies for not responding sooner. It seems that later gcc releases are compiled without a default sysroot (`--with-sysroot=/`) configured:
gcc 6.3.0 has it:
```
Using built-in specs.
COLLECT_GCC=/usr/bin/gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/arm-linux-gnueabihf/6/lto-wrapper
Target: arm-linux-gnueabihf
Configured with: ../src/configure -v --with-pkgversion='Raspbian 6.3.0-18+rpi1+deb9u1' --with-bugurl=file:///usr/share/doc/gcc-6/README.Bugs --enable-languages=c,ada,c++,java,go,d,fortran,objc,obj-c++ --prefix=/usr --program-suffix=-6 --program-prefix=arm-linux-gnueabihf- --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --with-sysroot=/ --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-libitm --disable-libquadmath --enable-plugin --with-system-zlib --disable-browser-plugin --enable-java-awt=gtk --enable-gtk-cairo --with-java-home=/usr/lib/jvm/java-1.5.0-gcj-6-armhf/jre --enable-java-home --with-jvm-root-dir=/usr/lib/jvm/java-1.5.0-gcj-6-armhf --with-jvm-jar-dir=/usr/lib/jvm-exports/java-1.5.0-gcj-6-armhf --with-arch-directory=arm --with-ecj-jar=/usr/share/java/eclipse-ecj.jar --with-target-system-zlib --enable-objc-gc=auto --enable-multiarch --disable-sjlj-exceptions --with-arch=armv6 --with-fpu=vfp --with-float=hard --enable-checking=release --build=arm-linux-gnueabihf --host=arm-linux-gnueabihf --target=arm-linux-gnueabihf
Thread model: posix
gcc version 6.3.0 20170516 (Raspbian 6.3.0-18+rpi1+deb9u1) 
```
vs:
```
pi@rpi4:~ $ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/arm-linux-gnueabihf/8/lto-wrapper
Target: arm-linux-gnueabihf
Configured with: ../src/configure -v --with-pkgversion='Raspbian 8.3.0-6+rpi1' --with-bugurl=file:///usr/share/doc/gcc-8/README.Bugs --enable-languages=c,ada,c++,go,d,fortran,objc,obj-c++ --prefix=/usr --with-gcc-major-version-only --program-suffix=-8 --program-prefix=arm-linux-gnueabihf- --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --enable-bootstrap --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-libitm --disable-libquadmath --disable-libquadmath-support --enable-plugin --with-system-zlib --with-target-system-zlib --enable-objc-gc=auto --enable-multiarch --disable-sjlj-exceptions --with-arch=armv6 --with-fpu=vfp --with-float=hard --disable-werror --enable-checking=release --build=arm-linux-gnueabihf --host=arm-linux-gnueabihf --target=arm-linux-gnueabihf
Thread model: posix
gcc version 8.3.0 (Raspbian 8.3.0-6+rpi1) 
```
I was going to propose a runtime check via `gcc -print-sysroot`, but it would probably be an ugly hack, as you'd need to capture the output, (gcc 8 prints nothing, but the exit code is no different to gcc 6 which prints `/`). Your idea to allow a configurable sysroot would have been better than that. Thanks.

vLKp:
> It seems that later gcc releases are compiled without a default sysroot (`--with-sysroot=/`) configured:

That explains it.  For me, even old native gcc releases were built without a default sysroot, so I wasn't able to find any native gcc versions that worked without your change.  That is why I decided to merge it as-is, without adding the knob I originally proposed.

If there are people building for the Raspberry Pi who need the `=` for sysroot support, hopefully one of them will post here or file an issue.  If so, I can add in the proposed knob I documented above.  In the interim, since `raspberrypi=` is just a convenience setting, it should always be possible to skip use of that flag and manually set what it would have done, with any required system-specific modifications (adding/removing the leading `=`, choosing a different path for the libraries, etc.) added on top of it.

PR #454, opened 2019-09-17
Update Raspberry Pi Mesa build for Buster & Pi 4B

psyke83:
The new Raspberry Pi 4B revision has a minimum requirement of Raspbian
buster, and uses the Mesa V3D driver by default; in fact, the legacy
Broadcom drivers are no longer compatible with its VideoCore VC6 chip, so
building against Mesa is mandatory for this Pi revision.

Unfortunately, the Debian buster version of the Mesa packages
have intentionally dropped the GLESv1 headers, making building of GLESv1
software impractical (even though the VC4/V3D driver can still run GLESv1
applications).

The OpenGL driver in conjunction with SDL2* runs at full
speed on Raspberry Pi 3B & 4B, so it makes sense to modify the "mesa"
build to this configuration in light of these facts.

* SDL2 is highly desirable due to the availability of the KMSDRM driver, which is
the only way to run dxx-rebirth outside of X (we can no longer use
the dispmanx/rpi driver of SDL1).

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -3534,7 +3534,11 @@ class DXXCommon(LazyObjectConstructor):
 			return self.debug
 		# automatic setup for raspberrypi
 		def default_opengles(self):
-			if self.raspberrypi in ('yes', 'mesa'):
+			if self.raspberrypi in ('yes'):
+				return True
+			return False
+		def default_sdl2(self):
+			if self.raspberrypi in ('mesa'):
 				return True
 			return False
 		@classmethod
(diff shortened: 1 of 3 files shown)


vLKp:
Merged.  It looks like you didn't test this with `raspberrypi` unset.  In that case, a `TypeError` is raised.  I included commit [Fix TypeError in `default_opengles`, `default_sdl2`](/dxx-rebirth/dxx-rebirth/commit/b438f0f704e76c78c61067f4074b4aaac917a7e7) to resolve that.

PR #462, opened 2019-10-25
SDL2+Win32: readd Red Book support

ziplantil:
This readds Red Book (CD Audio) support for Win32 when using SDL2. With SDL2, there are no SDL functions for handling CD playback, unlike in SDL1. Therefore we need to implement our own code, and  this case, I implemented something that works on top of MCI. This allows using CD music on SDL2, although only on Windows for now.

common/arch/win32/rbaudio.cpp is a bit messy right now, since it only contains anything if we are on SDL2, but is still compiled even on SDL1 as long as the platform is Win32. It's probably a better idea to remove those #if's and modify the SConstruct script to include it only if both SDL2 and Win32, but I don't have the expertise to work with SConstruct scripts to a sufficient degree.

There are some hacks in the code - the most major one allows track end detection to be more lenient, as MCI on some setups stops playback just before your specified time, making it difficult to detect when the track actually ends. MCI would also support a notification system passing a Win32 message (MM_MCINOTIFY), but this doesn't work either for likely the same reason mentioned before (that it's unable to play properly until the actual end point).

Disc ID detection *should* produce the same results as on SDL1. I only tested it with some unrelated discs, as I don't possess actual Descent discs to test it with.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1358,7 +1358,7 @@ static void terminate_handler()
 			('DXX_MAX_AXES_PER_JOYSTICK', user_settings.max_axes_per_joystick),
 			('DXX_MAX_BUTTONS_PER_JOYSTICK', user_settings.max_buttons_per_joystick),
 			('DXX_MAX_HATS_PER_JOYSTICK', user_settings.max_hats_per_joystick),
-			('DXX_USE_SDL_REDBOOK_AUDIO', int(not sdl2)),
+			('DXX_USE_SDL_REDBOOK_AUDIO', int(not sdl2 or sys.platform in ['win32', 'msys'])), # there is a win32 rbaudio now
 		))
 		context.Display('%s: checking whether to enable joystick support...%s\n' % (self.msgprefix, 'yes' if user_settings.max_joysticks else 'no'))
 		# SDL2 removed CD-rom support.
@@ -4512,6 +4512,7 @@ class DXXArchive(DXXCommon):
 		get_platform_objects = LazyObjectConstructor.create_lazy_object_getter((
 'common/arch/win32/except.cpp',
 'common/arch/win32/messagebox.cpp',
+'common/arch/win32/rbaudio.cpp',
 ))
 	class DarwinPlatformSettings(DXXCommon.DarwinPlatformSettings):
 		get_platform_objects = LazyObjectConstructor.create_lazy_object_getter((
--- a/SConstruct
+++ b/SConstruct
@@ -1353,12 +1353,16 @@ static void terminate_handler()
 			# inputs.
 			user_settings.max_axes_per_joystick = user_settings.max_buttons_per_joystick = user_settings.max_hats_per_joystick = 0
 		successflags['CPPDEFINES'] = CPPDEFINES = successflags.get('CPPDEFINES', [])[:]
+		# use Redbook if at least one of the following applies
+		#    1. we are on SDL1
+		#    2. we are building for a platform for which we have a custom CD implementation (currently only win32)
+		use_redbook = int(not sdl2 or user_settings.host_platform == 'win32')
 		CPPDEFINES.extend((
 			('DXX_MAX_JOYSTICKS', user_settings.max_joysticks),
 			('DXX_MAX_AXES_PER_JOYSTICK', user_settings.max_axes_per_joystick),
 			('DXX_MAX_BUTTONS_PER_JOYSTICK', user_settings.max_buttons_per_joystick),
 			('DXX_MAX_HATS_PER_JOYSTICK', user_settings.max_hats_per_joystick),
-			('DXX_USE_SDL_REDBOOK_AUDIO', int(not sdl2 or sys.platform in ['win32', 'msys'])), # there is a win32 rbaudio now
+			('DXX_USE_SDL_REDBOOK_AUDIO', use_redbook),
 		))
 		context.Display('%s: checking whether to enable joystick support...%s\n' % (self.msgprefix, 'yes' if user_settings.max_joysticks else 'no'))
 		# SDL2 removed CD-rom support.
--- a/common/arch/win32/rbaudio.cpp
+++ b/common/arch/win32/rbaudio.cpp
@@ -16,6 +16,7 @@
 #include <stdlib.h>
 #include <windows.h>
 #include <mmsystem.h>
+#include <climits>
 
 #include "pstypes.h"
 #include "dxxerror.h"
@@ -36,7 +37,9 @@ namespace dcx {
 
 static UINT wCDDeviceID = 0U;
 static int initialised = 0;
-static DWORD playEnd;
+static int playEnd = 0;
+static int lastFrames = 0;
+static int isPaused = 0;
 
 void RBAExit()
 {
@@ -48,9 +51,9 @@ void RBAExit()
     }
 }
 
-static int mci_HasMedia()
+static bool mci_HasMedia()
 {
-    if (!wCDDeviceID) return 0;
+    if (!wCDDeviceID) return false;
 
     MCIERROR mciError;
     MCI_STATUS_PARMS mciStatusParms;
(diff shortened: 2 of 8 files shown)


common/arch/win32/rbaudio.cpp
```diff
@@ -307,7 +310,8 @@ int RBAPauseResume()
         con_puts(CON_VERBOSE, "RBAudio win32/MCI: Toggle Playback pause");
         RBAPause();
     }
-    else if (mciStatusParms.dwReturn == MCI_MODE_PAUSE)
+        // MCI_MODE_STOP is actually paused, because logic
```
vLKp: `because logic` is a bit vague.  Could you elaborate on this, either inline or in a block comment that explains the overall design?

common/arch/win32/rbaudio.cpp
```diff
@@ -36,7 +37,8 @@ namespace dcx {
 
 static UINT wCDDeviceID = 0U;
 static int initialised = 0;
-static DWORD playEnd;
+static int playEnd = 0;
```
vLKp: `static` variables automatically initialize to `0`.

vLKp: `int` is signed, whereas `DWORD` is unsigned.  Why do you need to change the signedness of this variable?

common/arch/win32/rbaudio.cpp
```diff
@@ -48,91 +50,114 @@ void RBAExit()
     }
 }
 
-static int mci_HasMedia()
+static bool mci_HasMedia()
```
vLKp: Why is this return type changed?

common/arch/win32/rbaudio.cpp
```diff
@@ -48,91 +50,114 @@ void RBAExit()
...
         RBAExit();
-		return 0;
+        return false;
     }
-    return static_cast<int>(mciStatusParms.dwReturn);
+    return mciStatusParms.dwReturn != 0;
 }
 
-static int mci_enableMsf()
+static unsigned int mci_TotalFramesMsf(int msf)
```
vLKp: Simple `unsigned` is preferred over `unsigned int`.

vLKp: If `msf` can be `const`, then it should be.

common/arch/win32/rbaudio.cpp
```diff
@@ -221,9 +236,9 @@ void RBAStop()
 
     if ((mciError = mciSendCommand(wCDDeviceID, MCI_STOP, MCI_WAIT, 0)))
     {
-		Warning("RBAudio win32: could not stop music (%lx)", mciError);
+        Warning("RBAudio win32/MCI: could not stop music (%lx)", mciError);
     }
-	redbook_finished_hook = NULL;
+    redbook_finished_hook = NULL;
```
vLKp: Prefer `nullptr` over `NULL`.

common/arch/win32/rbaudio.cpp
```diff
@@ -392,92 +417,69 @@ int RBAPeekPlayStatus()
...
 
-	if (mciStatusParms.dwReturn == MCI_MODE_PLAY)
+    if (mciStatusParms.dwReturn == MCI_MODE_PLAY)
         return 1;
-	else if (mciStatusParms.dwReturn == MCI_MODE_PAUSE)
-		return -1; // hack so it doesn't keep restarting paused music
-	else
-		return 0;
+    else if (mciStatusParms.dwReturn == MCI_MODE_PAUSE)
+        return -1; // hack so it doesn't keep restarting paused music
```
vLKp: Please document why `-1` is a hack that has this effect.  If possible, avoid using a hack at all, and instead structure the logic so that paused music stays paused without this special case.

common/arch/win32/rbaudio.cpp
```diff
@@ -392,92 +417,69 @@ int RBAPeekPlayStatus()
...
+    else
+        return 0;
 }
 
 static int cddb_sum(int n)
 {
-	int ret;
+    int ret;
```
vLKp: Please don't change indentation unless (1) the old layout is visibly confusing or (2) you are changing the surrounding logic such that a new indentation makes more sense (such as adding or removing a level of control flow statements, and reindenting the guarded block(s) accordingly).

Indentation should be done using tabs, with tabstops at 4 spaces.

SConstruct
```diff
@@ -1358,7 +1358,7 @@ static void terminate_handler()
 			('DXX_MAX_AXES_PER_JOYSTICK', user_settings.max_axes_per_joystick),
 			('DXX_MAX_BUTTONS_PER_JOYSTICK', user_settings.max_buttons_per_joystick),
 			('DXX_MAX_HATS_PER_JOYSTICK', user_settings.max_hats_per_joystick),
-			('DXX_USE_SDL_REDBOOK_AUDIO', int(not sdl2)),
+			('DXX_USE_SDL_REDBOOK_AUDIO', int(not sdl2 or sys.platform in ['win32', 'msys'])), # there is a win32 rbaudio now
```
vLKp: This is not correct.  `sys.platform` describes the system on which the program is built, not the system on which it will be run.  When I cross-compile from Linux to produce a program for Windows users to run, this should default to true, but `sys.platform` in that case would be `linux2`.

common/arch/sdl/event.cpp
```diff
@@ -44,6 +49,29 @@ static void windowevent_handler(const SDL_WindowEvent &windowevent)
...
+{
+#if defined(_WIN32) && SDL_MAJOR_VERSION == 2
+#define MM_MCINOTIFY 0x3B9
+	if (message->subsystem == SDL_SYSWM_WINDOWS) {
+		con_printf(CON_VERBOSE, "SYSWM message M:%08x (%6d) - W:%08x L:%08lx", message->msg.win.msg, message->msg.win.msg, message->msg.win.wParam, message->msg.win.lParam);
+		if (message->msg.win.msg == MM_MCINOTIFY)
+			RBAGotMCIEvent(message->msg.win.wParam, message->msg.win.lParam);
+	}
+#else
+	message
```
vLKp: This should be `(void)message;` if you want to suppress the unused variable diagnostic.

common/arch/win32/rbaudio.cpp
```diff
@@ -0,0 +1,530 @@
+/*
```
vLKp: This file looks very familiar.  Why are you copying it instead of patching the existing file?  What are your plans for avoiding duplicate work maintaining separate copies of the file, one for SDL1 and one for SDL2, in different paths?

common/include/rbaudio.h
```diff
@@ -74,7 +75,41 @@ unsigned long RBAGetDiscID();
 
 // List the tracks on the CD
 void RBAList(void);
+
+#elif SDL_MAJOR_VERSION == 2
+
```
vLKp: Much of this appears to be duplicated from above.  Why not move the parts that are common out of the is-SDL-1 test, instead of duplicating common declarations in both blocks?

common/arch/win32/rbaudio.cpp
```diff
@@ -372,7 +372,7 @@ void RBACheckFinishedHook()
         if (thisFrames == lastFrames)
             checkValue -= 64;
 
-        if (playEnd >= 0 && lastFrames >= checkValue)
+        if (redbook_finished_hook && playEnd >= 0 && lastFrames >= checkValue)
```
vLKp: Please squash this into the commit that added the call.

common/arch/win32/rbaudio.cpp
ziplantil: This is an MCi design element which this code can't do anything about. The comment could probably be more clear on that, though.

common/arch/win32/rbaudio.cpp
ziplantil: NULL is also used by the SDL code right now too, so it should be changed in both.

common/arch/win32/rbaudio.cpp
ziplantil: This is also used in the SDL CD code.

SConstruct
ziplantil: Yes, it should check the target environment, but I'm not sure how to achieve that.

common/arch/win32/rbaudio.cpp
ziplantil: common/arch/sdl/rbaudio.cpp is for SDL1 which has its own CD code, common/arch/win32/rbaudio.cpp will be for SDL2 which does _not_ have its own CD code (it was removed in SDL2 and there is no replacement, as their documentation says), so instead it uses a Windows-exclusive API to play from the CD directly without using SDL as the middle man.

common/arch/win32/rbaudio.cpp
ziplantil: `bool` makes no sense, as the function can ever only logically return true/false.

common/arch/win32/rbaudio.cpp
ziplantil: The value can actually never be negative, and `playEnd` is always assigned from `dwReturn`, itself `DWORD_PTR` but actually containing a `DWORD` value.

SConstruct
vLKp: Make the default value a function, not an inline value.  When a default is needed, the function will be called.  That function can inspect settings that were defined above the setting for which this default is being computed.  Host platform is one of the first settings defined, so that functions for computing defaults can inspect it.  See `__default_DATA_DIR` for an example (though unfortunately not a very obvious example).

https://github.com/dxx-rebirth/dxx-rebirth/blob/d0566119891c7f381cbe7a375c09ccda6c0c689a/SConstruct#L3573-L3578

common/arch/win32/rbaudio.cpp
vLKp: I am aware SDL removed their support.  In my initial review, it looked like there would be a substantial amount of code common between the two implementations.  In reviewing a `diff` between the two files, I see that even after suppressing whitespace noise, there is very little in common.  I withdraw my objection here.

common/arch/win32/rbaudio.cpp
ziplantil: Wouldn't it make more sense to just squash all commits thus far?

SConstruct
ziplantil: The value is set in a completely different context for `DXX_USE_SDL_REDBOOK_AUDIO` than `sharepath`, and it seems the former would not have access to `self._program` or a similar structure.

common/arch/win32/rbaudio.cpp
vLKp: On Win64, `DWORD_PTR` will be 64 bits, but `DWORD` will be 32 bits.  Will that truncation cause an issue?

SConstruct
vLKp: This makes it even easier, then.  The defines immediately above the changed line are based on `user_settings`, and `user_settings` is where you can find `host_platform`.  I think you could do this:

```diff
--- a/SConstruct
+++ b/SConstruct
@@ -1358,7 +1358,7 @@
 			('DXX_MAX_AXES_PER_JOYSTICK', user_settings.max_axes_per_joystick),
 			('DXX_MAX_BUTTONS_PER_JOYSTICK', user_settings.max_buttons_per_joystick),
 			('DXX_MAX_HATS_PER_JOYSTICK', user_settings.max_hats_per_joystick),
-			('DXX_USE_SDL_REDBOOK_AUDIO', int(not sdl2 or sys.platform in ['win32', 'msys'])), # there is a win32 rbaudio now
+			('DXX_USE_SDL_REDBOOK_AUDIO', int(not sdl2 or user_settings.host_platform == 'win32')), # there is a win32 rbaudio now
 		))
 		context.Display('%s: checking whether to enable joystick support...%s\n' % (self.msgprefix, 'yes' if user_settings.max_joysticks else 'no'))
 		# SDL2 removed CD-rom support.
```

I dropped the `msys` test since `host_platform` is `win32` even for msys.

https://github.com/dxx-rebirth/dxx-rebirth/blob/d0566119891c7f381cbe7a375c09ccda6c0c689a/SConstruct#L3652-L3659

common/arch/win32/rbaudio.cpp
ziplantil: MSF is a 32-bit format as specified by the MCI documentation, so no, truncation won't be an issue, but I'll add a comment about it.

vLKp:
This looks fine to me.  I'd like to see the Win32/SDL2 `rbaudio.cpp` handled a bit differently, but I can take care of that later.  I can't test any of this, though.  @zicodxx: any comments or objections before I merge this?

vLKp:
No objections from anyone.  Merged.

Issue #467, opened 2019-11-07
Runtime crash: Crash if Vertigo bots present after exit sequence plays


ziplantil:
### Environment

93d3793b306cebb50be1dc86f974204d5d877ac0

#### Operating System Environment

* Microsoft Windows
  * [ ] Microsoft Windows XP
  * [ ] Microsoft Windows Vista
  * [ ] Microsoft Windows 7
  * [ ] Microsoft Windows 8
  * [ ] Microsoft Windows 8.1
  * [x] Microsoft Windows 10

* Mac OS X
  * [ ] Mac OS X 10.8
  * [ ] Mac OS X 10.9
  * [ ] Mac OS X 10.10
  * [ ] Mac OS X 10.11
  * [ ] Mac OS X 10.12
  * [ ] Mac OS X 10.13

* Debian
  * [ ] Debian Wheezy
  * [ ] Debian Jessie
  * [ ] Debian Stretch
  * [ ] Debian Buster
* Fedora
  * [ ] Fedora 26
  * [ ] Fedora 27
  * [ ] Fedora 28
  * [ ] Fedora 29
  * [ ] Fedora 30
  * [ ] Rawhide
* Ubuntu
  * [ ] Ubuntu 16.04 LTS (Xenial Xerus)
  * [ ] Ubuntu 16.10 (Yakkety Yak)
  * [ ] Ubuntu 17.04 (Zesty Zapus)
  * [ ] Ubuntu 17.10 (Artful Aardvark)
  * [ ] Ubuntu 18.04 LTS (Bionic Beaver)
  * [ ] Ubuntu 18.10 (Cosmic Cuttlefish)
  * [ ] Ubuntu 19.04 (Disco Dingo)

* [ ] Arch
* [ ] Gentoo

* [ ] OpenBSD
* [ ] FreeBSD
* [ ] NetBSD

#### CPU environment

* [ ] x86 (32-bit Intel/AMD)
* [x] x86\_64 (64-bit Intel/AMD)
* [ ] ARM (32-bit)
* [ ] ARM64 (64-bit; sometimes called AArch64)

#### Game environment
Any custom level set with more than one level, using D2 exit sequences (not the movies, but with exit.ham etc.; such custom missions are currently few and far between, if they exist at all) and Vertigo or other V-HAM robots. I have included a minimal example here: [exitd2v2.zip](https://github.com/dxx-rebirth/dxx-rebirth/files/3821925/exitd2v2.zip)

### Description
[...]

PR #479, opened 2019-12-12
Do not free extra models when loading exit data on D2

### LINKED ISSUE #460: Runtime crash: D2 exit sequences crash if Vertigo bots present
### Environment
d0566119891c7f381cbe7a375c09ccda6c0c689a

* Microsoft Windows
  * [ ] Microsoft Windows XP
  * [ ] Microsoft Windows Vista
  * [ ] Microsoft Windows 7
  * [ ] Microsoft Windows 8
  * [ ] Microsoft Windows 8.1
  * [x] Microsoft Windows 10

* Mac OS X
  * [ ] Mac OS X 10.8
  * [ ] Mac OS X 10.9
  * [ ] Mac OS X 10.10
  * [ ] Mac OS X 10.11
  * [ ] Mac OS X 10.12
  * [ ] Mac OS X 10.13

* Debian
  * [ ] Debian Wheezy
  * [ ] Debian Jessie
  * [ ] Debian Stretch
  * [ ] Debian Buster
* Fedora
  * [ ] Fedora 26
  * [ ] Fedora 27
  * [ ] Fedora 28
  * [ ] Fedora 29
  * [ ] Fedora 30
  * [ ] Rawhide
* Ubuntu
  * [ ] Ubuntu 16.04 LTS (Xenial Xerus)
  * [ ] Ubuntu 16.10 (Yakkety Yak)
  * [ ] Ubuntu 17.04 (Zesty Zapus)
  * [ ] Ubuntu 17.10 (Artful Aardvark)
  * [ ] Ubuntu 18.04 LTS (Bionic Beaver)
  * [ ] Ubuntu 18.10 (Cosmic Cuttlefish)
  * [ ] Ubuntu 19.04 (Disco Dingo)

* [ ] Arch
* [ ] Gentoo

* [ ] OpenBSD
* [ ] FreeBSD
* [ ] NetBSD

#### CPU environment

* [ ] x86 (32-bit Intel/AMD)
* [x] x86\_64 (64-bit Intel/AMD)
* [ ] ARM (32-bit)
* [ ] ARM64 (64-bit; sometimes called AArch64)

#### Game environment
Any custom level using D2 exit sequences (**not** the movies, but with exit.ham etc.; such custom missions are currently few and far between, if they exist at all) and Vertigo or other V-HAM robots. A minimal example is included [here](https://github.com/dxx-rebirth/dxx-rebirth/files/3762403/exitd2v.zip) (Vertigo required to play).

### Description
The game crashes as the exit sequence begins. The most likely explanation for this is because when the exit sequence model data is loaded under similar/main/bm.cpp, lines [647-648](https://github.com/dxx-rebirth/dxx-rebirth/blob/340dc79bec2bac5a5d696251fc639214b0002fc1/similar/main/bm.cpp#L647-L648) unload all custom robots and textures, which causes a bad index whenever trying to render the robot model as it is no longer in memory. and there is indeed an assertion failure in [similar/main/polyobj.cpp:504](https://github.com/dxx-rebirth/dxx-rebirth/blob/87125c505354a383b8fe2b057a29733685f8639b/similar/main/polyobj.cpp#L504).

Note that with two Vertigo robots (unidentified and Fervid-99), the game does _not_ crash, but instead renders the robots with exit hatch models.

The issue may be difficult to fix. The "better" solution is to get rid of those free calls entirely and replace them with something else if necessary, while an "okay" solution would be to check whether the index is correct when rendering polymodels once the exit sequence has started (but in this case, the exit hatch model indexes still need to be disjoint with any Vertigo model indexes), or simply deleting all Vertigo robots once the exit sequence begins.

#### Regression status
The problem is likely present in all Rebirth and D2X versions ever since the D2X version that reintroduces D2 exit sequence support. The DOS version also unloads extra robots when the exit data is loaded, which however never happens as the exit sequence code is never called / not present in its entirety in the release version (instead, D2 always uses movies).

### Steps to Reproduce
1. Start level
2. Open and fly through the exit door

Expected result: The exit sequence plays normally, with the robot acting like D2 robots, standing in place, or alternatively vanishing as the sequence begins.

Actual result: The game crashes. No crash dump seems to be generated, but it's more than likely that the issue is with rendering the Vertigo robot present in the exit tube.

* Engines affected:
  * [ ] D1X-Rebirth
  * [x] D2X-Rebirth

* Game modes affected:
  * [x] Single player
  * [ ] Multiplayer competitive
    * [ ] Anarchy
    * [ ] Team Anarchy
    * [ ] Robo-Anarchy
    * [ ] Capture the Flag
    * [ ] Bounty
  * [ ] Multiplayer cooperative

ziplantil:
This should fix #460 and #467, as now the Vertigo models are **not** unloaded in load_exit_models. The original game unloads all models there, presumably to make space for the exit models (and to unload the models if they are already loaded). However, this also unloads Vertigo robot models, which causes crashes if they are present after the exit data is loaded (even after a level change).

The fix instead adds a separate "exit models loaded" flag. It is set when exit data is loaded and *should* be cleared whenever it is unloaded. If the exit data is then loaded again while the flag is already set, no new data is loaded and instead the function returns immediately.


similar/main/bm.cpp
```diff
@@ -295,6 +296,7 @@ int gamedata_init()
 
 namespace dsx {
 
+// (currently only ever called if EDITOR is set)
```
vLKp: This is not correct.  This function is included in `!EDITOR` builds and is called during early startup.

```(gdb) bt
#0  d2x::bm_read_all (Vclip=..., fp=0x5555564d5300) at similar/main/bm.cpp:305
#1  0x000055555560c40a in read_hamfile () at common/include/physfsx.h:421
#2  0x000055555560c96e in properties_init () at similar/main/piggy.cpp:1131
#3  0x00005555555922ae in gamedata_init () at similar/main/bm.cpp:288
#4  0x000055555556cbff in d2x::main () at similar/main/inferno.cpp:623
#5  main () at similar/main/inferno.cpp:746
```

common/main/bm.h
```diff
@@ -150,6 +150,7 @@ extern int Robot_replacements_loaded;
 constexpr std::integral_constant<unsigned, 610> MAX_OBJ_BITMAPS{};
 extern unsigned N_ObjBitmaps;
 extern int extra_bitmap_num;
+extern int Exit_models_loaded;
```
vLKp: This could be a `int8_t`, correct?

similar/main/bm.cpp
```diff
@@ -396,6 +398,7 @@ void bm_read_all(d_vclip_array &Vclip, PHYSFS_File * fp)
 }
 
 int extra_bitmap_num = 0;
+int Exit_models_loaded = 0; // only really used for D2
```
vLKp: Globals default to `0`.  Some toolchains are not clever enough to place explicitly zero-initialized variables in `.bss`.

similar/main/polyobj.cpp
```diff
@@ -569,6 +570,9 @@ void draw_polygon_model(grs_canvas &canvas, const vms_vector &pos, const vms_mat
 
 void free_polygon_models()
 {
+#if defined(DXX_BUILD_DESCENT_II)
+	Exit_models_loaded = 0; // reload exit models again if needed
```
vLKp: Could you explain why this reset is here?  This function only unloads polygon model data, but does not call `bm_unload_last_objbitmaps` or do equivalent work.

similar/main/bm.cpp
```diff
@@ -640,12 +646,33 @@ static grs_bitmap *bm_load_extra_objbitmap(const char *name)
 	}
 }
 
+static void bm_unload_last_objbitmaps(int count)
```
vLKp: Generally, counters should be `unsigned`, because it does not make sense to free a negative quantity of something.

similar/main/bm.cpp
```diff
@@ -640,12 +646,33 @@ static grs_bitmap *bm_load_extra_objbitmap(const char *name)
...
+{
+	extra_bitmap_num = Num_bitmap_files;
+	// unload last texture count times
+	while (count--)
```
vLKp: This looks like a good place to use `partial_range`.

common/main/bm.h
ziplantil: It could be any type that can contain both a 0 and a 1. In fact, it could even be convertible to a `bool`.

similar/main/bm.cpp
ziplantil: Is explicit initialization really ever *wrong*, though?

similar/main/polyobj.cpp
ziplantil: Because the exit models are still unloaded, and would need to be loaded again. It is fairly safe to assume that unloading models on their own never happens separate from also unloading other game data, which would include the bitmaps. If not, the only side effect is that the exit textures get temporarily doubled in memory.

similar/main/bm.cpp
ziplantil: Ah, indeed. I was probably confused by the fact that it's inside an #if block that according to a comment _used_ to be EDITOR-dependent, but on a second look it was originally `#ifndef EDITOR` and would've been called anyway.

That makes it somewhat more challenging, since I'm currently not sure if the values it sets for `exit_modelnum` and `destroyed_exit_modelnum` matter. It seems that they're initialized to the number of models (i.e. invalid values) on D2 (PIG_VERSION >= 3), which implies they would be loaded later - so only the comment needs to be changed.

similar/main/polyobj.cpp
ziplantil: Of course, it would be possible to have two separate flags instead - one for the bitmaps and one for the models.

ziplantil: I did separate the two flags - I couldn't find any more spots where Exit_bitmaps_loaded should be reset.

vLKp:
I don't see anything particularly wrong with the code, so I merged it locally.  However, in testing, I am seeing a strange graphical anomaly.  I ran `d2x-rebirth` and started the `Descent: Counterstrike` (D1) campaign.  I used Del-Shift-B to force detonate the mine, then flew across the exit trigger.  I did this on the first three levels in quick order.  On the second and later levels, after the ship flies out into open space above the surface of the mine, the building that sits on the exit should be a fairly consistent dark grey, but is instead using what appear to be textures inherited from the level.  Valgrind doesn't report anything wrong, so the memory is not directly corrupt.  Do you see this failure mode as well?  Would you mind debugging this, or do I need to fix this before I merge it?

ziplantil:
I can take a look at it - it's very possible that the exit textures are loaded again, but it won't actually change the texture indexes the already loaded models are using, even though it should.

ziplantil:
I can't reproduce the textures being wrong after my latest commit

vLKp:
I still see this problem.
```
# Checkout current master
# Merge changes proposed here
git checkout 065bcb3da8c62b36e60cb8beadd5104040b3eed2
git merge e5c7d471991377bb5bdc9a591f6eabf6ec0b25b1 -m 'Merge'
scons d2x=gcc9 editor=1
./d2x-rebirth-editor
```
- Start `Descent: First Strike`, level 1.
- Use Del-shift-B.  First level looks correct.
- Use Del-shift-B again to terminate level 2.  Exit model is now incorrect.
- Use Del-shift-B again to terminate level 3.  Exit model is still incorrect, and I caught a screenshot of it.  Unfortunately, Github is refusing to attach the screenshot with `Something went really wrong` and no explanation.

ziplantil:
Confirmed there. For some reason I missed the fact that it was the D1 mission being tested; I was testing a D2 mission that provided its own data to re-enable the exit sequences. The natural solution here is to probably just reload everything when playing a D1 mission, since the values are set elsewhere anyway, and there's no easy way to handle those cases.

vLKp:
Looks good.  Merged.  Thank you for the fixes.

Issue #475, opened 2019-11-20
Build failure: SConstruct is python2 but scons is python3


SanskritFritz:
Build fails on Archlinux:
```
scons: Reading SConscript files ...
  File "dxx-rebirth/src/dxx-rebirth-0.59.100/SConstruct", line 17
    print "%s: %s" % (program.program_message_prefix, msg)
                 ^
SyntaxError: invalid syntax
```
Scons depends on python3 in archlinux.
How do I build dxx-rebirth with this scenario?

parkerlreed:
Is this recent? I swear this has been building for me just fine.

```
[parker@stealth build]$ pacman -Q d1x-rebirth-git 
d1x-rebirth-git 0.60.0.beta2.r575.gca383c1fe-1
```

So yeah newest commit is installed (Using the dxx-rebirth-git from AUR)

parkerlreed:
Wait why are you trying to build a 3 year old release?

![image](https://user-images.githubusercontent.com/841440/69283630-c1b66480-0bba-11ea-86ad-4add21ec197f.png)

Either use the newer or pull form master

vLKp:
As parkerlreed suggests, this is _definitely_ not recent.  Line 17 doesn't even have a `print` statement on it and has not since at least a3d6bdd3f1a642df29df3b61dc2b53120a6308e8 in April 2018, which added quite a bit of Python3 compatibility and also pushed the `print` statement deeper into the file.  Digging deeper into history, line 17 has not had the particular `print` statement shown in the opening post since af863fdef7eb3d351c7ac6d6677a245fc3aae3af added Python3 print compatibility in October 2017.

To answer your immediate question: the preferred solution is to do as parkerlreed said, and use a supported version of current code.  You could also choose to run `scons` with a python2 interpreter.

I don't usually close issues outright, but since the limited information provided shows that this was filed for the old and unsupported snapshot 0.59.100 and the reported issue has been fixed for more than a year, I think a closure is appropriate here.  @SanskritFritz: please try a current snapshot.

@SanskritFritz: did you download the code by hand or is there an Arch Linux package that tried to build this snapshot?  If the latter, please advise its maintainer that an update to a newer version is needed.

SanskritFritz:
I'm trying to create a package for the "stable" release and I thought 0.59 is the last stable release. Which release should I consider stable now?

vLKp:
[...]

PR #482, opened 2019-12-29
Archlinux PKGBUILD fixes.

SanskritFritz:
I simply copied my version over the old files, making it up-to-date now.


SanskritFritz:
#475 for reference.

contrib/arch/d1x-rebirth/PKGBUILD
```diff
@@ -1,36 +1,44 @@
...
 
 pkgname=d1x-rebirth
-pkgver=0.61
-pkgrel=1
-pkgdesc="An enhanced engine to play with Descent1 datas"
-arch=('i686' 'x86_64')
+pkgver=20191121
+pkgrel=2
+pkgdesc="An enhanced engine to play with Descent1 data."
+arch=('x86_64')
```
vLKp: Rebirth still supports i686 targets.  What would be required to leave this enabled?

contrib/arch/d1x-rebirth/PKGBUILD
```diff
@@ -1,36 +1,44 @@
...
-source=(git+https://github.com/dxx-rebirth/dxx-rebirth.git $pkgname.desktop $pkgname.install https://www.dxx-rebirth.com/download/dxx/res/d1xr-hires.dxa)
-noextract=(d1xr-hires.dxa d1xr-sc55-music.dxa)
-md5sums=('SKIP'
-         'ded1f8c2deb3825134e1a984985684cb'
-         '78c79b2b82175ebd4073edf257a9bd68'
-         '7d978ce3875995e54cbf426167e88b94') 
+makedepends=('scons' 'unzip')
+install="$pkgname.install"
+source=("https://www.dxx-rebirth.com/download/dxx/rebirth/dxx-rebirth_$pkgver-src.tar.xz"
+        "http://www.dxx-rebirth.com/download/dxx/res/d1xr-hires.dxa"
```
vLKp: These should be `https`.

contrib/arch/d1x-rebirth/d1x-rebirth.install
```diff
@@ -9,10 +9,10 @@ descent.pig
 
 You don't own the full game, yet?
 Just visit GOOD OLD GAMES via:
-https://www.gog.com/game/descent/pp/fc074d501302eb2b93e2554793fcaf50b3bf7291
+http://www.gog.com/en/gamecard/descent_1_descent_2/pp/fc074d501302eb2b93e2554793fcaf50b3bf7291
```
vLKp: Your link is out of date.  The URL in the Rebirth repository is correct.

contrib/arch/d1x-rebirth/d1x-rebirth.install
```diff
@@ -9,10 +9,10 @@ descent.pig
...
 and download it for a very low price! DRM-FREE!
 
-You can get AddOns for the game from https://www.dxx-rebirth.com/addons/
+You can get AddOns for the game from http://www.dxx-rebirth.com/addons/
```
vLKp: This should remain `https`.

contrib/arch/d1x-rebirth/PKGBUILD
SanskritFritz: Archlinux doesn't support i686 anymore so we drop that target most of the time. But I can leave it enabled, it causes no problems but I thought I'd stick to archlinux recommendations.

contrib/arch/d1x-rebirth/d1x-rebirth.install
SanskritFritz: Oops, thanks for the heads-up.

contrib/arch/d1x-rebirth/PKGBUILD
SanskritFritz: Yes, you're right.

contrib/arch/d1x-rebirth/d1x-rebirth.install
SanskritFritz: Well, none of the above links actually work here, this is what I have found:
https://www.gog.com/game/descent
Should we use that from now on?

contrib/arch/d1x-rebirth/PKGBUILD
vLKp: If Arch has dropped i686 support, I'm fine following their lead for the PKGBUILD file.  Leave this line as you have it.

contrib/arch/d1x-rebirth/d1x-rebirth.install
vLKp: It looks like they changed the format of their partner link.  However, since the `INSTALL.markdown` instructions are not using a partner link, I think using a non-partner link here is right for consistency.  Yes, use the bare form that you proposed.

vLKp:
Looks good.  Merged.

PR #490, opened 2020-01-09
Small fix: Thief bot immediately drops all map-specified powerups if suppressed

AlumiuN:
Currently, if the option to suppress the thief bot is enabled, any objects it was to drop will be lost along with it. This is mostly only relevant when the thief bot is carrying keys, but this fix will drop any item it was carrying before deleting the bot (as the level could feasibly rely on other objects, for example requiring a Phoenix Cannon or Guided Missiles for hitting an out-of-sight switch). I've tested this in single player, and I believe (albeit with my minimal knowledge of the network code) that it should work in multiplayer as well, but I haven't tested that.

Also, I apologise for the extra commits; my fork also includes the load_mission_by_name fix, and I couldn't figure out how to get rid of them beyond just resetting it with an extra commit. Nothing that I've encountered in software development thus far has made me feel more like I need an adult than version control. :P

### DIFF
--- a/similar/main/gameseq.cpp
+++ b/similar/main/gameseq.cpp
@@ -419,7 +419,10 @@ static void gameseq_init_network_players(object_array &Objects)
 			auto &ri = Robot_info[get_robot_id(o)];
 			if ((!retain_guidebot && robot_is_companion(ri)) ||
 				(remove_thief && robot_is_thief(ri)))
+			{
+				object_create_robot_egg(o);
 				obj_delete(LevelUniqueObjectState, Segments, o);		//kill the buddy in netgames
+			}
 		}
 #endif
 	}
@@ -475,7 +478,10 @@ void gameseq_remove_unused_players()
 				{
 					auto &ri = Robot_info[get_robot_id(o)];
 					if (robot_is_thief(ri))
+					{
+						object_create_robot_egg(o);
 						obj_delete(LevelUniqueObjectState, Segments, o);
+					}
 				}
 			}
 		}
--- a/similar/main/gameseq.cpp
+++ b/similar/main/gameseq.cpp
@@ -420,7 +420,10 @@ static void gameseq_init_network_players(object_array &Objects)
 			if ((!retain_guidebot && robot_is_companion(ri)) ||
 				(remove_thief && robot_is_thief(ri)))
 			{
-				object_create_robot_egg(o);
+				if (o->contains_count > 0)
+				{
+					object_create_robot_egg(o);
+				}
 				obj_delete(LevelUniqueObjectState, Segments, o);		//kill the buddy in netgames
 			}
 		}
@@ -479,7 +482,10 @@ void gameseq_remove_unused_players()
 					auto &ri = Robot_info[get_robot_id(o)];
 					if (robot_is_thief(ri))
 					{
-						object_create_robot_egg(o);
+						if (o->contains_count > 0)
+						{
+							object_create_robot_egg(o);
+						}
 						obj_delete(LevelUniqueObjectState, Segments, o);
 					}
 				}


vLKp:
> Also, I apologise for the extra commits

```
git checkout origin/master -b thief-suppress-fix
git cherry-pick b68bc945043ef153747b22719366cf53231fb91d
```

vLKp:
I recall a mention of this problem elsewhere, but it looks like it was never filed or fixed.  I'd like to see it work in multiplayer before I merge it.  Do you have any example levels where the lack of this fix is a serious problem?  Such levels may make it easier to validate that the fix works properly in multiplayer (including synchronizing the items properly).

AlumiuN:
I had no idea that cherry-pick existed, that's very useful, thanks you.

Unfortunately I don't have any examples of maps that use it; the problem was brought to my attention by Zico, and it sounds like there is more than one, but I haven't encountered any myself and Zico couldn't remember a specific one.

On the subject of whether it'll work in multiplayer, I'm actually not sure it will; looking at multi_explode_robot_sub suggests to me that just adding the object to Net_create_objnums (as drop_robot_egg and drop_powerup do) may not be sufficient (a multi_send_create_robot_powerups might also be required, although co-op, which is mostly where this is relevant, doesn't appear to need it), but I haven't looked into it deeply enough to be able to tell properly. I'll see about testing it sometime in the near future, hopefully tonight.

AlumiuN:
OK, it appears to work in multiplayer for keys; however, other powerups (I tested with a Phoenix Cannon) cause a crash, or it creates whatever key was last created if a level where it occurs with a key was loaded was played beforehand. I'll look into it.

(also, the cherry-pick doesn't appear to be working; should I be deleting the branch this pull request is based off and creating a new one to cherry pick the commit?)

AlumiuN:
I've added some more code which appears to fix the issue (and thanks to a knowledgable friend, rebased the branch so it contains only the two relevant commits), however I'm having another issue with the test map which I believe to be unrelated. I'm going to attempt to ensure it's unrelated by verifying that it happens without the changes so I can be confident that this actually fixes the problem without breaking anything else.

vLKp:
The original patch (using only `object_create_robot_egg`) might have been properly synchronized since this code runs very early during game setup, so if everyone ran the same code, they would all permute the level in the same way and the game would act as if the level author had built the level with no thief and with the items placed floating free.  Other load-time modifications (remapping disallowed powerups to shields, cloning powerups) rely on this principle.  However, I was unsure from memory whether that would work here, hence my request to see it work.  I would have done that test if I had a sample map where I could readily validate it.  I appreciate you testing it for me.

> (also, the cherry-pick doesn't appear to be working; should I be deleting the branch this pull request is based off and creating a new one to cherry pick the commit?)

You need to start from a branch which has none of those commits, yes.  You can do that by using a different branch name (remember: branch names do not need to match between local and remote, though it may be easier to reason about them if they do match), by deleting your local branch first, or by resetting it using `git reset --hard $new_head_commit`.  (Caution: `--hard` discards any uncommitted local changes.  You __will lose work__ if you use this incorrectly.)

AlumiuN:
Actually, you're right, after all the work figuring out that the crash was unrelated I forgot to test whether the modification I made was actually necessary, and it turns out that `object_create_robot_egg` works fine. I'll leave the check against o->contains_count (and add it to the single-player check) and remove the other extraneous bits.

vLKp:
That last commit is not actually needed.  `object_create_robot_egg` handles being called for `->contains_count == 0`.

I will merge this pull the next time I work on the code.

AlumiuN:
Ah, OK; I think as part of learning to code by modifying Wolf3D, I usually put the check outside the function to avoid an unnecessary function call, but that probably hasn't had any measurable impact since the 90s (i.e. before I started coding) :P

vLKp:
It might still matter if:

1. The function call is virtual (these aren't) or if
2. a. The path is very hot (this isn't) and
  b. The function call is usually skipped (it will be).

Although somewhat limited due to the cross-translation unit setup here, modern compilers can perform partial inlining to move such checks up when the compiler decides the movement is useful.  For cross-TU usage, link time optimization would be required to make partial inlining possible.

AlumiuN:
Yeah, that's fair; I forget compilers will probably do most of that thinking for me these days.

vLKp:
I merged the first commit, which does the work, and skipped the second commit, which only added the out-of-line count checking.  I think this is sufficient to address the intent of the merge request.  Please reopen if you disagree.

PR #492, opened 2020-01-24
Improve handling of doors with linked walls other than their connected side

AlumiuN:
Currently, the handling of doors with linked walls other than their current sides is both inaccurate to DOS Descent and in a lot of cases (probably 5 out of 6 times on average) will cause a crash. The extent and reasoning of the changes I've made are as follows:

- The code segment in wall_open_door was previously looking in the wrong segment for the linked wall's connected side; as such, whenever the segment it was looking in had no wall on the same side number where the linked wall's connected side was, back_wallnum[1] was being set to wall_none, which would then almost immediately cause a crash when do_door_open used it to call kill_stuck_objects.
- do_close_door had a similar issue, whereby it was not using the correct segment when dealing with the linked wall's door. This didn't seem to result in a crash in my testing, but it did usually result in the linked wall's door not opening.
- Additionally, because do_close_door was calling wall_close_door_ref inside the loop, but this no longer actually deletes the door, if the animation length of the linked door was longer than that of the primary door, the linked door would be set to the closed frame when the main door finished, and then immediately set to the frame otherwise appropriate for its animation at that time, meaning that it wouldn't fully close.

These changes mostly bring Rebirth in line with DOS, including the double-speed animations (a result of d.time being increased inside the n_parts loop). However, there is still a missing behaviour (a glitch, strictly speaking) that may be more annoying to implement; in DOS, opening the linked door normally, followed by opening the door that links to it, such that the linked door will close while the main door is in the waiting state, results in the main door remaining visually shut but physically open permanently (or at least, until a trigger acts upon it). This is due to how DOS Descent deletes doors as they close, as it results in the main door's second part (the linked wall's door) being processed without resetting d.time and, as a result, immediately closing the door without setting the WALL_DOOR_CLOSED flag (the same deletion method also otherwise results in any door, linked or otherwise, sequentially after one that closes skipping a frame, which is a bug I don't think Rebirth has).

### DIFF
--- a/similar/main/wall.cpp
+++ b/similar/main/wall.cpp
@@ -479,9 +479,10 @@ void wall_open_door(const vmsegptridx_t seg, const unsigned side)
 		w2->state = WALL_DOOR_OPENING;
 
 		const auto &&seg2 = vcsegptridx(w2->segnum);
-		Connectside = find_connect_side(seg2, vcsegptr(seg2->children[w2->sidenum]));
+		const auto &&csegp2 = vcsegptr(seg2->children[w2->sidenum]);
+		Connectside = find_connect_side(seg2, csegp2);
 		Assert(Connectside != side_none);
-		const auto cwall_num = csegp->shared_segment::sides[Connectside].wall_num;
+		const auto cwall_num = csegp2->shared_segment::sides[Connectside].wall_num;
 		auto &imwallptr = Walls.imptr;
 		if (const auto &&w3 = imwallptr(cwall_num))
 			w3->state = WALL_DOOR_OPENING;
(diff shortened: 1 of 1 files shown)


vLKp:
[Use vwallptr](/dxx-rebirth/dxx-rebirth/commit/553b31309946a6c2d201e17ddac53c4652baf243) added a reload of `cwall_num` from the wrong segment.

Your change looks good, and a basic test reveals no problems.  Merged.  Thank you for including a fix with the diagnosis.

AlumiuN:
No problem :) Do you think it's worth (as weird as this sounds) implementing the glitch I mentioned? Not that I'm sure what would be the best method of going about doing so.

vLKp:
Based on your description of the glitch, it's almost certainly a bug in the original that it worked that way.  Generally, I prefer that Rebirth only replicate original game bugs when the bug becomes a relatively popular ascended glitch.  For example, I've seen some claims that dichording / trichording wasn't supposed to grant a faster total travel speed than flying in only one axis, but almost everyone quickly learns that it does and builds strategies around it, so I would consider it at least an ascended glitch, if not a designed-in feature.

AlumiuN:
That's fair. The reason I ask is that it is conceivable that said glitch could be part of a map puzzle, but I think it's unlikely enough that unless someone finds a level where it's used, it's probably not worth worrying about (I haven't even personally seen a map that uses it intentionally yet, I mainly investigated the system because I thought slightly weird behaviour was better than crashing in levels that inadvertantly had the issue).

PR #497, opened 2020-02-06
Fix display of hostages lost on level end screen

teratorn:
if the player has more hostages_on_board than exist in the level,
perhaps because they picked up additional hostages in a secret level,
then the calculation here will underflow and tell you that you lost
a very large number of hostages, when in fact you collected all the
hostages (and then some)

### DIFF
--- a/similar/main/gameseq.cpp
+++ b/similar/main/gameseq.cpp
@@ -1180,8 +1180,8 @@ static void DoEndLevelScoreGlitz()
 	auto &hostage_text = m_str[c++];
 	if (cheats.enabled)
 		snprintf(hostage_text, sizeof(hostage_text), "Hostages saved:   \t%u", hostages_on_board);
-	else if (const auto hostages_lost = LevelUniqueObjectState.total_hostages - hostages_on_board)
-		snprintf(hostage_text, sizeof(hostage_text), "Hostages lost:    \t%u", hostages_lost);
+	else if (hostages_on_board < LevelUniqueObjectState.total_hostages)
+		snprintf(hostage_text, sizeof(hostage_text), "Hostages lost:    \t%u", LevelUniqueObjectState.total_hostages - hostages_on_board);
 	else
 	{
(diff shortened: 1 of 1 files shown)


vLKp:
Wouldn't it be better to track total-hostages correctly, so that the count does not overflow _and_ the player is accurately rewarded/penalized based on whether the secret level hostages are recovered?

teratorn:
Yes. This isn't a perfect fix. e.g. you could miss some or all of the level hostages, and pick up all the secret level hostages, and still potentially be told that you didn't lose any.

To do it properly you would have to remember if the player ever entered the secret level. You don't want to be penalized for missing hostages in a secret level if you never even found the secret level entrance, right?

vLKp:
Right, although that particular concern is mitigated by the way the secret level transition is handled.  I think if you never use the secret level exit, then the game will never read the secret level's data, and will not know how many hostages were on the secret level.

I was thinking of the following rules:
- On entering a secret level, increase `total_hostages` by the number of hostages on the level.
- On exiting a secret level:
  - If the countdown is not running, decrease `total_hostages` by the number of hostages left behind.  This allows the player to enter and exit the secret level freely without the `total_hostages` count growing without bound.  This is slightly incorrect, since it means that a player can ignore hostages on a secret level and not be penalized for it when ending the main level.  Fixing it properly would require remembering whether the player has visited the secret level, and only updating `total_hostages` on first entry.
  - If the countdown is running, the secret level becomes inaccessible, and `total_hostages` is not reduced.  Any hostages not collected are lost, and the player is penalized for not recovering them prior to destroying the secret level.

vLKp:
This pull request was left open to track the more thorough changes required for proper hostage accounting discussed above, because at the time I believed this had never been right, and so should be fixed properly, rather than ignoring the improper tally at level exit time.  No response from original reporter.  Issue #648 subsequently re-reported the problem, and noted that it was a regression relative to 0.58.1.  The regression was fixed for that issue.  At this point, it seems unlikely anyone will work on the full fix, so closing.  Please reopen if you wish to pursue the full fix.

PR #501, opened 2020-04-04
Add PIGGY_PAGE_IN to ogl_cache_polymodel_textures

arbruijn:
Fix data = NULL crash in ogl_loadtexture after piggy_bitmap_page_out_all() ran at an unfortunate time. Happens with an unreleased level by TRUEpiiiicness.
https://descentbb.net/viewtopic.php?f=22&t=24641


vLKp:
Most other uses of `ogl_loadbmtexture` are preceded by that call, so it seems reasonable to have it here too.  Merged.

PR #512, opened 2020-06-09
clipper.cpp: Fix build with gcc10

schnitzeltony:
| common/3d/clipper.cpp:29:14: error: 'out_of_range' is not a member of 'std'

Signed-off-by: Andreas Müller <schnitzeltony@gmail.com>


vLKp:
That file is only used in the SDL-only build, which sees less attention than the GL build.  The fix is obviously correct.  Merged.

Issue #469, opened 2019-11-09
Allow navigating menu with a gamepad/joystick


GadWissberg:
Hello, I've been using DXX-Rebirth for a long time. Now I started using it with RetroPie on my TV using gamepad only. I came across a problem where I can't navigate the menus with the controller. This forces me to have a keyboard connected to my Pi only for navigating the menu.
I'd be happy to have an option to navigate the menus too with the gamepad.
Thanks

vLKp:
This was [requested a year ago on the Rebirth forum](https://forum.dxx-rebirth.com/showthread.php?tid=1085).  The summary version is that I have no interest in writing such a feature for myself, and none of the people who want it have done anything to advance it.  I suggested in that thread (and quoted below, since the forum will close Real Soon Now) that a good first step would be for someone to write up a requirements document on what exactly the feature should do.  One user responded saying he would.  If he did, he never posted back to announce its completion.

My position posted then stands:
- I have no use for it.
- I have other things I use my time to work on.
- If someone else writes it, I will review it for inclusion.
- If someone writes a good enough requirements document, I _might_ work on the feature despite having no use for it.


> From the forum:
---
> No promises on when or if this gets implemented, but if you want to help, a great way to start would be to write up a requirements document.  From the posts here, I understand that interested users want the game to be usable with just a joystick, without using any other input devices (keyboard, mouse).  There is considerable wiggle room in satisfying that generality, and at least some ways of satisfying it probably won't satisfy the people who want to use it.  For example:
> * Is it acceptable if you need to use joystick buttons, not axes, to navigate?  Axes are analog inputs, so if an axis can be used, then the navigation code is more complex because movement below a certain joystick-specific threshold is noise and needs to be ignored, but movement above that threshold is user action and needs to be handled.  Buttons are digital inputs, so we can assume that all button presses are deliberate user action and all button presses should be handled.
> * If button-only is acceptable, how should the end user specify these buttons?  Does Rebirth ship with a fixed list and you are required to use those, with no ability to change them without recompiling the game?  Does Rebirth load user customizations and, if it does, how does the user enter those customizations?  Can we assume the user has a keyboard for entering complex configuration files, but he chooses to use the keyboard <u>only</u> for data entry, not when actually playing the game?
>     * If we can make that assumption, then it's easy to say that the user who wants customizations must use the keyboard to change how the buttons map to actions before he can play the game.
[...]

PR #516, opened 2020-06-29
make menus controllable via joystick

kajott:
This patchset allows all parts of d1x and d2x to be controlled with a joystick or gamepad. Menus can be navigated, and a new configurable joystick button binding is added to open the in-game menu. This should fix issue #469.
Please see the individual commit messages for details.


common/arch/sdl/joy.cpp
```diff
@@ -466,5 +467,48 @@ int apply_deadzone(int value, int deadzone)
...
+bool joy_translate_menu_key(const d_event &event) {
+	if (event.type != EVENT_JOYSTICK_BUTTON_DOWN)
+		return false;
+#if DXX_MAX_BUTTONS_PER_JOYSTICK || DXX_MAX_HATS_PER_JOYSTICK || DXX_MAX_AXES_PER_JOYSTICK
+	// Find out which button has been pressed by decoding the button
+	// description text. This might not be the ideal way to do things,
+	// but we currently don't have data structures that allow us to do
+	// such a mapping in a saner way, so we'll take what we can get.
+	auto &e = static_cast<const d_event_joystickbutton &>(event);
+	Assert((e.button >= 0) && (e.button < joybutton_text.size()));
```
vLKp, line 479: Please use standard C assert, not the macro inherited from the original Descent source.

vLKp, line 479: `e.button` is `unsigned`, so it is always `>= 0`.  Some compiler configurations may warn about a comparison that can never fail.

common/arch/sdl/joy.cpp
```diff
@@ -466,5 +467,48 @@ int apply_deadzone(int value, int deadzone)
 		return 0;
 }
 
+bool joy_translate_menu_key(const d_event &event) {
+	if (event.type != EVENT_JOYSTICK_BUTTON_DOWN)
+		return false;
+#if DXX_MAX_BUTTONS_PER_JOYSTICK || DXX_MAX_HATS_PER_JOYSTICK || DXX_MAX_AXES_PER_JOYSTICK
+	// Find out which button has been pressed by decoding the button
```
vLKp, line 474: I dislike this approach.  If you need new data structures to track this properly, please add them.

common/arch/sdl/joy.cpp
```diff
@@ -466,5 +467,48 @@ int apply_deadzone(int value, int deadzone)
 		return 0;
 }
 
+bool joy_translate_menu_key(const d_event &event) {
+	if (event.type != EVENT_JOYSTICK_BUTTON_DOWN)
+		return false;
+#if DXX_MAX_BUTTONS_PER_JOYSTICK || DXX_MAX_HATS_PER_JOYSTICK || DXX_MAX_AXES_PER_JOYSTICK
```
vLKp, line 473: I think this could be simplified to `#if DXX_MAX_JOYSTICKS`.  `SConstruct` ensures that the maximum number of joysticks is forced to `0` if none of the inputs are enabled.

https://github.com/dxx-rebirth/dxx-rebirth/blob/096a678ff759dd19f5a1d50036931926f17944e1/SConstruct#L1423-L1424

common/arch/sdl/key.cpp
```diff
@@ -587,6 +587,11 @@ void key_flush()
 		restore_sticky_key(keystate, key);
 }
 
+void event_keycommand_send(int key) {
+	const d_event_keycommand event{EVENT_KEY_COMMAND, unsigned(key)};
```
vLKp, line 591: Instead of using a function-style cast, make the parameter the right type on entry.

common/include/joy.h
```diff
@@ -36,6 +36,7 @@ extern void joy_init();
 extern void joy_close();
 const d_event_joystick_axis_value &event_joystick_get_axis(const d_event &event);
 extern void joy_flush();
+extern bool joy_translate_menu_key(const d_event &event);
```
vLKp, line 39: `extern` is unnecessary here.  The existing declarations have not been stripped, but there is no need to include it in new ones.

similar/main/titles.cpp
```diff
@@ -1529,6 +1536,24 @@ static window_event_result briefing_handler(window *, const d_event &event, brie
 			}
 			break;
 
+		case EVENT_JOYSTICK_BUTTON_DOWN:
+			// using joy_translate_menu_key doesn't work here for unclear
```
vLKp, line 1540: How does it fail?

common/arch/sdl/joy.cpp
kajott: OK, I will change this. I tried to keep it consistent with the surrounding source code, not knowing that the upper-case `Assert`s are legacy.

common/arch/sdl/joy.cpp
kajott: OK. Would a direct mapping from joystick buttons to key codes be acceptable? That would be the easiest way, I think.

common/arch/sdl/joy.cpp
kajott: OK. Again, this was copied from other parts of the code.

common/include/joy.h
kajott: OK. Again, I tried to follow the surrounding code :)

similar/main/titles.cpp
kajott: The game hangs with a black screen on level entry (before the "Prepare for Descent" box appears, but after the briefing screens). The stack trace shows multiple levels of calls within the event system; to be honest, I just didn't investigate further at this point and came up with that alternative approach instead.

kajott:
The new commit should resolve the coding issues. A joybutton-to-key mapping table is now used; it's populated at the same time as the button description string list.

I also added another minor feature that I've overlooked previously, but was simple to add and should cause no harm.

common/arch/sdl/joy.cpp
vLKp: Yes.

common/arch/sdl/joy.cpp
vLKp: I will fix the existing versions to eliminate this misleading hint.

common/include/joy.h
vLKp: That's fine.  I appreciate the attempt to maintain consistency.  Rebirth hasn't been truly consistent for a long time though. ;)

similar/main/titles.cpp
vLKp: That stack trace is somewhat reasonable if the game is waiting for something to make forward progress.  Perhaps it got confused and was not able to make forward progress.

common/arch/sdl/joy.cpp
vLKp: I see the problem.  A previously contributed improvement blurred the distinction between axes and buttons, and required some preprocessor adjustments to ensure the right variables were present.  Prior to that, the tests would be for (buttons || hats), but not axes.  To support that change, the contributor added `|| axes` in various places, and I did not ask for it to be handled differently then, so now there are constructs which can mislead readers.

similar/main/state.cpp
```diff
@@ -922,7 +922,12 @@ static d_game_unique_state::save_slot state_get_savegame_filename(d_game_unique_
 		const auto uchoice = static_cast<std::size_t>(choice);
 		fname = filename[uchoice];
 		if (dsc)
+		{
 			*dsc = desc[uchoice];
+			if (!dsc->data()[0]) {
+				strcpy(dsc->data(), "-no title-");
```
vLKp, line 928: It might be more useful to generate a descriptive string here.  For autosaves, I insert _month-day hour:minute:second_.

https://github.com/dxx-rebirth/dxx-rebirth/blob/a20551191afc87b4fc7e2afdeb97c254458f0984/similar/main/state.cpp#L691-L695

You probably don't want the literal text `auto` for the case of a joystick-entered save.  Note that the maximum description length prevents using a verbose description.

I can handle changing this later.

common/arch/sdl/joy.cpp
```diff
@@ -372,10 +376,19 @@ void joy_init()
...
 			const auto n_hats = check_warn_joy_support_limit(SDL_JoystickNumHats(handle), "hat", DXX_MAX_HATS_PER_JOYSTICK);
 
-			joybutton_text.resize(joybutton_text.size() + n_buttons + (4 * n_hats) + (2 * n_axes));
+			const auto n_virtual_buttons = n_buttons + (4 * n_hats) + (2 * n_axes);
+			joybutton_text.resize(joybutton_text.size() + n_virtual_buttons);
+			joy_key_map.resize(joy_key_map.size() + n_virtual_buttons, 0);
 #if DXX_MAX_BUTTONS_PER_JOYSTICK
 			range_for (auto &&e, enumerate(partial_range(joystick.button_map(), n_buttons), 1))
 			{
+				switch (e.idx) {
```
vLKp, line 385: General style is that `switch` `case` statements are multiline.  No need to change that for this review, though.

common/arch/sdl/joy.cpp
```diff
@@ -372,10 +376,19 @@ void joy_init()
 			const auto n_buttons = check_warn_joy_support_limit(SDL_JoystickNumButtons(handle), "button", DXX_MAX_BUTTONS_PER_JOYSTICK);
 			const auto n_hats = check_warn_joy_support_limit(SDL_JoystickNumHats(handle), "hat", DXX_MAX_HATS_PER_JOYSTICK);
 
-			joybutton_text.resize(joybutton_text.size() + n_buttons + (4 * n_hats) + (2 * n_axes));
+			const auto n_virtual_buttons = n_buttons + (4 * n_hats) + (2 * n_axes);
+			joybutton_text.resize(joybutton_text.size() + n_virtual_buttons);
+			joy_key_map.resize(joy_key_map.size() + n_virtual_buttons, 0);
```
vLKp, line 381: The single argument form of `std::vector<T>::resize` would zero-initialize the new members, so there is no need to explicitly use 0.

No need to change this.

common/arch/sdl/joy.cpp
```diff
@@ -372,10 +376,19 @@ void joy_init()
...
 #if DXX_MAX_BUTTONS_PER_JOYSTICK
 			range_for (auto &&e, enumerate(partial_range(joystick.button_map(), n_buttons), 1))
 			{
+				switch (e.idx) {
+					case 1: joy_key_map[joystick_n_buttons] = KEY_ENTER; break;
+					case 2: joy_key_map[joystick_n_buttons] = KEY_ESC; break;
+					case 3: joy_key_map[joystick_n_buttons] = KEY_SPACEBAR; break;
+					case 4: joy_key_map[joystick_n_buttons] = KEY_DELETE; break;
+					default: break;
+				}
```
vLKp, line 391: This switch could be avoided by keeping a `static const` table and pulling the values from there:

```suggestion
    static const unsigned initial_buttons[] = {
        KEY_ENTER,
        KEY_ESC,
        KEY_SPACEBAR,
        KEY_DELETE
    };
    if (e.idx < std::size(initial_buttons)) {
        joy_key_map[joystick_n_buttons] = initial_buttons[e.idx];
    }
```

vLKp:
For the record, I consider the whole approach of modifying the game engine to synthesize input to be undesirable because it must be replicated in each game that users wish to use, and done separately for each input device in each game.  I think users who want a joystick-only (or mouse-only, or touchscreen-only, ...) interface should configure their input device(s) to synthesize appropriate keyboard input at the OS level, because then the input device will do the right and useful thing for _every_ game, even closed-source games, after one round of configuration.  Patching the game to handle these devices specially is inherently unscaleable.

That said, you have done the work, and I see no technical fault with the work, so I see no reason not to merge it.

similar/main/state.cpp
kajott: I agree that it would be a neat feature, but then we'd also need some kind of tracking or detection that a save slot has such an auto-generated name and update the timestamp when saving to the same slot later on. Otherwise, the slot would always keep the timestamp from when it was first created, which is not _that_ more helpful than the current "no title" ;)

common/arch/sdl/joy.cpp
kajott: Sorry for that -- muscle memory from my usual coding style :)

common/arch/sdl/joy.cpp
kajott: I thought the same, but in the end opted for "better explicit than imprecise".

common/arch/sdl/joy.cpp
kajott: You're right, that would be much better. Shall I change it and make a new PR?

kajott:
I can see the scaling issue you described, but I'd like to spin it the other way round: Requiring each user to set up OS-level input remapping tools doesn't scale either :)

For example, this PR is a direct and immediate improvement for all users of the popular RetroPie distribution, whose control scheme is heavily joystick-centric. Until now, you needed an additional keyboard every time you wanted to play DXX-Rebirth there; with this PR, you need a keyboard exactly *once* for initial setup to enter a pilot name.

vLKp:
For a standardized and widely used system, such as RetroPie, the input mapping could and probably should be done by the distribution maintainer, so that everything is ready after installation.  Individual users would not need to adjust anything, unless they disagreed with the distribution's default bindings.

PR #517, opened 2020-06-30
Prevent the destruction of one-sided prison doors from causing a crash

AlumiuN:
Currently, if a prison door is one-sided, the destruction of that side of the prison door will always attempt to process the destruction for a wall on the other side even if one isn't present, which causes a crash. This should fix the problem in an acceptable manner.

### DIFF
--- a/similar/main/fireball.cpp
+++ b/similar/main/fireball.cpp
@@ -1380,25 +1380,31 @@ unsigned do_exploding_wall_frame(wall &w1)
 
 		auto &Walls = LevelUniqueWallSubsystemState.Walls;
 		auto &vmwallptr = Walls.vmptr;
-		auto &w2 = *vmwallptr(csegp->shared_segment::sides[cside].wall_num);
-		assert(&w1 != &w2);
-		assert((w1.flags & WALL_EXPLODING) || (w2.flags & WALL_EXPLODING));
-		w1.flags |= WALL_BLASTED;
-		w2.flags |= WALL_BLASTED;
-		if (w1_explode_time_elapsed >= EXPL_WALL_TIME)
+
+		auto cwall_num = csegp->shared_segment::sides[cside].wall_num;
+		if (cwall_num != wall_none)
 		{
-			if (w1.flags & WALL_EXPLODING)
-			{
-				w1.flags &= ~WALL_EXPLODING;
-				++ walls_updated;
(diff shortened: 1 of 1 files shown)


vLKp:
This is worth fixing, but I must ask: when would a level author use a one-sided blastable door?

similar/main/fireball.cpp
```diff
@@ -1380,25 +1380,31 @@ unsigned do_exploding_wall_frame(wall &w1)
...
 		auto &Walls = LevelUniqueWallSubsystemState.Walls;
 		auto &vmwallptr = Walls.vmptr;
-		auto &w2 = *vmwallptr(csegp->shared_segment::sides[cside].wall_num);
-		assert(&w1 != &w2);
-		assert((w1.flags & WALL_EXPLODING) || (w2.flags & WALL_EXPLODING));
-		w1.flags |= WALL_BLASTED;
-		w2.flags |= WALL_BLASTED;
-		if (w1_explode_time_elapsed >= EXPL_WALL_TIME)
+
+		auto cwall_num = csegp->shared_segment::sides[cside].wall_num;
```
vLKp, line 1384: I generally make local variables `const` wherever possible, but there is no need to rework this for that.

similar/main/fireball.cpp
```diff
@@ -1380,25 +1380,31 @@ unsigned do_exploding_wall_frame(wall &w1)
...
+		if (cwall_num != wall_none)
```
vLKp, line 1385: Now that C++17 is used, you can use an init-statement in a selection statement.  I have not made an effort to make this conversion in old code, though.

vLKp:
This looks good.  I will merge it once I have time to do a smoke test.

AlumiuN:
It's not so much "when would they choose to use it" as "when they accidentally use it". I know there are multiple D1 levels where it's the case (Descend Again level 13 being my motivation for fixing this). I'll remember the notes for next time :)

vLKp:
As expected, the game is stable with this change and two-sided doors blast out as normal.  Merged.

Thank you for working on this.  There are many ways that ill-formed levels can break the engine, and closing them off is always a nice improvement.

AlumiuN:
You're welcome; I'm just happy I'm actually able to be helpful. :)

Issue #515, opened 2020-06-29
Runtime crash: Complicated and unlikely scenario involving D1 one-shot triggers leads to a level crashing


AlumiuN:
Tested on 096a678, Win10, 32-bit EXE.

When a D1 level is exited via a standard exit, check_trigger will still run through the routine for processing the one-shot-ness of the trigger in question, meaning that if the level being entered has a one-shot trigger with the same number as the exit trigger from the previous level, the seg and side arguments passed into check_trigger are no longer relevant but will still be used.
The one level that I've found where this happens is [Descend Again](https://www.enspiar.com/dmdb/viewMission.php?id=200) level 6->7, and it causes a crash as a result.

vLKp:
That is definitely an obscure situation.  Did you find this via code inspection and seek out a level where it happens, or get the crash in the level and debug from there?

Either way, this warrants fixing.

AlumiuN:
I got the crash while playing the level and debugged from there.

AlumiuN:
After a quick look at the original D1 source, I think slapping everything in that block after `t.flags &= ~TRIGGER_ON;` inside an `if (IS_CHILD(seg->children[side]))` block is probably the way to stop it crashing while maintaining the behaviour of D1.

vLKp:
Your edit seems correct.  I cannot find any indication that TRIGGER_ON is meaningful, and when I removed it locally, the game still built.  I see that D2 has a trigger flag `TF_DISABLED`, which suggests that the trigger disabling logic in D1 was not working and that `TF_DISABLED` was how they fixed it.

vLKp:
I pushed an alternative fix of removing the offending code.  More detailed discussion is included in the pull request (#518).  I believe that fix will resolve this crash, so I am closing this issue.  If you still see problems, please reopen.

PR #518, opened 2020-07-02
Prevent extremely uncommon crash upon D1 level load… (fixes issue #515)

AlumiuN:
…when one-shot trigger has same trigger number as previous level's exit trigger.

An alternative solution would be to remove the one-shot block entirely, as it currently has no effect on gameplay.

### DIFF
--- a/similar/main/switch.cpp
+++ b/similar/main/switch.cpp
@@ -570,17 +570,20 @@ window_event_result check_trigger(const vcsegptridx_t seg, const unsigned side, 
 		{
 			t.flags &= ~TRIGGER_ON;
 	
-			const shared_segment &csegp = *vcsegptr(seg->children[side]);
-			auto cside = find_connect_side(seg, csegp);
-			Assert(cside != side_none);
-		
-			const auto cwall_num = csegp.sides[cside].wall_num;
-			if (cwall_num == wall_none)
-				return window_event_result::ignored;
+			if (IS_CHILD(seg->children[side]))
+			{
+				const shared_segment &csegp = *vcsegptr(seg->children[side]);
+				auto cside = find_connect_side(seg, csegp);
+				Assert(cside != side_none);
 			
-			const auto ctrigger_num = vcwallptr(cwall_num)->trigger;
-			auto &ct = *vmtrgptr(ctrigger_num);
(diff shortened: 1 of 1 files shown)


vLKp:
I removed the block entirely, as you suggested in the second paragraph.

Looking at the code, I think adding an `IS_CHILD` check would have been insufficient to completely protect against a crash.  Consider the case that:

- the latter level happens to have a one-shot trigger at `Triggers[trigger_num]` (as the level you reported did)
  https://github.com/dxx-rebirth/dxx-rebirth/blob/a27b50a357a3e9fa6829405e5fc0bdcff8b7ae22/similar/main/switch.cpp#L569

- Unlike the case you reported, the hypothetical trigger is a flythrough trigger, so the proposed `IS_CHILD(seg->children[side])` succeeds.  Suppose further that because the level uses a flythrough trigger here, `csegp` is valid, and the connected segment has a wall, so `if (cwall_num == wall_none)` does not execute.
  https://github.com/dxx-rebirth/dxx-rebirth/blob/a27b50a357a3e9fa6829405e5fc0bdcff8b7ae22/similar/main/switch.cpp#L578
  Since the wall is valid, `vcwallptr` succeeds.
  https://github.com/dxx-rebirth/dxx-rebirth/blob/a27b50a357a3e9fa6829405e5fc0bdcff8b7ae22/similar/main/switch.cpp#L581
  However, the flythrough exists only on the origin side, and there is no trigger on the connected segment, so `->trigger` is `trigger_none`.  `vmtrgptr` would then throw.
  https://github.com/dxx-rebirth/dxx-rebirth/blob/a27b50a357a3e9fa6829405e5fc0bdcff8b7ae22/similar/main/switch.cpp#L582

On further consideration, the above scenario could occur even without the obscure level reuse problem.  Any time a one-shot trigger occurs on a bidirectional wall, but with no trigger on the reverse wall, this path could be hit, and would crash.  Thus, removing the logic seems to be the right fix, since `TRIGGER_ON` was write-only.

I pushed a commit that removes the code, and cited this pull request.  I think that is sufficient, so I am closing this pull and the associated issue.  If you disagree, please reopen.

AlumiuN:
Yep, good catch.

PR #519, opened 2020-07-02
Raspberry Pi bugfixes

kajott:
Fixes for two bugs that plagued DXX-Rebirth on Raspberry Pi systems:
- empty automap (no lines drawn)
- Medium Hulk and Heavy Hulk models were missing their heads


vLKp:
Thank you for the report and the proposed fixes.  I elected to fix the polymodel issue a bit differently, by instead reloading the variable after the call to `align_polygon_model_data`, so that the reload occurs only on platforms that need it.

As a general testing note for future readers, this problem could be triggered on any platform by explicitly enabling `words_need_alignment=1` at build time.

Issue #477, opened 2019-12-10
Runtime bug: Sound samples loop in demo playback


zicodxx:
### Environment

This seems to occur with the https://github.com/dxx-rebirth/dxx-rebirth/commit/ca383c1fe76bf519f3064d2188d77ab54f2b6b42. Tested on Windows 10.

#### Operating System Environment

* Microsoft Windows
  * [ ] Microsoft Windows XP
  * [ ] Microsoft Windows Vista
  * [ ] Microsoft Windows 7
  * [ ] Microsoft Windows 8
  * [ ] Microsoft Windows 8.1
  * [X] Microsoft Windows 10

* Mac OS X
  * [ ] Mac OS X 10.8
  * [ ] Mac OS X 10.9
  * [ ] Mac OS X 10.10
  * [ ] Mac OS X 10.11
  * [ ] Mac OS X 10.12
  * [ ] Mac OS X 10.13

* Debian
  * [ ] Debian Wheezy
  * [ ] Debian Jessie
  * [ ] Debian Stretch
  * [ ] Debian Buster
* Fedora
  * [ ] Fedora 26
  * [ ] Fedora 27
  * [ ] Fedora 28
  * [ ] Fedora 29
  * [ ] Fedora 30
  * [ ] Rawhide
* Ubuntu
  * [ ] Ubuntu 16.04 LTS (Xenial Xerus)
  * [ ] Ubuntu 16.10 (Yakkety Yak)
  * [ ] Ubuntu 17.04 (Zesty Zapus)
  * [ ] Ubuntu 17.10 (Artful Aardvark)
  * [ ] Ubuntu 18.04 LTS (Bionic Beaver)
  * [ ] Ubuntu 18.10 (Cosmic Cuttlefish)
  * [ ] Ubuntu 19.04 (Disco Dingo)

* [ ] Arch
* [ ] Gentoo

* [ ] OpenBSD
* [ ] FreeBSD
* [ ] NetBSD

#### CPU environment

* [ ] x86 (32-bit Intel/AMD)
* [X] x86\_64 (64-bit Intel/AMD)
* [ ] ARM (32-bit)
* [ ] ARM64 (64-bit; sometimes called AArch64)

#### Game environment

The issue seems to be present in D1X-Rebirth when playing back a recorded demo from Descent: First Strike. Demo attached for confirmation. 

[test.zip](https://github.com/dxx-rebirth/dxx-rebirth/files/3942340/test.zip)

### Description

[...]

PR #530, opened 2020-08-05
Workaround for missing forever flag in sound object demo recording

arbruijn:
In the original code there was a 'Hack to keep sounds from building up...'
which changed a non-forever linked sound to a non-linked sound. Therefore
the forever flag wasn't needed in the demo files. The commit
4a98e796abb099eb4c592d9520e4edbef7a15d7e removed the hack, which caused
non-forever sounds to start looping on demo playback.
This commit restores the hack as far as demo files are concerned.

Reported-by: zicodxx <https://github.com/dxx-rebirth/dxx-rebirth/issues/477>
Fixes: 4a98e796abb099eb4c592d9520e4edbef7a15d7e ("Prevent stacking weapon rotation sounds")

### DIFF
--- a/similar/main/digiobj.cpp
+++ b/similar/main/digiobj.cpp
@@ -418,7 +418,15 @@ void digi_link_sound_to_object3(const unsigned org_soundnum, const vcobjptridx_t
 	}
 
 	if ( Newdemo_state == ND_STATE_RECORDING )		{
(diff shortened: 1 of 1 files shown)


vLKp:
Thank you for the diagnosis and the fix.  Merged.

PR #536, opened 2020-08-28
Ensure cached cockpit textures are valid

arbruijn:
Check if the cockpit window textures are still valid when reusing the
previous cockpit data in cockpit_decode_alpha.

This fixes a crash when leaving the graphics options menu while in
the game with cockpit visible, since that calls
ogl_smash_texture_list_internal which invalidates all textures.


vLKp:
Thank you for the fix.  I still see a use-after-free in some font code in this path, but that has probably always been there.

PR #540, opened 2020-09-09
Add SConstruct option to allow using libraries and headers instead of frameworks on macOS

Kreeblah:
This gives people another option who use Homebrew or other package managers on macOS which don't include framework builds of the SDL libraries, and just include the regular libraries and headers.

Part of the driver for this is making building easier for people, and part of the driver is to be able to get the project added to the Homebrew repository.  They don't have the framework libraries installed on their build hosts, and while I'm looking into finding ways to build them as part of the SDL packages, that's a fair bit of effort.  In the meantime, this would let their build hosts build the project with just the libraries that they're already building.

### DIFF
--- a/INSTALL.markdown
+++ b/INSTALL.markdown
@@ -127,15 +127,15 @@ DXX-Rebirth can be built from the Terminal (via SCons) without Xcode; to build u
 
 ##### [Homebrew](https://github.com/Homebrew/homebrew/)
 * **brew install
- gcc5
  scons
  sdl
  sdl\_image
  sdl\_mixer
+ libpng
  physfs**
 
 ### Building
-Once prerequisites are installed, run **scons** *options* to build.  By default, both D1X-Rebirth and D2X-Rebirth are built.  To build only D1X-Rebirth, run **scons d1x=1**.  To build only D2X-Rebirth, run **scons d2x=1**.
+Once prerequisites are installed, run **scons** *options* to build.  By default, both D1X-Rebirth and D2X-Rebirth are built.  To build only D1X-Rebirth, run **scons d1x=1**.  To build only D2X-Rebirth, run **scons d2x=1**.  If using macOS without framework builds of the SDL libraries (such as with Homebrew), run **scons mac_libs=1**.
 
 If unspecified, **SConstruct** uses $CXX for the compiler, $CPPFLAGS for preprocessor options, $CXXFLAGS for C++ compiler options, and $LDFLAGS for linker options.  You may override any or all of these choices by setting options of the corresponding name: **scons CXX=/opt/my/c++ 'CXXFLAGS=-O0 -ggdb'**.  **SConstruct** supports numerous options for adjusting build details.  Run **scons -h** to see them.  Option names are case sensitive.  Commonly used options include:
 
--- a/SConstruct
+++ b/SConstruct
@@ -1530,10 +1530,12 @@ static void terminate_handler()
 		}
 		successflags = self.pkgconfig.merge(context, self.msgprefix, user_settings, library_name, library_name, guess_flags)
 		if user_settings.host_platform == 'darwin':
-			successflags = successflags.copy()
-			successflags['FRAMEWORKS'] = [library_name]
-			relative_headers = 'Library/Frameworks/%s.framework/Headers' % library_name
-			successflags['CPPPATH'] = [h for h in (os.path.join(os.getenv("HOME"), relative_headers), '/%s' % relative_headers) if os.path.isdir(h)]
+			macos_libs_only = ARGUMENTS.get('mac_libs', 0)
+			if macos_libs_only == 0:
+				successflags = successflags.copy()
+				successflags['FRAMEWORKS'] = [library_name]
+				relative_headers = 'Library/Frameworks/%s.framework/Headers' % library_name
+				successflags['CPPPATH'] = [h for h in (os.path.join(os.getenv("HOME"), relative_headers), '/%s' % relative_headers) if os.path.isdir(h)]
 		# SDL2 headers still use SDL_*.h for their filename, not
 		# SDL2_*.h, so expanded library_format_name with an explicitly
 		# blank insert, regardless of whether building for SDL1 or SDL2.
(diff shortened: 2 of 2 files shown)


SConstruct
```diff
@@ -1530,10 +1530,12 @@ static void terminate_handler()
 		}
 		successflags = self.pkgconfig.merge(context, self.msgprefix, user_settings, library_name, library_name, guess_flags)
 		if user_settings.host_platform == 'darwin':
-			successflags = successflags.copy()
-			successflags['FRAMEWORKS'] = [library_name]
-			relative_headers = 'Library/Frameworks/%s.framework/Headers' % library_name
-			successflags['CPPPATH'] = [h for h in (os.path.join(os.getenv("HOME"), relative_headers), '/%s' % relative_headers) if os.path.isdir(h)]
+			macos_libs_only = ARGUMENTS.get('mac_libs', 0)
```
vLKp, line 1533: This is not the best way to add a user-settable option.  As written, `mac_libs` is a top-level parameter that applies to every build, it does not show up in `--help` output, and it cannot be configured via the `site` stored preferences.  Generally, I try to have parameters handled through the user_settings support code, which will address all these points automatically.

INSTALL.markdown
```diff
@@ -127,15 +127,15 @@ DXX-Rebirth can be built from the Terminal (via SCons) without Xcode; to build u
 
 ##### [Homebrew](https://github.com/Homebrew/homebrew/)
 * **brew install
- gcc5
```
vLKp, line 130: Removing gcc5 is correct, since the minimum is now gcc-7.  Does Homebrew guarantee a new enough gcc that it is appropriate not to mention `gcc` at all in this list?

INSTALL.markdown
```diff
@@ -127,15 +127,15 @@ DXX-Rebirth can be built from the Terminal (via SCons) without Xcode; to build u
...
  scons
  sdl
  sdl\_image
  sdl\_mixer
+ libpng
  physfs**
 
 ### Building
-Once prerequisites are installed, run **scons** *options* to build.  By default, both D1X-Rebirth and D2X-Rebirth are built.  To build only D1X-Rebirth, run **scons d1x=1**.  To build only D2X-Rebirth, run **scons d2x=1**.
+Once prerequisites are installed, run **scons** *options* to build.  By default, both D1X-Rebirth and D2X-Rebirth are built.  To build only D1X-Rebirth, run **scons d1x=1**.  To build only D2X-Rebirth, run **scons d2x=1**.  If using macOS without framework builds of the SDL libraries (such as with Homebrew), run **scons mac_libs=1**.
```
vLKp, line 138: The phrasing here looks awkward to me.  The rest of the paragraph is very general instruction, and then one platform-specific parameter is advertised here.  None of the other user settings are advertised in this file, whether or not platform-specific.  I think it would be better to have the `--help` output mention this option, then add some general text here telling the user to see `--help` for a list of options, and possibly add some fallback handling in the SConf tests so that on failure to find the framework, the test suggests the option to ignore frameworks.  I don't know if that last point is easy though.

vLKp (changes requested):
I think the overall idea is a good one, but the implementation could use some adjustment.  If you like, I can make those changes on top of the merge when I put it in.

INSTALL.markdown
Kreeblah: Since Homebrew requires the Xcode command-line tools to be installed in order to run properly, installing packages from Homebrew guarantees a recent version of clang.  I don't have the any gcc packages installed at all (just relying on the clang suite), and it builds just fine on my machine.  That said, if gcc were needed for some reason, the gcc package in Homebrew currently installs 10.2.0, and is generally kept pretty up to date.

SConstruct
Kreeblah: I was wondering about that.  Initially I tried assigning it to user_settings.macos_libs_only, but I kept running into scoping issues, with Python telling me that either self or user_settings wasn't valid in the scope I was attempting to use it and I wasn't really sure why.  I'm not hugely familiar with Python, though, so tracking that down was a bit of a challenge.

So, I figured getting something working like this as a proof of concept would at least be a first step, and then talk about it if needed.

INSTALL.markdown
Kreeblah: That's fair.  I wasn't sure where a good place would be to note it, but I guess the help output would make sense.

Regarding fallback handling, there are definitely ways it could be more robust.  In my ideal world, the tests would test for an equivalent library if it failed to find a framework and use that instead if it finds it, but that was more complicated a task than I really wanted to tackle right now.

Kreeblah:
If you'd be up for making the changes on top of the merge, sure, that'd be fantastic.  Thank you!

And, thanks for reviewing this.  I really appreciate you taking the time for it.

vLKp:
I pushed [experimental/540/macos-no-framework](https://github.com/dxx-rebirth/dxx-rebirth/tree/experimental/540/macos-no-framework) with my proposed changes.  It's based on the first commit of your two, and drops the second since that only contained the README changes.  The version I pushed is a valid build script that runs on Linux, but I cannot test whether it does what you want on OS X.  Please test it and, if you are satisfied with how it works, I will merge it to `master`.

Kreeblah:
Yep, I just built D1X and D2X with the flag you added, and they build and run fine using only Homebrew-provided libraries.  I'd say it's good to merge.

Thank you for your help with this!

SConstruct
vLKp: That's fine.  Both `self` and `user_settings` are local variables, and may not be visible in all the places you wanted to use them.  I try to keep consistent naming for `user_settings`, so if it is available, it will be called that.

INSTALL.markdown
vLKp: That makes sense.  I can't do the fallback work for you since I have no way to test it.  However, I hope having the flag that you proposed will make Homebrew easy enough to use that the fallback is not needed.

vLKp:
I merged that change to `master`.  On reviewing this pull request, I see that by dropping the change to INSTALL.md, I dropped your removal of gcc5.  That should go in.  I will handle that when I next work on this.

vLKp:
The change to INSTALL is now pushed.  I believe that resolves this pull.

PR #543, opened 2020-09-14
Fixed macOS compilation by bypassing GL library check

Kreeblah:
Commit https://github.com/dxx-rebirth/dxx-rebirth/commit/969caa8c0cb5de8a8af29e7368ca2d21c02e34d7 (which didn't get captured in that experimental branch) broke compilation on macOS.  In macOS, the OpenGL headers and libraries are included in a framework that's bundled as part of the system SDK, so the test always fails.  However, since they're included as part of the system SDK, they should be guaranteed to always be there unless somebody's build environment is broken.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1397,6 +1397,9 @@ static void terminate_handler()
 		if self.user_settings.opengles:
 			# GLES builds do not use the GL utility functions
 			return
+		if sys.platform == 'darwin':
+			# macOS provides these in a system framework, so this test always fails
+			return
 		self._check_system_library(context, header=['GL/glu.h'], main='''
 	gluPerspective(90.0,1.0,0.1,5000.0);
 	gluBuild2DMipmaps (GL_TEXTURE_2D, 0, 1, 1, 1, GL_UNSIGNED_BYTE, nullptr);


vLKp:
This is a test for header from the OpenGL Utility library.  If these are always provided, why is the test failing to find them?

Kreeblah:
It's because they're provided in a framework, which effectively acts as a library and header namespace.  The compiler and linker don't look at any frameworks that aren't specified via -framework (one per -framework flag, so, for example, `-framework ApplicationServices -framework Cocoa -framework OpenGL` is what's passed to the compiler and the linker for all macOS builds), and will throw an error if it's told to look for one which doesn't exist.  So, in order to get run this test on macOS, it would need to be run with special casing to add `-framework OpenGL` to the compilation and link commands.

MaddTheSane:
> If these are always provided, why is the test failing to find them?

It could be that OpenGL headers aren't in the same places than other architectures: `<OpenGL/glu.h>` as opposed to `<gl/glu.h>`.

Kreeblah:
Oh.  Yeah.  There's that, too.  I forgot about that.

`similar/arch/ogl/ogl.cpp` has an example of that, where it uses some ifdefs for Macs to include `<OpenGL/gl.h>` and `<OpenGL/glu.h>` instead of `<GL/gl.h>` and `<GL/glu.h>`.

vLKp:
I would expect that any frameworks added to the SCons Environment configuration ought to be used during the configure tests, too.  The mismatch in paths seems more likely.  @Kreeblah : could you attach a build log of the failure?  I would prefer to have the test correctly detect environments where the headers are missing, rather than just bypass the test entirely.  I added the test because someone had an environment where the header was missing, and the failure was not reported until compilation reached a file that needed it.

Kreeblah:
[sconf.log](https://github.com/dxx-rebirth/dxx-rebirth/files/5289832/sconf.log)

Sure, here you go.  That's a log from a build attempt with the current commit (1fc8663efffb99d5c7de1422e1be6d5efabb64d8).  As far as I can tell, it's not including any of the -framework flags for this test.  It is for some other tests, but not all of them.

vLKp:
I held off on this for a while because, as above, I would prefer that the test run and be correct.  However, I decided that merging the original tested change as-is, and bypassing the test entirely for OS X users, is better than leaving this broken until I have time to work on it, especially since I cannot test any fix I would write.  Based on the comments here, it seems like OS X is guaranteed to have the required header, so testing for it may not be that useful.  Windows and Linux systems can lack the header, and so a test is useful.

PR #547, opened 2020-10-26
Fixed 'format string is not a string literal' warning, which gets tre…

Kreeblah:
…ated as an error

I'm not sure whether this started in a recent commit, or as the result of the recent new version of clang, but the existing codebase fails on similar/main/powerup.cpp because one of the methods eventually passes an argument through to printf as a format string.  This gets reported as a warning, which then gets elevated to an error, causing compilation to fail.  This just adds an attribute to the method to indicate that that's expected.

### DIFF
--- a/similar/main/powerup.cpp
+++ b/similar/main/powerup.cpp
@@ -140,6 +140,7 @@ static void _powerup_basic_nonhud(int redadd, int greenadd, int blueadd, int sco
 	add_points_to_score(ConsoleObject->ctype.player_info, score);
 }
 
+__attribute__((__format__ (__printf__, 5, 0)))
 void (powerup_basic)(int redadd, int greenadd, int blueadd, int score, const char *format, ...)
 {
 	va_list	args;


vLKp:
Thank you for the report.  However, I think your fix is incorrect.  A third argument of `0` is appropriate when the compiler cannot check the arguments.  In this case, the arguments can be checked, so the third parameter should indicate the position of the `...`.  I think this broke in [Pass control_info & to various functions](/dxx-rebirth/dxx-rebirth/commit/ffb653c0b810bac5e1229f91ecedae515ae329e2), which dropped the prototype for `powerup_basic` on the basis that the function was used in only one file.  However, removing the prototype removed the `__attribute__` marker, which should have been retained.  I pushed a028ea14ae9d3492411871b9f49ed4dca79b1b0f to add the attribute to the definition, including setting the third parameter as it was before the declaration in the header was removed.  This eliminates the clang warning for me.

Kreeblah:
Oh, I didn't realize that the `...` was a valid target for that.  If that can be checked, then, yeah, it definitely makes sense to do so.

Your fix resolves it for me, too, so I'm going to go ahead and close this.

Thank you!

Issue #468, opened 2019-11-08
Alternative to GLU Dependency


shoober420:
### Environment

https://github.com/dxx-rebirth/dxx-rebirth/commit/93d3793b306cebb50be1dc86f974204d5d877ac0

#### Operating System Environment

* [ ] Arch

#### CPU environment

* [ ] x86\_64 (64-bit Intel/AMD)

### Description

The only program that needs "glu" installed I use personally, is dxx-rebirth, even when building the sdl2 port. Why is "glu" a hard depencency? Can an alternative to "glu" be used?

### Steps to Reproduce

Try to compile dxx-rebirth with "glu" uninstalled, and the compile error will occur.

```
similar/arch/ogl/gr.cpp:70:10: fatal error: GL/glu.h: No such file or directory
   70 | #include <GL/glu.h>
      |          ^~~~~~~~~~
compilation terminated.
scons: *** [build/similar/arch/ogl/.d1x-rebirth.gr.o] Error 1
scons: building terminated because of errors.
Failed target count: total=1; targets with enable_build_failure_summary=1: 1
Failed node list:
        build/similar/arch/ogl/.d1x-rebirth.gr.o
Failed command list:
[...]

PR #548, opened 2020-10-28
Fix building and runtime crash on Windows

cbrt64:
Hi, I've been occasionally building (and playing) d1x-rebirth for over a year on Windows 10. (Though I don't pretend to be any good at either.) This is my first attempted contribution, here goes...

This PR addresses two problems:

1. Since GLU detection was added in https://github.com/dxx-rebirth/dxx-rebirth/commit/969caa8c0cb5de8a8af29e7368ca2d21c02e34d7, configuration wouldn't complete on MSYS2/mingw64. Microsoft provides [opengl32/glu32](https://www.opengl.org/archives/resources/faq/technical/mswindows.htm#mswi0130), and from what I can tell each portable project needs to accommodate importing the different library names; see also https://github.com/msys2/MINGW-packages/issues/6693#issuecomment-660605885.

2. When MSYS2's mingw64/gcc recently started building image bases above 32 bits (I think they "reverted" that with msys2/MINGW-packages#6907), d1x-rebirth would crash calling `midi_callback()`. Debugging showed Windows calling the "correct" address, but truncated to 32 bits, which of course worked fine with a lower image base.

Now with these patches, d{1,2}x-rebirth will build and run on x86_64. Sorry, I don't have the i686 toolchain installed anymore to test that platform.

A quick-and-dirty grepping of the codebase didn't find any other suspicious casts. I am by no means fluent in python or C++, and I don't know if any of this will break builds on MSVC or whatever else, so advice and criticism are welcome.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1392,6 +1392,10 @@ static void terminate_handler()
 
 	@_custom_test
 	def check_glu(self,context):
+		glulib = (
+			'GLU32' if self.user_settings.host_platform == 'win32' else
+				'GLU'
+			)
 		if not self.user_settings.opengl:
 			return
 		if self.user_settings.opengles:
@@ -1403,7 +1407,7 @@ static void terminate_handler()
 		self._check_system_library(context, header=['GL/glu.h'], main='''
 	gluPerspective(90.0,1.0,0.1,5000.0);
 	gluBuild2DMipmaps (GL_TEXTURE_2D, 0, 1, 1, 1, GL_UNSIGNED_BYTE, nullptr);
-''', lib='GLU', testflags={'LIBS': ['GLU']})
+''', lib=glulib, testflags={'LIBS': [glulib]})
 
 	@_custom_test
 	def _check_SDL(self,context):
--- a/common/misc/hmp.cpp
+++ b/common/misc/hmp.cpp
@@ -428,7 +428,7 @@ int hmp_play(hmp_file *hmp, int bLoop)
 
 	if ((rc = setup_buffers(hmp)))
 		return rc;
-	if ((midiStreamOpen(&hmp->hmidi, &hmp->devid,1, static_cast<DWORD>(reinterpret_cast<uintptr_t>(&midi_callback)), 0, CALLBACK_FUNCTION)) != MMSYSERR_NOERROR)
+	if ((midiStreamOpen(&hmp->hmidi, &hmp->devid, 1, reinterpret_cast<DWORD_PTR>(&midi_callback), 0, CALLBACK_FUNCTION)) != MMSYSERR_NOERROR)
 	{
 		hmp->hmidi = NULL;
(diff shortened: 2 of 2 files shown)


vLKp:
For 1.: In issue #468, @lemming104 reported that the GLU test was incorrect for Windows, and I pushed 6af0595cfe463e2ce2aa5646a6aa36f41663b80f to address that.  @lemming104 followed up to report that the GLU test then worked.  Is that commit sufficient to fix your environment as well?  With current `master`, I can build for Win32, Win64, and Linux/amd64 successfully.

For 2.: yes, the call looks wrong in `master`.  The prototype for `midiStreamOpen` is no help here.  It should have taken the callback as a proper function pointer, not as a cast of the function pointer to an integer.  If it had done so, then no cast would have been needed, and thus the cast could not be written incorrectly.  Although I think this will not cause any problems with the fix, I will also note that the prototype is inconsistent across headers:
```
mingw32:usr/include/mmsystem.h:MMRESULT WINAPI midiStreamOpen(LPHMIDISTRM,LPUINT,DWORD,DWORD,DWORD,DWORD);
i686-w64-mingw32:usr/include/mmsystem.h:  WINMMAPI MMRESULT WINAPI midiStreamOpen(LPHMIDISTRM phms,LPUINT puDeviceID,DWORD cMidi,DWORD_PTR dwCallback,DWORD_PTR dwInstance,DWORD fdwOpen);
x86_64-w64-mingw32:usr/include/mmsystem.h:  WINMMAPI MMRESULT WINAPI midiStreamOpen(LPHMIDISTRM phms,LPUINT puDeviceID,DWORD cMidi,DWORD_PTR dwCallback,DWORD_PTR dwInstance,DWORD fdwOpen);
```
Old mingw32 uses DWORD, but i686-w64 and x86_64-w64 use DWORD_PTR.

I believe your code change is correct for `midiStreamOpen`.  Due to other changes, likely the fix for GLU, your pull request does not merge cleanly.  I cherry-picked out just the cast fix.  I can still build Win32, Win64, and Linux/amd64 with that change in my tree.

cbrt64:
> I pushed 6af0595 ... Is that commit sufficient to fix your environment as well?

Builds just fine here, thanks. I had looked at #468 the day before, but just missed @lemming104 beating me to it. Anyway, I'm sure your fix is better. 😉

> The prototype for `midiStreamOpen` is no help here. It should have taken the callback as a proper function pointer, not as a cast of the function pointer to an integer.

Windows 3.1 SDK's mmsystem.h is full of `DWORD dwCallback`; my guess is MS years ago created these "pointer-ish" integer types to preserve backward compatibility.

> I cherry-picked out just the cast fix.

Thanks, LGTM.

vLKp:
Based on your response, I think there is no need to resolve the merge conflict and then merge this request, as equivalent work is already in `master`.  Please reopen if you disagree.

PR #560, opened 2020-12-19
Instructions for building with MSYS2

cbrt64:
First commit: it was always `BIN_DIR`, not `BINDIR`, as far as I can tell. Just something I noticed a while back while learning to build this project natively.

Second commit: This is the simplest way I've found to build dxx-rebirth on Windows. I welcome nitpicks if you think the layout doesn't match the rest of the document or whatever.

### DIFF
--- a/INSTALL.markdown
+++ b/INSTALL.markdown
@@ -69,6 +69,9 @@ previous Rebirth release and installing current PhysFS headers.
 However, building from source is recommended to ensure a consistent
 environment.
 
+#### MSYS2/mingw-w64 (Windows alternate method)
+* `pacman -S git ${MINGW_PACKAGE_PREFIX}-{gcc,pkgconf,scons,SDL,SDL_image,SDL_mixer,libpng,physfs}`
+
 #### Linux
 Install the listed prerequisites through your system package manager.
 * Arch PKGBUILD files are in `contrib/arch/`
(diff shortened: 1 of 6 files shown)


vLKp:
> First commit: it was always BIN_DIR, not BINDIR, as far as I can tell. Just something I noticed a while back while learning to build this project natively.

This is not a typo.  BINDIR is a placeholder in the text, for the expansion that happens later in the paragraph.  It has no connection to `UserSettings.BIN_DIR`, which is an implementation detail used for managing that path.  They do not need to agree because the user cannot directly affect `UserSettings.BIN_DIR`.  It can only be changed indirectly, by changing the prefix.

As for the instructions, I cannot tell if they are correct, but they look reasonable.  The last paragraph, discussing how to run from the Windows command-line, is a bit confusing.  I think you want the user to prepend the mingw bin path to $PATH, but then what?  Just run the executable normally as a statement on a subsequent line?  A fully worked example might be helpful here.

cbrt64:
Sorry for the delay.

My well-meaning BINDIR patch was over a year ago. I recall my confusion then about what scons even was, and should have given this another look before pushing. Dropped.

While MSYS2 has a really shallow learning curve, especially for someone already used to GNU/Linux tools, I went ahead and added a little more explanatory background to the build instructions. Also minor reformatting.

I've tried to clarify the .exe running instructions enough to be "initiate-compatible", without going into unnecessary detail. Is `contrib/` a good place for example scripts?

vLKp:
> Is `contrib/` a good place for example scripts?

That is fine, yes.

dimag0g:
Nice guide. A small note: you don't need to manually install `libpng` and `SDL` as they are both dependencies of `SDL_image`.

And I don't think the bat files to run executables are necessary: rather, you should find out which DLLs should be distributed along with the exetutables. Users shouldn't have to install Msys2 just to play a game, and developers can always rung the game from inside Msys.

vLKp:
@cbrt64, are you planning to do more work on this, or should I take over and try to write up the examples?

cbrt64:
I had already pushed an update a month ago, don't know if you saw that already. Today I've clarified it a bit, rebased and force-pushed again.

My sole intent with this was to simplify the build-it-yourself process for Windows. I'm willing to work on this more if you think it needs it, but at this point I would need input on which direction to take it. Or feel free to redo it or add to it as you see fit.

cbrt64:
@dimag0g:

I thought about the redundant dependencies, but figured it best to explicitly list dxx-rebirth's direct dependencies for the sake of completeness/consistency. Anyway they were already listed like that, [here](https://github.com/cbrt64/dxx-rebirth/blob/t/INSTALL.markdown#homebrew) for example.

> find out which DLLs should be distributed along with the exetutables.

I don't know a way to automate or simplify that process. I have a different idea, and plan on starting another issue to address this. But I'm slow, don't hold your breath. ;)

> developers can always run the game from inside Msys.

As I said above, This PR is aimed mainly at developers (DIY or otherwise).

> Users shouldn't have to install Msys2 just to play a game

I agree completely.

P.S. Thanks for [confirming](https://github.com/dxx-rebirth/dxx-rebirth/issues/564#issuecomment-761807946) 32-bit works, I haven't had the i686 toolchain installed for a long while and was just assuming it would work.

vLKp:
> I had already pushed an update a month ago, don't know if you saw that already.

I saw the Github note that there was a push, but I had not read the commits, and that push preceded our discussion of example scripts, so I was unsure whether you were working on the example scripts to be done in this MR, or if that will be done later (or not at all, if you decide you don't want to work on that).  Providing example scripts is not required for this MR.  I'll try to get this re-reviewed and merged this weekend.

PR #568, opened 2021-02-05
Fix legacy screenshot errors

dimag0g:
Currently (commit 5bc927b) the build fails with `scons screenshot=legacy`. There are two compilation errors:

- missing cast from `pcx_result` to `unsigned` in `similar/main/game.cpp`
- visibility problems for `pcx_encode_line` and `pcx_encode_byte` in `similar/2d/pcx.cpp`

The first problem is certainly a missing cast. As for the second one, I just removed the anonymous namespace wrapped around the problematic functions. Supposedly, a better fix is possible which preserves the namespace. Feel free to tell me how you prefer this to be fixed, or drop this PR and implement both fixes yourself.

### DIFF
--- a/similar/2d/pcx.cpp
+++ b/similar/2d/pcx.cpp
@@ -48,8 +48,6 @@ COPYRIGHT 1993-1998 PARALLAX SOFTWARE CORPORATION.  ALL RIGHTS RESERVED.
 
 namespace dcx {
 
-namespace {
-
 #if DXX_USE_SDLIMAGE
 struct RAII_SDL_Surface
 {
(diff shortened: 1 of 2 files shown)


vLKp:
Confirmed, but only for `opengl=0`.  This is not a common configuration, though it is a supported one.  I elected to fix this a bit differently.  Rather than add a cast, I switched the type back to `unsigned` since it is only used as a success/failure flag.  For the namespace, I moved the affected definitions into an anonymous namespace instead of moving the declarations out of one.

I saw both of the problems you reported before my commit, and neither now.  Please reopen if the commit does not work for you.

PR #569, opened 2021-02-06
optimize include files (include what you use)

dimag0g:
I have spent some time optimizing include files, mostly, removing include lines which are not necessary. I ended up removing 400+ such include lines, which should have a positive effect on complication time, and make dependencies between files more obvious.

This will need some testing, especially for rarely used compilation options and features. Unfortunately, the builds are not binary identical even when they could be, as some object files depend on date, time and git status, so there's no way to prove there were no functional changes. I have tested several configurations on Linux and Windows, with various options: `debug=1`, `editor=1`, `opengl=0`, `sdl2=1` and `screenshot=legacy`, and they look OK. Still, it's possible there are subtle differences I have missed, and there probably are other configuration options I didn't test. Luckily, missing includes usually result in rather prompt build failures and are easy to fix.

Let me know what you think of this change, and sorry for not discussing it upfront.


vLKp:
> Let me know what you think of this change

Overall, I like the idea.  I've wanted to do this for a while, but never found the time for it.  Looking through the changes, I see a few places where the change was not a simple drop of an include file (such as where a declaration was added to a non-fwd header).  I also see a couple, such as `d_array.h` / `fwd-d_array.h` where the old design was deliberate.  The forward declaration was included so that a change which caused a mismatch between the forward declaration and the definition would be reliably flagged in any file that included the header that provided the definition.

I'll need to look through it in more detail.  I'll almost certainly pull in the removals of unnecessary `#include` directives.  I have not decided how to handle some of the others.  Regardless, thank you for putting in the effort to compute what is unneeded, and for testing the various configurations.

> sorry for not discussing it upfront.

That is fine.  I can't commit to taking unannounced projects like this, but I'm happy to see the initiative.

dimag0g:
@vLKp, I think I understand the idea of forward declarations consistency enforcement. Let me know if you want me to revert such changes.

vLKp:
> Let me know if you want me to revert changes which violate it.

No need.  I added a minor commit to do that when I pushed the series.

Thank you for finding these unnecessary includes, and the associated testing.  Sorry for the delay in getting this merged.

Issue #564, opened 2021-01-16
Runtime bug: New game dialog text is shifted to the side if open with mouse


dimag0g:
### Environment
Commit 7204b91e1aad6d359661f1084cad8e9971f28408
`scons -j4`

#### Operating System Environment
Raspbian Buster, Pi 4B
Windows 10, Msys2 + MinGW-w64

#### CPU environment

* [X] x86 (32-bit Intel/AMD)
* [ ] x86\_64 (64-bit Intel/AMD)
* [X] ARM (32-bit)
* [ ] ARM64 (64-bit; sometimes called AArch64)

### Description
The text in the New Game dialog is rendered outside of dialog borders.

#### Regression status
Not present in commit 03512841b19315532a95cadc1d6351e24a70136c. Seems to be related to newmenu changes.

### Steps to Reproduce
Start the game and click on "New Game" with the mouse pointer.

![2021-01-16-224949_1280x720_scrot](https://user-images.githubusercontent.com/10945319/104823975-82cb7f80-584e-11eb-82e5-0f9658fc7364.png)

* Engines affected:
  * [X] D1X-Rebirth
  * [X] D2X-Rebirth

* Game modes affected:
  * [X] Single player
  * [ ] Multiplayer competitive
    * [ ] Anarchy
    * [ ] Team Anarchy
    * [ ] Robo-Anarchy
    * [ ] Capture the Flag
    * [ ] Bounty
  * [ ] Multiplayer cooperative

AlumiuN:
I don't have time to test further at the moment, but this doesn't happen for me at 2560x1440 in fullscreen. Does it happen for you at higher resolutions?

dimag0g:
@AlumiuN, Yes, it happens in fullscreen too (at 1280x720). The important step is to click on the menu with the mouse, selecting "New Game" with the keyboard works fine. Could be a Linux/SDL1 exclusive, which OS/build options do you have?

AlumiuN:
I'm running on Windows 10 20H2, using the SDL1 build (I'm not specifying any other options). I have tested it again and it functions correctly for me at lower resolutions.

dimag0g:
[...]

PR #581, opened 2021-03-05
Fix alignment of screen resolution menu text

dimag0g:
The screen resolution menu is affected by the same issue as the new game menu: https://github.com/dxx-rebirth/dxx-rebirth/issues/564

I applied the same fix as before: point to the top level canvas in the constructor. I hope my cargo cult programming skills are up to the task :)


vLKp:
Merged.  Long term, this suite of changes (using screen instead of canvas) probably ought to be reverted, but until the drawing code uses current canvas consistently and correctly, this set of changes is appropriate to keep things working.

Issue #561, opened 2020-12-25
Mac OS X build notes


raptor:
This isn't a bug/issue, rather a success:  

I have built the latest version of d2x (revision 0d70c3831c31700ca1ad6edb73f6d3bce1d58a84) on macOS 10.13 using the following steps:
```
1. Get the dxx-rebirth source code and put it somewhere:

   git clone https://github.com/dxx-rebirth/dxx-rebirth.git

2. Install brew:

   /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"

3. Install dependencies:

   brew install scons sdl sdl_image sdl_mixer physfs

4. Make earlier macOS compatible (if wanted):

   export MACOSX_DEPLOYMENT_TARGET=10.11

5. Run scons in the dxx-rebirth dir:

   scons opengl=1 sdlmixer=1 d2x=1 macos_add_frameworks=0

This builds the D2X-Rebirth.app (Xcode command line tools required:  xcode-select --install)

6. Fix up the executable and libraries in the .app to make it self-contained by getting the following tool:

   git clone https://github.com/auriamg/macdylibbundler.git
   cd macdylibbundler
   make

7. Then run the command against the .app:

   cd /path/to/dxx-rebirth
   /path/to/dylibbundler -od -b -x D2X-Rebirth.app/Contents/MacOS/d2x-rebirth -d D2X-Rebirth.app/Contents/Frameworks

The .app now has the .dylib files contained therein to run on any macOS system.

8. Copy descent 2 resources into:  

   D2X-Rebirth.app/Contents/Resources/

9. PLAY
```

Attached is the .app created using the above steps, but without the game data files (step 8).  

Steps 6 and 7 were key to getting friends to run it without requiring extra dependencies. Should be compatible with macOS 10.11 and later.

[D2X-Rebirth.app.zip](https://github.com/dxx-rebirth/dxx-rebirth/files/5741155/D2X-Rebirth.app.zip)

Thanks for your efforts. This game is great!

vLKp:
Well done.  Would you be interested in writing this up and submitting a pull request to improve the documentation, so that other users can readily find how to do this?  If not, I can write something from the steps you posted here, but I won't be able to test it.

raptor:
I think I can do that. The INSTALL.markdown file can be slimmed down and simplified. Would it be better to split out the macOS stuff into it's own file, like INSTALL.macos.md?

[...]

PR #583, opened 2021-03-26
Embed dylibs in app bundles

Kreeblah:
I'm having trouble figuring out how to test for the existence of a binary named `dylibbundler` on the system path, but these changes implement steps 6 and 7 from https://github.com/dxx-rebirth/dxx-rebirth/issues/561 and add a note about the additional dependency in INSTALL.markdown.

### DIFF
--- a/INSTALL.markdown
+++ b/INSTALL.markdown
@@ -136,7 +136,8 @@ DXX-Rebirth can be built from the Terminal (via SCons) without Xcode; to build u
  sdl\_mixer
  physfs
  libpng
- pkg-config**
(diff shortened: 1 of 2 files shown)


vLKp:
One option, which would be expedient if not exactly elegant, would be to add a configure step that runs `dylibbundler --version` or similar and prints its output.  If that step fails, then assume `dylibbundler` is not properly installed, display an error message, and stop.  If it succeeds, then the version string is now part of the SCons log output, and may be useful if anyone reports problems with it.

Issue #588, opened 2021-04-18
Trouble shooting through grates


CHILLYBUS:
#### Operating System Environment
* Microsoft Windows
  * [ ] Microsoft Windows XP
  * [ ] Microsoft Windows Vista
  * [ ] Microsoft Windows 7
  * [ ] Microsoft Windows 8
  * [ ] Microsoft Windows 8.1
  * [X] Microsoft Windows 10

#### CPU environment
* [ ] x86 (32-bit Intel/AMD)
* [X] x86\_64 (64-bit Intel/AMD)
* [ ] ARM (32-bit)
* [ ] ARM64 (64-bit; sometimes called AArch64)

#### Game environment
Any location with transparent grate assets.

### Description
Grates typically can be shot through in older builds of Descent II - the original hitboxes aren't particularly exacting, but are approximate enough that any weapons fire can be reasonably sent through them provided the player's aim is good enough. That isn't the case in recent versions of D2X-Rebirth. I've done some cursory tests myself with recent builds after seeing some players have trouble during some Twitch streams of the game, and encounter roughly the same issues.

Some streamers report an identical issue in D1X, and other players report this issue not occurring for them at all.

#### Regression status
This issue was not present in early builds of Rebirth 60/61, but is present in late-2020 and early-2021 builds of the game, if that's helpful. 

v0.61 0.60.0-beta2-709-GF0A47C358BA6* and earlier appear to be working. Still unsure which exact build this issue starts with.

### Steps to Reproduce
Fly to a grate, try shooting through the grate. Best demo of the problem is shooting through wide grates - the vulcan window into the red key room on level 1 of D2: Counterstrike is the best example, but switches behind tight grates in that level are good tests of the issue too since those need to be shoot-through-able too.

* Engines affected:
  * [ ] D1X-Rebirth
  * [X] D2X-Rebirth

* Game modes affected:
  * [X] Single player
  * [ ] Multiplayer competitive
    * [ ] Anarchy
    * [ ] Team Anarchy
    * [ ] Robo-Anarchy
    * [ ] Capture the Flag
    * [ ] Bounty
  * [?] Multiplayer cooperative

Haven't tested cooperative play to see if this is present.

vLKp:
> Some streamers report an identical issue in D1X, and other players report this issue not occurring for them at all.

[...]

PR #589, opened 2021-04-24
Fix for shooting through gratings bug (#588)

AlumiuN:
I'm not sure if this is the best fix for this in the long run (as figuring out why the code being in the closure triggered the issue is probably a more complete fix), but it appears to fix it in the short term.


vLKp:
Thank you for the time analyzing and debugging this.  Based on your proposed change, I identified why the closure was causing a problem, and applied a more targeted fix instead.  A much longer explanation is in issue #588.  I am closing this pull request without merging as the change I made seems to achieve the same result.  I'm not sure how much time you saved me by proposing this, but probably at least as much time as it took you to bisect down to the bad revision.

PR #591, opened 2021-05-06
Added test for dylibbundler on macOS

Kreeblah:
Since macOS builds now fail if dylibbundler isn't installed, this adds a simple test to verify that it exists and is executable.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -2901,7 +2901,7 @@ BOOST_AUTO_TEST_CASE(f)
 		dylibbundler_output = os.popen('dylibbundler -h').read()
 		if 'dylibbundler is a utility' not in dylibbundler_output:
 			context.Result('no')
-			raise SCons.Errors.StopError('dylibbundler is required for compilation on macOS when not using frameworks')
+			raise SCons.Errors.StopError('dylibbundler is required for compilation for a macOS target when not using frameworks')
 		context.Result('yes')
 		return
 
--- a/SConstruct
+++ b/SConstruct
@@ -2889,9 +2889,9 @@ BOOST_AUTO_TEST_CASE(f)
 			self.check_boost_test(context)
 
 	# dylibbundler is used to embed libraries into macOS app bundles,
-	# however it's only available on macOS.  While required for macOS
-	# builds, builds should not fail on other operating systems if
-	# it's absent.
+	# however it's only meaningful for macOS builds, and, for those,
+	# only required when frameworks are not used for the build.  Builds
+	# should not fail on other operating system targets if it's absent.
 	@_custom_test
 	def _check_dylibbundler(self,context):
 		context.Display('%s: checking whether dylibbundler is installed...' % self.msgprefix)
--- a/SConstruct
+++ b/SConstruct
@@ -2901,7 +2901,7 @@ BOOST_AUTO_TEST_CASE(f)
 		dylibbundler_output = os.popen('dylibbundler -h').read()
 		if 'dylibbundler is a utility' not in dylibbundler_output and not self.user_settings.macos_add_frameworks:
 			context.Result('no')
-			raise SCons.Errors.StopError('dylibbundler is required for compilation on macOS')
+			raise SCons.Errors.StopError('dylibbundler is required for compilation on macOS when not using frameworks')
 		context.Result('yes')
 		return
 
--- a/SConstruct
+++ b/SConstruct
@@ -2888,6 +2888,23 @@ BOOST_AUTO_TEST_CASE(f)
 		if register_runtime_test_link_targets:
 			self.check_boost_test(context)
 
+	# dylibbundler is used to embed libraries into macOS app bundles,
+	# however it's only available on macOS.  While required for macOS
+	# builds, builds should not fail on other operating systems if
+	# it's absent.
+	@_custom_test
+	def _check_dylibbundler(self,context):
+		context.Display('%s: checking whether dylibbundler is installed...' % self.msgprefix)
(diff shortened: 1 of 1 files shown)


SConstruct
```diff
@@ -2888,6 +2888,23 @@ BOOST_AUTO_TEST_CASE(f)
...
 			self.check_boost_test(context)
 
+	# dylibbundler is used to embed libraries into macOS app bundles,
+	# however it's only available on macOS.  While required for macOS
+	# builds, builds should not fail on other operating systems if
+	# it's absent.
+	@_custom_test
+	def _check_dylibbundler(self,context):
+		context.Display('%s: checking whether dylibbundler is installed...' % self.msgprefix)
+		if sys.platform != 'darwin':
```
vLKp, line 2898: I think this test is incorrect.  A MacOS user can cross-compile for a non-MacOS target, and `sys.platform` would be `darwin` in that case because the Python interpreter is running on MacOS, but the desired code flow would be to skip the dylibbundler check, because dylibbundler is not required when cross-compiling for a non-MacOS target.

I suggest instead inspecting the target host_platform.

SConstruct
```diff
@@ -2888,6 +2888,23 @@ BOOST_AUTO_TEST_CASE(f)
...
+			context.Result('n/a')
+			return True
+		dylibbundler_output = os.popen('dylibbundler -h').read()
+		if 'dylibbundler is a utility' not in dylibbundler_output:
+			context.Result('no')
+			raise SCons.Errors.StopError('dylibbundler is required for compilation on macOS\n')
```
vLKp, line 2904: I don't think this is quite accurate.  As I read the code flow, dylibbundler is only required `if not self.user_settings.macos_add_frameworks`.  Users who leave frameworks enabled will not attempt to run `dylibbundler`.

Also, as a minor point, no trailing newline is needed for the string passed to `StopError`.

SConstruct
vLKp: You may find the design _guarded_test_windows decorator useful here.  Guarded tests are not even registered on platforms for which the guard is false, so there is no need for them to display a message.

SConstruct
```diff
@@ -2888,6 +2888,23 @@ BOOST_AUTO_TEST_CASE(f)
...
+	# dylibbundler is used to embed libraries into macOS app bundles,
+	# however it's only available on macOS.  While required for macOS
+	# builds, builds should not fail on other operating systems if
+	# it's absent.
+	@_custom_test
+	def _check_dylibbundler(self,context):
+		context.Display('%s: checking whether dylibbundler is installed...' % self.msgprefix)
+		if sys.platform != 'darwin':
+			context.Result('n/a')
+			return True
```
vLKp, line 2900: Tests are not required to return any particular value, and many use a bare `return`.  Unacceptable results are reported via exception, rather than return code.

vLKp (changes requested):

SConstruct
Kreeblah: That makes sense, sure.

For the guarded tests, are those evaluated based on the target platform that's being compiled for, or the platform the compiler is running on?  In other words, if somebody is cross-compiling on a Mac for a non-macOS target, would a test that uses _guarded_test_darwin be skipped?

SConstruct
Kreeblah: That's a good point about the framework situation.  I'll fix that.

SConstruct
Kreeblah: Fair enough.  I'll change these to just return statements, then.

SConstruct
vLKp: That would depend on how you wrote the guard.  The existing `_guarded_test_windows` decorator inspects the `host_platform` which, following autotools convention, is the platform on which the built program is meant to run, rather than the platform on which SCons and the compiler are running.  Assuming you followed that pattern, then yes, a MacOS user cross-compiling to a non-MacOS target would skip the guarded test.

Definition of `_guarded_test_windows`:

https://github.com/dxx-rebirth/dxx-rebirth/blob/48589d0fb019eb971818c15665358e30fa778dc4/SConstruct#L416

Logic that uses the guards to exclude tests from execution:

https://github.com/dxx-rebirth/dxx-rebirth/blob/48589d0fb019eb971818c15665358e30fa778dc4/SConstruct#L606

Kreeblah:
Alright, I think this is ready for another review.

vLKp:
Merged.  On playing with it, I decided to try to fix an issue that using `os.popen` allowed the "not found" text from the shell to write directly to stderr, bypassing SCons.  That reminded me of the use, and value, of StaticSubprocess for caching external command runs.  I got a bit carried away, but I think what I pushed is faithful to what you wanted (and what I pushed still records your commits in the history).  Please sync to master and report whether it works as you want.

Kreeblah:
Yep, it fails when macos_add_frameworks is false and dylibbundler is not installed, but passes when macos_add_frameworks is true and/or dylibbundler is installed.  That's what I was hoping for.

Thanks for your help with this!

Kreeblah:
Oh.  Actually, there is one thing.  I just re-read the error this returns if it fails.  It suggests that to fix it, the user can set macos_add_frameworks=False or to install dylibbundler.  That should probably be suggesting to set macos_add_frameworks=True, I'd think?

vLKp:
Good catch.  When I was reworking this area, I temporarily inverted the darwin guard and sometimes the add_frameworks guard, so that I could explore all the paths on Linux.  I suspect I wrote that advice when I had the guards inverted.

Fixed.

Issue #594, opened 2021-06-12
Runtime bug: Mission menu severely broken when using subdirectories


AlumiuN:
Using Win10 64-bit, 32b7899 (traced the break to commit 7f2df64649796bbbd74edc400baab2edde1287ad)

The mission menu subdirectory handling seems to be severely broken - only the same number of items that are present in the root directory are displayed in any subdirectory (with the notable exception of a folder that the mission selection opens to), and if the entry in the same position in the list as the selected mission is a directory in the root, that directory is opened instead of it selecting the mission in question. My theory is that the subdirectory window is being passed the correct strings for mission names and window title, but incorrect behaviour for the entries.

vLKp:
Agreed.  It looks like the breaking commit had a copy-paste error that passed the wrong variable.  I pushed your proposed fix as-is.

PR #595, opened 2021-06-13
Fixed mission menu subdirectory handling

AlumiuN:
A fix for issue #594, which turned out to be surprisingly simple.


vLKp:
Thank you for the fix.  Looking at the commit you identified as the problem, it looks like this was a copy-paste error on my part when doing the transformation.

Merged.

PR #596, opened 2021-06-14
Fix door lock/unlock triggers causing a crash when given an invalid wall location

AlumiuN:
Currently, a door locking or unlocking trigger will cause a crash if given a location where there isn't a wall. This is not inline with original behaviour, as it seemingly ignores these (or at least doesn't crash consistently). This should (hopefully; it certainly seems to in my testing) make Rebirth simply ignore such invalid triggers instead.

### DIFF
--- a/similar/main/switch.cpp
+++ b/similar/main/switch.cpp
@@ -132,6 +132,8 @@ static void do_unlock_doors(fvcsegptr &vcsegptr, fvmwallptr &vmwallptr, const tr
 {
 	const auto op = [&vmwallptr](const shared_segment &segp, const unsigned sidenum) {
 		const auto wall_num = segp.sides[sidenum].wall_num;
+		if (wall_num == wall_none)
+			return;
 		auto &w = *vmwallptr(wall_num);
 		w.flags &= ~WALL_DOOR_LOCKED;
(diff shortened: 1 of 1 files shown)


vLKp:
The original game had very little checking on array indexing operations.  Using an invalid index would most likely access the memory near the wall array, potentially corrupting it.  In some cases, this corruption would be harmless, if the overwritten field was unused or was subsequently reset before being used.  I converted many array accesses to checked forms, which is why invalid accesses now raise an exception (which usually kills the program).  With the right build options, that exception can be suppressed, which would revert to the undefined behavior of accessing outside the bounds of the array.

Testing for `wall_none` is a fairly common pattern in the code, though on review I am doubtful that this is sufficient.  The `wall_num` field allows 16-bit wall identifiers, and the savegame loading path seems to allow loading a 16-bit wall value.

https://github.com/dxx-rebirth/dxx-rebirth/blob/a17129a9aafd55c046ece2ebadb67b619710d2cb/similar/main/state.cpp#L2038-L2039

Since that test pattern is common, I merged your change as-is.  Your change prevents a real crash, and if that test is insufficient, there are many places to review and modify.

PR #604, opened 2021-09-22
Update fireball.cpp

MaddTheSane:
Add include optional.
Fixes build failure on macOS.


vLKp:
Merged.  Due to an implementation detail, gcc on Linux worked because a gcc header had `#include <optional>`, by way of `<bits/node_handle.h>`, by way of `<bits/hashtable.h>`, by way of `<unordered_map>`, by way of `<functional>`, by way of `<pstl/glue_algorithm_defs.h>`, by way of `<algorithm>`.  `fireball.cpp` does include `<algorithm>`.

PR #610, opened 2021-11-05
Fix compilation on ARM Macs and make dylibbundler an optional step

Kreeblah:
I just got an ARM Mac, and found a couple of issues that don't exist on the Intel Mac builds.

First, it failed to compile due to the dylibbundler step.  Because the dylibbundler test used a different method for calling dylibbundler, it found it in the test but not in the actual call to use it to bundle libraries into the app bundle.  On ARM Macs, Homebrew is installed into a different location which requires adding a directory to the path in .zshrc (or equivalent file), which doesn't seem to be called in the previous method of using dylibbundler.  So, I updated it to use the same subprocess call that the test does, and inherit the same path settings.  This fixes the issue where the build process can't find dylibbundler during the build step, even though it found it during the test.

The other issue I found is that dylibbundler seems to give me binaries which just don't run on my ARM Mac.  The project is not maintained very well these days (per https://github.com/auriamg/macdylibbundler/issues/41#issuecomment-523246398), so while I will mention it upstream, it seemed that making this an optional step makes the most sense in the meantime.  To do that, I added `macos_bundle_libs` which defaults to `False`.  It will only use dylibbundler if `macos_use_frameworks` is `False` and `macos_bundle_libs` is `True`.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -5288,7 +5288,7 @@ class DXXProgram(DXXCommon):
 					icon_file=os.path.join(cocoa, '%s-rebirth.icns' % dxxstr),
 					resources=[[os.path.join(self.srcdir, s), s] for s in ['English.lproj/InfoPlist.strings']])
 			if self.user_settings.macos_bundle_libs and not self.user_settings.macos_add_frameworks:
-				print('Bundling libraries for  %s' % self.PROGRAM_NAME)
+				print('Bundling libraries for %s' % self.PROGRAM_NAME)
 				dylibbundler_process = StaticSubprocess.pcall(('dylibbundler', '-od', '-b', '-x', '%s.app/Contents/MacOS/%s-rebirth' % (self.PROGRAM_NAME, dxxstr), '-d', '%s.app/Contents/libs' % self.PROGRAM_NAME))
 
 class D1XProgram(DXXProgram):
--- a/SConstruct
+++ b/SConstruct
@@ -5282,14 +5282,32 @@ class DXXProgram(DXXCommon):
 			import tool_bundle
 			sys.path = syspath
 			tool_bundle.TOOL_BUNDLE(env)
-			env.MakeBundle(os.path.join(self.user_settings.builddir, '%s.app' % self.PROGRAM_NAME), exe_node,
+			# bundledir is an SCons.Environment.Dir corresponding to the
+			# first argument passed to MakeBundle, after any adjustments
+			# made inside MakeBundle.
+			#
+			# appfile is an SCons.Environment.File corresponding to the
+			# game executable copied into the `.app` directory.
+			bundledir_list = env.MakeBundle(os.path.join(self.user_settings.builddir, '%s.app' % self.PROGRAM_NAME), exe_node,
 					'free.%s-rebirth' % dxxstr, os.path.join(cocoa, 'Info.plist'),
 					typecode='APPL', creator='DCNT',
 					icon_file=os.path.join(cocoa, '%s-rebirth.icns' % dxxstr),
 					resources=[[os.path.join(self.srcdir, s), s] for s in ['English.lproj/InfoPlist.strings']])
+			bundledir = bundledir_list[0]
+			appfile = SCons.Node.FS.default_fs.File(str(bundledir.Dir('Contents/MacOS')) + '/%s-rebirth' % dxxstr)
 			if self.user_settings.macos_bundle_libs and not self.user_settings.macos_add_frameworks:
-				print('Bundling libraries for %s' % self.PROGRAM_NAME)
-				dylibbundler_process = StaticSubprocess.pcall(('dylibbundler', '-od', '-b', '-x', '%s.app/Contents/MacOS/%s-rebirth' % (self.PROGRAM_NAME, dxxstr), '-d', '%s.app/Contents/libs' % self.PROGRAM_NAME))
+				message(self, 'Bundling libraries for %s' % (str(bundledir),))
+				# If the user has set $PATH, use it in preference to the
+				# built-in SCons path.
+				dylibenv = env['ENV']
+				user_environment_path = os.environ.get('PATH')
+				if user_environment_path:
+					dylibenv = dylibenv.copy()
+					dylibenv['PATH'] = user_environment_path
+				env.Command(target=bundledir.Dir('Contents/libs'),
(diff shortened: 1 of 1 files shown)


vLKp:
Making dylibbundler optional is fine.  For a user who disables both frameworks and dylibbundler, how does the game work?  Does it find the libraries from the system, but the `.app` isn't portable across systems as it would be if using frameworks or dylibbundler?

I think the subprocess change is incorrect though.  `Command` defines a recipe for how SCons can build a target if the target needs to be built.  It may not result in any command execution.  For example, when running `scons ; scons`, the second run should find that the first run updated everything, and that nothing needs to be done, so the second will not run any `Command` recipes.  `StaticSubprocess` is designed to immediately run the command, if it has not been run before by this run of `scons`.  For this use case, that is wrong in two ways.  First, it will always re-run dylibbundler, even if that is unnecessary.  Second, it will _immediately_ run dylibbundler, before `scons` has the opportunity to run other recipes.  Therefore, I think in a clean checkout, `scons` will run dylibbundler, then compile the source files, then link them into an executable, then make the bundle.  I'm not sure if dylibbundler will fail gracefully or fail badly in that case, but I expect that running it before the bundle is created will, at the least, make dylibbundler not do anything useful.  Looking at the code currently in `master`, I think that the `Command` is actually wrong as-is, even ignoring your path search problem.  The bundle code above it respects `builddir`, but it looks like the `Command` always assumes `builddir` is `.`, so users who set `builddir` to a subdirectory will never pass the right target paths to the dylibbundler, even when it is found and executed.

As I understand your description, the problem is that `Command` does not search your custom directories, and so does not find `dylibbundler`, whereas `StaticSubprocess` did search them.  How did you define these extra directories?  (Preferably, show the line in your dotfile and which dotfile it is.)  I think the right fix would be to change the execution environment that is used for Command, so that it includes your custom path.  SCons has a documented behavior (`7.3. Controlling the Execution Environment for Issued Commands`) of using a trimmed down environment for executed commands.  Since StaticSubprocess is built on Python's `subprocess` module, if SCons doesn't modify `os.environ`, then `subprocess` can see your full $PATH, but environment-associated commands see the restricted version that SCons defines.  While I don't have an ARM Mac, I can test the principles and publish a branch for you to try, if you would prefer.

Kreeblah:
If the libraries aren't bundled into the app, then, yes, it will use the installed libraries, but the app bundle will not be portable to another system without them having the same library install.

That's an accurate description of the problem, yes.  As for how the custom directories are defined, that's done in my dotfiles, yes.  `config.fish`, `.zshrc`, and `.bashrc` all have a line in them which runs `eval` on the following:

`/opt/homebrew/bin/brew shellenv`

That then runs [this script](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cmd/shellenv.sh), which sets a number of environment variables, including adding the Homebrew command install location to my path.  This is the officially recommended way to do that for Homebrew, so it should be the same on other ARM Macs using Homebrew-installed dependencies.

If you would be up to creating a branch to test the principles you described, that would be fantastic.  I'm sure I could use it to create a better solution for this issue.  Thank you so much!

vLKp:
Based on your explanation, it's just the usual $PATH variable, albeit with a somewhat more indirect computation.  That's fine.  I needed to confirm that $PATH was the right way to find this, rather than a shell function or some other special wrapper.

I pushed https://github.com/dxx-rebirth/dxx-rebirth/tree/experimental/pull-610-dylibbundler-env (ref name `experimental/pull-610-dylibbundler-env`) with what I think you need.  I included some general updates to OSX code that I cannot normally test.  In this case, I bypassed the platform==darwin checks to force that code to run, so I can say at least that it's valid Python syntax and runs to completion.  I cannot say whether it puts files in the right place, and with the right contents, to satisfy requirements on OSX.  During my testing, `env.Command` seemed to search my $PATH even without the code I thought you needed, so I am unsure why `master` did not work for you as-is.  The key bit of new code is that if `os.environ` has `PATH` set, then a private copy of the construction environment's ENV (the process environment) is `copy()`ied into existence, and patched with the `PATH` from the `scons` environment.  Then, the `Command()` is instructed to use the (possibly copied+modified) process environment.  If the experimental branch doesn't work for you as-is, please feel free to ask questions or patch it, at your preference.

Kreeblah:
I updated the `SConstruct` file based on your changes, though I did need to make a few of my own.  It seems that `MakeBundle()` returns one value, which a list of `Dir` objects, though it seems to only have one actual member.  So, I had to get that object from the list and use it to make a `File` object referring to the executable in the bundle.

With those changes, though, `dylibbundler` runs on both my Intel Macs as well as my ARM Mac (though I still have the `dylibbundler`-related issue on my ARM Mac with incorrectly updating the executable, which will need to be fixed in the dylibbundler project) when `macos_bundle_libs` is `True`, and correctly builds the bundles without bundling the libraries when that is `False`.  So, I think this is ready for another review.

vLKp:
>  It seems that `MakeBundle()` returns one value, which a list of `Dir` objects

In `master`, yes.  I changed that in the experimental branch, too.  Check the second diff - https://github.com/dxx-rebirth/dxx-rebirth/commit/3af1999b2822b1f962da0fcbb78546607ab56b17#diff-34fc8112b90f1e14f86fd5dfb3d9e9b18a659427571989e32f7c32f52fc7a052

`MakeBundle` is a method in a supporting script committed in the Rebirth repository, so it is easy to patch that as part of this change.  `MakeBundle` has a fair bit of old logic that I'd like to revamp, which I was hoping would be tested as a side effect of you testing the experimental branch.

Kreeblah:
Ah.  I completely missed that file.

I just tested that branch directly, and it works fine as-is in all the use cases I have here (both Intel and ARM Macs, both with and without bundling libraries).  So, if you merge that one, that should resolve this, and I can just delete my fork.

Thanks again!

vLKp:
I pushed that experimental branch to `master`.  I amended the newest commit to have a more informative log message than the one I had used on the branch, but the committed files are the same as what you tested.  My push also has part of your earlier work visible in its history, since making dylibbundler optional is a useful feature for people who would prefer to avoid both dylibbundler and frameworks.

I am closing this pull request since the experimental branch conflicts with your merge, but also does everything you needed.

Kreeblah:
Sounds good.  Thank you so much!

Issue #520, opened 2020-07-11
Runtime bug: Host left the game


snytek:
### Environment

https://www.dxx-rebirth.com/download/dxx/rebirth/dxx-rebirth_20200630-win.zip

#### Operating System Environment

* Microsoft Windows
  * [ ] Microsoft Windows XP
  * [ ] Microsoft Windows Vista
  * [ ] Microsoft Windows 7
  * [ ] Microsoft Windows 8
  * [ ] Microsoft Windows 8.1
  * [X] Microsoft Windows 10

#### CPU environment

* [ ] x86 (32-bit Intel/AMD)
* [X] x86\_64 (64-bit Intel/AMD)
* [ ] ARM (32-bit)
* [ ] ARM64 (64-bit; sometimes called AArch64)

#### Game environment

This issue occurs in different 1-level missions or in the last level of a set.

### Description

A multiplayer game (1 level mission) is started and clients join (internet with port forwarding).
Reactor is blown, countdown starts
Either the host leaves the mine by entering the exit or gets killed on the way he gets to the summary screen.
At the bottom the the countdown is still visible and counting down about 5 seconds but then suddenly stops. 
Alle clients get the message "Host left the game"

#### Regression status

Unknown.

### Steps to Reproduce

For reproducing steps see my description.
It happens all the time and should be easy to reproduce.

* Engines affected:
  * [ ] D1X-Rebirth
  * [X] D2X-Rebirth

* Game modes affected:
  * [ ] Single player
  * [X] Multiplayer competitive
    * [X] Anarchy
    * [ ] Team Anarchy  <- not tested
    * [ ] Robo-Anarchy  <- not tested
    * [X] Capture the Flag
    * [X] Bounty
  * [ ] Multiplayer cooperative <- not tested

vLKp:
This probably impacts all multiplayer modes.  The loss of statistics is unfortunate, and worth fixing, but unless this turns out to be very easy to fix, it will likely not get fixed very soon.

MasonP57:
I can confirm that this affects all multiplayer issues. I played Rebirth 0.61 anarchy, with the Kill Limit option enabled. After we finished our game and all the survivors reached the exit, all the non-host clients crashed.

vLKp:
The original report says the clients get a message that the host left.  It said nothing about a crash.  Were you able to collect a backtrace from the crash?

MasonP57:
I will try to test it again and see. I don't know what a backtrace is.

[...]

PR #619, opened 2022-01-18
Send periodic endlevel packets after last level

arbruijn:
Add periodic calls to send_endlevel_packet in kmatrix if
Control_center_destroyed is false.

Needed because after finishing the last level, kmatrix is called from a
different place where Control_center_destroyed is already false.

It looks like this issue was introduced in Rebirth 0.56 with the
split off of the ipx code (febe5d1). That commit removes the call to
multi_endlevel_poll2 from kmatrix_view, which did the same periodic
sends. (Maybe because in D1 Control_center_destroyed was always true
in kmatrix so it was not needed there.)

Reported-by: snytek <#520>
Fixes: febe5d1 ("Abstracting networking protocols")

### DIFF
--- a/similar/main/kmatrix.cpp
+++ b/similar/main/kmatrix.cpp
@@ -202,6 +202,7 @@ struct kmatrix_window : window
 	}
 	grs_main_bitmap background;
 	fix64 end_time = -1;
+	fix64 last_endlevel_time = 0;
 	kmatrix_network network;
 	kmatrix_result &result;
 };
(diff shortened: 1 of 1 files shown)


vLKp:
As I understand #520, there were 2 problems reported:

- Guests were informed the host had left, even if the host had not yet returned to the main menu
- Guests crashed when the host left

Your proposed patch will probably fix the first issue.  I left #520 open intending to reproduce and fix the second issue first, since even with your change, eventually the host will return to the main menu.  When that happens, guests will get this disconnect message and, according to [@snytek's comment](https://github.com/dxx-rebirth/dxx-rebirth/issues/520#issuecomment-774707866), then crash.

PR #621, opened 2022-01-29
Re-add macOS OpenGL test

Kreeblah:
https://github.com/dxx-rebirth/dxx-rebirth/pull/543 disabled this test, as the OpenGL libraries are in a framework in macOS, and I wasn't really sure how to address that.  This corrects that, and now builds correctly on macOS.

I'm unable to test it on Linux or other platforms, but I believe the test should still pass on those.  However, it does need to be verified.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1479,11 +1479,13 @@ static void terminate_handler()
 		if self.user_settings.opengles:
 			# GLES builds do not use the GL utility functions
 			return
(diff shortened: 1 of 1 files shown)


vLKp:
I reworked this a bit to use a Python ternary, and drop an unnecessary `else` clause.  The result builds for both Linux hosts and Windows 64 cross-compiled from Linux.

PR #649, opened 2022-08-01
Packages for Windows, Linux, and macOS

raptor:
Packages for Windows, Linux, and macOS can now be created with a Github workflow under the Actions section. I've tested these on all three platforms (Linux x86_64, macOS 10.15, Windows 10 x64) with d2x-rebirth and it works!

I set the packages to be manually triggered by a workflow run. You can download the artifacts at the bottom of the run. 

# Notes

* All bundles compiled with `SDL2`
* `-hogdir` and other command line options work for all bundles.
* There is one workflow file `packages.yml` that triggers the packaging scripts i put under `contrib/packaging` in the repo.

## Windows

This is created with `msys2` and `mingw-w64-x86_64`. The ZIP mostly matches your release one, but I couldn't find some of the `.txt` files that exist in your archive.

Example run:
```
d2x-rebirth.exe -hogdir c:\games\d2data
```

## Linux

The AppImages created need to made executable: `chmod a+x d2x-rebirth-abcdeff.AppImage`

Example run:
```
d2x-rebirth-abcdeff.AppImage -hogdir ~/games/d2data
```

## macOS

These are unsigned `.app` files that are zipped up. Running requires an additional command line parameter to run. This is build for macOS 10.5, but that will be deprecated soon.

Example run:
```
open D2X-Rebirth.app --args -hogdir ~/games/d2data
```

Hope this helps.

### DIFF
--- a/common/include/3d.h
+++ b/common/include/3d.h
@@ -35,7 +35,11 @@ COPYRIGHT 1993-1998 PARALLAX SOFTWARE CORPORATION.  ALL RIGHTS RESERVED.
 #include <array>
 
 #if DXX_USE_OGL
+#if defined(__APPLE__) && defined(__MACH__)
+#include <OpenGL/gl.h>
+#else
 #include <GL/gl.h>
+#endif
 #endif
 
 #if DXX_USE_EDITOR


vLKp:
Thank you for looking into this.  I looked at Github Actions back when they first came out, but found the documentation difficult to use, and abandoned it.

I am leery of having Github publishing build artifacts that I cannot test, but I will consider this for merge anyway.

raptor:
Yeah I had to have a friend test the macOS one. I could divide up the workflows by platform and then you could only post the windows one as official (if it works) and the others as 'untested, unsupported, danger, etc.'

AlumiuN:
The only problem I can see with this as things currently stand is that D1X sounds wrong under SDL2, as per issue #474.

vLKp:
I do all my testing and development on Linux, so I'm not set up to test the Windows build, other than using Wine on it.

With regard to marking the other platforms as unsupported, is there a way to do that in the yaml, or do you mean that the README should state that and I should trust users to read the documentation?  After reading through the Actions documentation, I can't tell whether this is supported or not.

I noticed on your fork of the rebirth repository that the artifacts aren't download links unless I'm logged in to Github.  I can't fetch them anonymously.

I can deal with splitting up the yaml file if that turns out to be necessary.  I wish the Actions were a bit easier to use though.  The documentation hasn't improved much, and the yaml files have too much magic for my taste.

On closer inspection of the shell scripts, there are a few things I would like to change.  Notably, I don't want to have it `git clone` a repository of unspecified state just to get one tool (dylibbundler).  I can take care of fixing this up, once I get set up to test the changes.  From what I've read, that involves pushing the changes to `master` repeatedly until it works, which seems like it would make a mess.  This is one of the reasons I originally abandoned the idea of using Github Actions.  I could never find a reasonable way to test it without polluting the public repository history with failed attempts.  Is there a better approach now?  I could fork the repository to my user account and test there, but this feels like an ugly workaround for a basic missing feature.

raptor:
I meant to separate the 'jobs' from the YAML so you theoretically could auto-publish an artifact as a 'release'. Right now I only used the workflow to create the artifacts, not publish them. I figured you could download them and publish them at your convenience.

Everything else you said was my experience, too - I created a fork and made about 50+ changes until I had it right. Then I cloned locally, deleted the github fork, squashed the commits locally, created a new fork, and then made the pull request. I don't know of a better approach yet.

vLKp:
I see.  I feel a bit better about not finding a better way if you also cannot find a clean approach.

raptor:
Actually, maybe a better way is to do github actions on a separate branch then on pull request allow 'squashing' via:

https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests

If you tell me your 'ideal' scenario and any other problems you currently see, I can work on them over the next couple days since I'm already in the thick of it. Then I'll make a new pull request and try the squashing (if you make sure it's enabled on the main repo).

vLKp:
For Mac, I'd prefer that it use `git archive` or `git clone --depth=1` to fetch macdylibbundler without history.  I'd also prefer to pin it to a specific good commit, instead of letting it pull whatever is current at the instant it runs.

raptor:
I fixed `macdylibbundler` to a specific revision. I also split out the jobs so every change wouldn't require running them all. I wonder, do you have to option to squash these?

vLKp:
The Github UI does not offer an obvious way to do that, but you could squash them locally and force-push the result over the branch you used as the basis for this pull request.  I could also handle that for you as part of merging it, since presently the Github UI is not offering a web-based merge button on this pull request, so I will need to do this through the CLI anyway.

PR #651, opened 2022-08-10
Use custom fixed-point audio resampler

raptor:
- Hopefully fixes issues with poor quality resampling when using SDL_AudioCVT
- Converts 11025 Hz or 22050 Hz samples to the default 44100 Hz outputs
- Uses high order brick wall filter like Sound Blaster 16

I expect this needs some cleanup and I welcome any comments or criticism. I'm not quite sure what coding style to follow, so I welcome feedback on this, too.

Also, this code is 4x-8x faster than SDL_ConvertAudio for this game.

### DIFF
--- a/similar/arch/sdl/digi_mixer.cpp
+++ b/similar/arch/sdl/digi_mixer.cpp
@@ -153,6 +153,123 @@ void digi_mixer_close() {
 namespace dsx {
 
 /*
+ * Blackman windowed-sinc filter coefficients at 1/4 bandwidth of upsampled
+ * frequency. Chosen for linear phase and approximates ~10th order IIR
+ *
+ * MATLAB/Octave code:
+
+	N = 51;   % Num coeffs (odd)
+	B = 0.25;  % 1/4 band
+	half = (N-1)/2;
+	n = (-half:half); % the sample index
+	% Windowed sinc
+	h_ideal = 2 * B .* sinc(B*n);
+	h_win = blackman(N);
+	b = h_win .* h_ideal.';
+	% Convert to fix point to ultimately apply to signed 16-bit data
+	b_s16 = int32(round(b * (2^16 -1)));  % coeffs!
+
+ */
+#define FILTER_LEN 51
+static int32_t coeffs_quarterband[FILTER_LEN] =
+{
+		0, 0, -7, -25, -35, 0, 94, 200, 205, 0, -395, -751, -702, 0, 1178, 2127,
+		1907, 0, -3050, -5490, -5011, 0, 9275, 20326, 29311, 32767, 29311,
+		20326, 9275, 0, -5011, -5490, -3050, 0, 1907, 2127, 1178, 0, -702,
+		-751, -395, 0, 205, 200, 94, 0, -35, -25, -7, 0, 0
+};
+
+// Coefficient set for half-band (e.g. 22050 -> 44100)
+static int32_t coeffs_halfband[FILTER_LEN] =
+{
+		0, 0, -11, 0, 49, 0, -133, 0, 290, 0, -558, 0, 992, 0, -1666, 0, 2697, 0,
+		-4313, 0, 7086, 0, -13117, 0, 41452, 65535, 41452, 0, -13117, 0, 7086, 0,
+		-4313, 0, 2697, 0, -1666, 0, 992, 0, -558, 0, 290, 0, -133, 0, 49, 0, -11,
+		0, 0
+};
+
+// Fixed-point FIR filtering
+// Not optimal: consider optimization with 1/4, 1/2 band filters, and symmetric kernels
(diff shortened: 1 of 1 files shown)


vLKp:
That speed increase is impressive.  I will try to provide a detailed code review within the next few days.

vLKp:
My apologies for the delay.  The Github inline code review interface has not been working well, and I was hoping that with a few days and a forced refresh, it would recover.  It hasn't.

https://github.com/dxx-rebirth/dxx-rebirth/blob/5165efbc46b9f31a4f36b935322387e86955894a/similar/arch/sdl/digi_mixer.cpp#L155

These tables are shared between the games, so they should be in a `namespace dcx` block.  Code or data which varies in Descent 1 versus Descent 2 should be in `namespace dsx`.

https://github.com/dxx-rebirth/dxx-rebirth/blob/5165efbc46b9f31a4f36b935322387e86955894a/similar/arch/sdl/digi_mixer.cpp#L174

This should be `constexpr`, since the table is never modified.  Also, where possible, I prefer `std::array<T, N> id` over `T id[N]`.

https://github.com/dxx-rebirth/dxx-rebirth/blob/5165efbc46b9f31a4f36b935322387e86955894a/similar/arch/sdl/digi_mixer.cpp#L183

Same comments.

https://github.com/dxx-rebirth/dxx-rebirth/blob/5165efbc46b9f31a4f36b935322387e86955894a/similar/arch/sdl/digi_mixer.cpp#L193

- Parameters should be `const` when possible.
- Length variables should be `std::size_t`.  In this case, `int32_t *coeffs, int coeffsLen` could be combined as `std::span<int32_t> coeffs`, and let the compiler deduce the extent.

https://github.com/dxx-rebirth/dxx-rebirth/blob/5165efbc46b9f31a4f36b935322387e86955894a/similar/arch/sdl/digi_mixer.cpp#L197

When the loop iteration variable (here, `nn`) is only updated by the loop control statement, I prefer using `xrange` to make this clearer:

```c++
for (const unsigned nn : xrange(signalLen))
```

Note that `xrange` requires an unsigned input.  `int` is signed; `std::size_t` is unsigned, so addressing one of the earlier requests will implicitly handle this requirement.

https://github.com/dxx-rebirth/dxx-rebirth/blob/5165efbc46b9f31a4f36b935322387e86955894a/similar/arch/sdl/digi_mixer.cpp#L199

Spelling error: `Determin` is missing an `e`.

https://github.com/dxx-rebirth/dxx-rebirth/blob/5165efbc46b9f31a4f36b935322387e86955894a/similar/arch/sdl/digi_mixer.cpp#L200-L201

Use an unsigned type for array indexing.

https://github.com/dxx-rebirth/dxx-rebirth/blob/5165efbc46b9f31a4f36b935322387e86955894a/similar/arch/sdl/digi_mixer.cpp#L205

I suggest:

```c++
for (const unsigned kk : xrange(min_idx, max_idx + 1))
```

Note that `xrange` follows the STL pattern that `end` is excluded, hence the need to add 1 here to match your use of `<=`.

https://github.com/dxx-rebirth/dxx-rebirth/blob/5165efbc46b9f31a4f36b935322387e86955894a/similar/arch/sdl/digi_mixer.cpp#L207

Prefer use of C++ `static_cast<>` instead of the functional style cast used here.

https://github.com/dxx-rebirth/dxx-rebirth/blob/5165efbc46b9f31a4f36b935322387e86955894a/similar/arch/sdl/digi_mixer.cpp#L224-L228

Where possible, use a syntax that allows the compiler to readily generate a bulk store operation.  For this purpose, it might be better, although slightly wasteful, to zero-fill the entire `output` buffer before beginning, and then not need to zero slices of it during iteration.

https://github.com/dxx-rebirth/dxx-rebirth/blob/5165efbc46b9f31a4f36b935322387e86955894a/similar/arch/sdl/digi_mixer.cpp#L259-L261

`coeffs` can be made `const auto`:

```c++
const auto &coeffs = upFactor ? coeffs_halfband : coeffs_quarterband;
```

https://github.com/dxx-rebirth/dxx-rebirth/blob/5165efbc46b9f31a4f36b935322387e86955894a/similar/arch/sdl/digi_mixer.cpp#L289

If a statement is not needed, delete it, rather than comment it out.

https://github.com/dxx-rebirth/dxx-rebirth/blob/5165efbc46b9f31a4f36b935322387e86955894a/similar/arch/sdl/digi_mixer.cpp#L305-L306

`unique_ptr` provides value by managing lifetime.  Defer the `.release()` until after `convert_audio` is done, so that the memory moves directly from a local `unique_ptr` into the `RAIIMix_Chunk`.

raptor:
Hi, I have been quite busy but hope to get to this soon.

vLKp:
I pulled this into my working copy and am working on the cleanups locally.  There is no need for you to work on this further.

PR #652, opened 2022-08-10
GitHub packaging

raptor:
Continuation of #649

Squashed all commits together and re-opened a pull request. Also created a legitimate branch for this instead of working off of master.


Kreeblah:
I'm not sure what vLKp's thoughts are on the macOS parts, but I had a couple of thoughts around those in these builds.  First off, unless you really want to pin a specific `macdylibbundler` version, it does exist in Homebrew under the `dylibbundler` package name.  So, you could add that to your `brew` installation statement to satisfy the dependency, which seems like it'd simplify things.

Second, there is already functionality in the `SConstruct` script to run `macdylibbundler`, if you pass `macos_bundle_libs=1` to your `scons` command.  So, the only thing that would really be required to generate the same output as your script currently does would be to compress the `.app` bundles.  The only path to generate compressed files in the `SConstruct` script today is part of the notarization workflow, so that'd need to be handled outside of that process.

The second change in particular would exercise more of the regular build process, which could be valuable for noticing issues that might impact it on the off chance that somebody modifies that part of the build process.

vLKp:
Since I have no ability to test the MacOS build, and I rarely change anything specific to the MacOS build, I have no preference as to how that part should be done.  I am happy to take suggestions from people who use MacOS on a regular basis as to the right way to handle the MacOS build.

The version pin was done at my request, because the earlier iteration was pulling `HEAD` from the macdylibbundler repository.  I didn't like the idea that the workflow could randomly fail because someone in that repository pushed a broken commit.  Although I am wary of pulling arbitrary unreleased commits, I am willing to trust that any macdylibbundler build that gets released to `brew` is good enough to use here.  Therefore, I have no objection to using `brew` to install it, and to omitting any version pinning of that `brew` install.

I like the idea of maximizing the amount of regular code covered by these workflows, although I am happy just to have automated testing of MacOS builds.  I think currently the first I know of a MacOS-specific break is when someone reports it as an issue.

Kreeblah:
Ah, that makes sense.

The Homebrew policy is to only allow packages for release-tagged versions plus any patches required for compilation on macOS which can't be pulled from upstream for some reason (such as SDL 1.2 never making another stable release, even though they still accept fixes into the code base; in that case, the last released SDL 1.2 code archive is still pulled, with specific patches then applied to it).  So, the Homebrew builds should be reliable, as they should always come from source that the `macdylibbundler` project believes to be of release quality.

raptor:
That is a good idea. I will use brew and test this.

raptor:
`dylibbunder` is now brought in with `brew` and the `scons`  option `macos_bundle_libs=1` is used instead. This works well and the macOS .app runs fine.

vLKp:
I merged the above change, but I don't see any indication it actually did anything.  No packages were published.  No logs were generated.  I see no sign of errors, or any indication Github is even aware that Actions are now configured.

raptor:
I just sync'd my clone with upstream, and after a delay the workflows showed up. Maybe there is something that needs to be enabled in the settings?

*Update* FYI, I just ran all three workflows and they all failed. Your switch to c++20 is too new for the compilers on macOS and Linux, and the Windows one had a syntax error. The compiler on my machines are also too old.

vLKp:
> Maybe there is something that needs to be enabled in the settings?

Maybe.  As I mentioned in the prior pull request, I found the Github Actions documentation hard to read, and I now don't recall most of it, so if there are any special requirements, I probably have not done anything to meet them.

> Your switch to c++20 is too new for the compilers on macOS and Linux

That is strange.  I carefully avoided using anything that does not work in gcc-10, the oldest still supported gcc.

> the Windows one had a syntax error

I see that locally now, too.  I must have missed running a Windows build before publishing.  Usually I catch things like that.

> The compiler on my machines are also too old.

What compiler(s) are you using?  gcc-10 and clang-13 are both sufficient to handle everything I am doing, and I picked gcc-10 because even Debian stable can get gcc-10.  I probably would raise to gcc-11 if that did not exclude Debian users.

raptor:
OK, I see Ubuntu 20.04 (what github uses) can get that - I will update these building scripts.

Kreeblah:
For the macOS part, it builds fine locally for me, but I'm on 12.5.  I know Apple limits which Xcode versions can be installed on which major OS releases, so it might be worth trying to bump up the host version from 10.15 to something higher?  I believe it should still be able to target 10.14, even if you bump it to 12.5.

maggu2810:
Many thanks for that contribution.
I wanted to test the linux packaging - the AppImage build. To trigger the workflow I forked the repository.
After triggering and waiting for the succeeded workflow I downloaded the linux.zip and unpacked it.
After adding the executable flag to the AppImage files I tried to execute them.
Both executables complained about missing `libphysfs.so.1`.

```
[maggu2810@m3800 dxx-rebirth]$ ./d1x-rebirth-841ebcc.AppImage 
d1x-rebirth: error while loading shared libraries: libphysfs.so.1: cannot open shared object file: No such file or directory
[maggu2810@m3800 dxx-rebirth]$ ./d2x-rebirth-841ebcc.AppImage 
d2x-rebirth: error while loading shared libraries: libphysfs.so.1: cannot open shared object file: No such file or directory
```

vLKp:
As I read [@raptor's profile page](https://github.com/raptor), @raptor has not been active on Github since late October or early November, so you may not get a quick response.

---

To your implied question:

`libphysfs.so.1` is normally a symbolic link in `$libdir` (`/usr/lib` on x86 systems; `/usr/lib64` on x86_64 systems) that points to `libphysfs.so.3.0.2` (or such other minor version as appropriate).  The build could not have succeeded without this library present in the build filesystem.  As I understand it, AppImage should have bundled all needed libraries from the build filesystem into the output.  I do not know why it failed to do this.

I accepted the AppImage Github workflow because it seemed like it would be useful, but I was never able to get Github to run it, so I have no ability to debug its output.  I consider Linux distribution packaging the supported way to run Rebirth.  When I attempt to run the AppImage Github workflow, I get an empty window with an unlabeled spinner that never finishes or displays any messages.  If you propose specific changes to make the Github workflow produce better results, I will review them for inclusion.  Otherwise, you will need to use a non-AppImage workflow, or find someone for whom the workflow does not silently hang.

maggu2810:
Sorry for cross posting. I added another comment to the flatpak issue because it has been about "run on Steam Deck".
I read a little bit about AppImage and give it a try myself.
The current AppImage raptor added does not bundle shared objects, so it is not really what AppImage is about (IMHO).
I modified the workflow and packaging, so AppImages are created that bundle the dependencies and run on e.g. Steam Deck.

PR #654, opened 2022-08-15
Fix Linux/macOS builds with c++20 dialect

raptor:
Packaging! take 3...

Fixes for c++20

This supersedes #652


vLKp:
Installing gcc-10 and setting `$CC` are not strictly necessary.  Only the C++ compiler is directly accessed, so I think installing `g++10` and setting `$CXX` is sufficient..  However, installing both is likely harmless.

Issue #646, opened 2022-07-27
Runtime crash: Out-of-bounds access on ailp.goal_state


heftig:
### Environment

a1bdd81d419a2cf70d9f2fe3a2e4ec2ff0b3cab4

#### Operating System Environment

* [x] Arch

#### CPU environment

* [x] x86\_64 (64-bit Intel/AMD)

#### Game environment

DESCENT2.HOG: 7,595,079 bytes (v1.2)

EAF2 Level 1 (Same campaign as issue #506)

https://web.archive.org/web/20170728184712/http://descent2.com/eaf2/
https://web.archive.org/web/20170728184712/http://download.descent2.com/levels/descent2/siplayer/eaf2.exe (`unrar`able)

Save game shortly before crash: [heftig.sg9.zip](https://github.com/dxx-rebirth/dxx-rebirth/files/9203319/heftig.sg9.zip)

### Description

Screencast: ![screencast.webm](https://user-images.githubusercontent.com/204006/181367286-cf316c61-113a-44b0-b1b1-e1ca5a2f819a.webm)

Shortly after entering the reactor room, the game built with libstdc++ assertions (`-Wp,-D_GLIBCXX_ASSERTIONS`) crashes.

```
/usr/include/c++/12.1.0/array:208: constexpr std::array<_Tp, _Nm>::value_type& std::array<_Tp, _Nm>::operator[](size_type) [with _Tp = dcx::ai_static_state; long unsigned int _Nm = 10; reference = dcx::ai_static_state&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.
```

Inspecting the coredump shows:

```rust
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
#1  0x00007f8f4b7e03d3 in __pthread_kill_internal (signo=6, threadid=<optimized out>) at pthread_kill.c:78
#2  0x00007f8f4b790838 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007f8f4b77a535 in __GI_abort () at abort.c:79
[...]

Issue #650, opened 2022-08-04
[Feature Request] Pitch Speed Equal to Yaw Speed Option


narzepatel:
Similar but separate to the option of uncapped turning, I'm requesting an option to have the pitching speed be the same as the yaw speed. To clarify, this is not a sensitivity issue. With uncapped turning disabled, the pitch speed is clearly not as fast as the yaw speed.

vLKp:
When ship performance is capped, the cap for pitch speed is half the cap for yaw and roll.  This is implemented at the bottom of `kconfig_end_loop`:

https://github.com/dxx-rebirth/dxx-rebirth/blob/1c8ff1ac0f137aef68a50611c6714cd1cdf0a043/similar/main/kconfig.cpp#L1396-L1401

Removing the `/ 2` divisor would allow pitch to equal yaw.

narzepatel:
oh there it is, was searching for it but didnt think to check kconfig, thanks

vLKp:
I am undecided as to whether to require a custom build for this, so I will leave the issue open for now.  If there are other users who would like to see this feature included, I encourage them to leave a thumbs-up emoji response to the initial comment, so that I can gauge community interest.  Adding a build time flag to support this would be pretty easy.  Adding a user-configurable toggle would be a bit more work.

Given the release frequency of the project, one might say that putting anything into the next release is out of the question, since the date of the next release is presently indeterminate. ;)

vLKp:
Another user proposed an implementation for this in #657, which I merged.  It includes an in-game tunable to allow the player to activate/deactivate the new rule.  Please reopen if this feature does not satisfy your goal.

PR #657, opened 2022-08-20
Feat: Pitch unlocker

RedZakuII:
A friend of mine doesn't like to play with mouselook because it's "cheating". However, she also dislikes the halving of pitch speed as it discourages/punishes vertical movement. 

I made this patch to the engine so we could play the game together without the halving of pitch. It ended up feeling pretty nice in many respects, especially when dueling in maps that aren't merely flat planes. 

I figure maybe other people will also enjoy this. I haven't done a lot of C, so I hope the code is okay.

### DIFF
--- a/common/main/kconfig.h
+++ b/common/main/kconfig.h
@@ -148,6 +148,8 @@ void kconfig_end_loop(control_info &, const fix frametime);
 #define CONTROL_USING_JOYSTICK	1
 #define CONTROL_USING_MOUSE		2
 #define MOUSEFS_DELTA_RANGE 512
+#define LOCKED_PITCH_FACTOR 2
+#define FREE_PITCH_FACTOR 1
 #ifdef dsx
 namespace dsx {
 #if defined(DXX_BUILD_DESCENT_I)
--- a/common/main/multi.h
+++ b/common/main/multi.h
@@ -108,7 +108,7 @@ enum class multiplayer_data_priority : uint8_t
 #define MULTI_PROTO_UDP 1 // UDP protocol
 
 // What version of the multiplayer protocol is this? Increment each time something drastic changes in Multiplayer without the version number changes. Reset to 0 each time the version of the game changes
-#define MULTI_PROTO_VERSION	static_cast<uint16_t>(15)
+#define MULTI_PROTO_VERSION	static_cast<uint16_t>(16)
 // PROTOCOL VARIABLES AND DEFINES - END
 
 // limits for Packets (i.e. positional updates) per sec
@@ -836,6 +836,7 @@ struct netgame_info : prohibit_void_ptr<netgame_info>
 	ubyte   					team_vector;
 	uint8_t						SecludedSpawns;
 	uint8_t MouselookFlags;
+	uint8_t PitchLockFlags;
 	uint32_t					AllowedItems;
 	packed_spawn_granted_items SpawnGrantedItems;
 	packed_netduplicate_items DuplicatePowerups;
--- a/common/main/playsave.h
+++ b/common/main/playsave.h
@@ -159,6 +159,7 @@ struct player_config : prohibit_void_ptr<player_config>
 	HudType HudMode;
 	RespawnPress RespawnMode;
 	uint8_t MouselookFlags;
+	uint8_t PitchLockFlags;
 	using primary_weapon_order = std::array<uint8_t, MAX_PRIMARY_WEAPONS + 1>;
 	using secondary_weapon_order = std::array<uint8_t, MAX_SECONDARY_WEAPONS + 1>;
 	primary_weapon_order PrimaryOrder;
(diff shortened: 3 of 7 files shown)


vLKp:
On an initial read, the code looks fine.  Your editor seems to have reformatted many lines that you are not actually changing, though:

```
$ git diff --stat master refs/pull/657/head
 common/main/kconfig.h     |  2 ++
 common/main/multi.h       |  3 ++-
 common/main/playsave.h    |  1 +
 similar/main/kconfig.cpp  | 43 ++++++++++++++++++++++++++++++++-----------
 similar/main/menu.cpp     | 17 ++++++++++++++++-
 similar/main/net_udp.cpp  | 76 ++++++++++++++++++++++++++++++++++++++++------------------------------------
 similar/main/playsave.cpp | 10 ++++++++++
 7 files changed, 103 insertions(+), 49 deletions(-)

$ git diff --stat master refs/pull/657/head --ignore-space-change 
 common/main/kconfig.h     |  2 ++
 common/main/multi.h       |  3 ++-
 common/main/playsave.h    |  1 +
 similar/main/kconfig.cpp  | 27 ++++++++++++++++++++++++---
 similar/main/menu.cpp     | 15 +++++++++++++++
 similar/main/net_udp.cpp  |  4 ++++
 similar/main/playsave.cpp | 10 ++++++++++
 7 files changed, 58 insertions(+), 4 deletions(-)
```

Generally, I prefer that those two `diff --stat` runs yield very similar results.  In your case, it looks like 45 lines are just whitespace changes, with no functional change.  That makes the difference harder to review in the Github pull request web interface, and confuses `git blame` later, since it will default to attributing those changed lines to your pull, even though your change had no effect on the code.

Is your friend @narzepatel, the author of issue #650?  That issue requested this feature, but did not provide an implementation.

Barring some major discovery during a line-by-line review, I will likely merge this the next time I work on the code.

RedZakuII:
It looks like there's a define that VS code reformatted a little bit - the backslashes got moved so they were aligned. Would you like me to force push a fix for that?

He's not someone I know, but I hope he enjoys the change!

GoatMaiden:
It was me!

I was the one who also requested this specifically from @RedZakuII. 

I've been making a lot of levels lately, and the one thing on my mind is VERTICALITY. 

But I'm not seeing "Try banking and then turning left or right" as a feature. No, that's clearly a workaround. Even the pros have their pitch sensitivity turned up higher than turn, so we all know what we want.

The benefit of no-pitch-lock extends to joystick and keyboard users, unlike with "uncapped" turning, which only benefits mouse users. Furthermore, I've fought in many an epic air battle, regarding flightsims, and at least in WWI era planes, pitch was how you change direction fast, not turning. 

So there's really no precedent for it, and according to an original developer, Mike Kulas, "I think the main reason pitch & yaw had different speeds is we thought it was more difficult for people to manage choppy pitch over choppy yaw. We couldn't rely on the framerate being high enough, so we compensated that way."

So anyways, yeah. That's all I wanted to say. =D

I hope RedZakuII fixes those little issues you pointed out, and then I'll be supremely pleased to begin playing with friends and enemies using unhalved pitch.

vLKp:
@RedZakuII wrote:

> Would you like me to force push a fix for that?

If you are comfortable using git tools to reverse that, go ahead.  Otherwise, I can amend the commit when I merge it.

> He's not someone I know, but I hope he enjoys the change!

Interesting.  Two independent requests (across three people) for the feature, all in relatively short order.

---

@GoatMaiden wrote:

> I've been making a lot of levels lately

I am pleased to see that.  I thought level creation had largely dropped off, so it is nice to hear that people are still making them.

> Even the pros have their pitch sensitivity turned up higher than turn, so we all know what we want.

Actually, I keep all mine the same, and all maxed out. ;)

> The benefit of no-pitch-lock extends to joystick and keyboard users, unlike with "uncapped" turning, which only benefits mouse users. 

I disagree here.  The cap is applied late in the kconfig processing, so all user input types should be affected by removing the cap.  However, I can certainly see that mouse users might find it easier than others to generate the inputs required to get the most benefit out of it.  I believe I found in testing that with the cap removed, I could stack keyboard input and joystick input in the same axis to go faster than either would do alone.

> So there's really no precedent for it, and according to an original developer

I have not seen that claim before.  Do you have a citation for it?  I would find it quite unfortunate that this cap, which multiple maintainers have preserved for the sake of tradition, was rooted in a workaround for input limitations, rather than concern about game balance.

> I hope RedZakuII fixes those little issues you pointed out

Even if he doesn't, I'm happy to fix them as part of the merge.  The current delay is because I wanted to fix issue #646 at the same time, and that is proving more time consuming than expected to isolate.

RedZakuII:
I went ahead and fixed that big old block. Hopefully that'll make the commit look a little better.

GoatMaiden:
Yeah, all I've got is this Discord screencap I saw in the r/Descent Discord channel. https://cdn.discordapp.com/attachments/426799704396005377/1006050617124474880/IMG_2847.png

I don't know if it was ever clarified in even more detail as to whether they'd like it "patched" if possible.

PR #663, opened 2022-09-21
Fix OpenGL ES build on SDL2.

JohnnyonFlame:
This makes it possible to run DXX-Rebirth with OpenGL ES and SDL2 on devices such as Odroid Go Advance and similar. For a better experience more changes are needed (such as proper support for gamepad api and better menu functionality), but such discussion would happen somewhere else other than this patchset.

### DIFF
--- a/similar/arch/ogl/gr.cpp
+++ b/similar/arch/ogl/gr.cpp
@@ -72,7 +72,7 @@ namespace dcx {
 
 static int ogl_brightness_r, ogl_brightness_g, ogl_brightness_b;
 
-#if DXX_USE_OGLES
+#if DXX_USE_OGLES && SDL_MAJOR_VERSION == 1
 static int sdl_video_flags;
 
 #ifdef RPI
@@ -125,7 +125,7 @@ void gr_set_mode_from_window_size()
 
 }
 
-#if DXX_USE_OGLES
+#if DXX_USE_OGLES && SDL_MAJOR_VERSION == 1 
 static EGLDisplay eglDisplay=EGL_NO_DISPLAY;
 static EGLConfig eglConfig;
 static EGLSurface eglSurface=EGL_NO_SURFACE;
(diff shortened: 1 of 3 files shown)


vLKp:
Most of this looks fairly innocuous.  Why is the `gr_set_attributes` call moved up?

I notice this disables the side-by-side VR mode.  I can see why a handheld device would not benefit from that, but I wonder whether there are any systems which use OpenGL ES and can use side-by-side mode.  I am fine disabling it in GL ES mode for now and waiting to see if anyone reports it missing - but I'd prefer that it be fully removed when it is disabled.  As is, only the implementation is disabled, but the configuration knobs remain.  I can handle hunting down and disabling those though.

JohnnyonFlame:
It didn't compile for me with side-by-side, and it not being incredibly useful on the devices that would need OpenGL ES support I disabled it instead of porting.

> Most of this looks fairly innocuous. Why is the `gr_set_attributes` call moved up?

I should've commented this (or put an ifdef there)... SDL2 wants SDL_GL_SetAttribute calls BEFORE an OpenGL screen is created. This might not cause issues on some backends, but on others it causes initialization failures.

vLKp:
Right, devices that _must_ use OpenGL ES are mostly small/embedded systems.  However, I thought at one time Rebirth could use OpenGL ES even on desktop Linux, where side-by-side can be usable (though even there, it isn't helpful on a regular monitor).

vLKp:
After some experimentation, I see that at least in the OpenGL ES I have available here, side-by-side does not work anyway, because it relies on a call to `glDrawBuffer`, which is not defined.

vLKp:
Your patch broke OpenGLES + SDL1.  I included a fix for that in the merge, by excluding the new `SDL_GL_*` constants when building for SDL1, since those only exist in SDL2.

With those changes, OpenGLES builds for me with both SDL1 and SDL2.

PR #682, opened 2022-11-14
Use the SDL messagebox for SDL2 Linux builds

arbruijn:


vLKp:
Changes made after this was posted caused the result of merging it to fail to build.  I intend to get it to build at some point, but since its absence is not critical, fixing it has not been a priority.  Sorry for the delay.  I hope to get to it soon.

PR #683, opened 2022-11-27
Make sure gun point is calculated when firing

arbruijn:
Gun point calculation is skipped in do_ai_frame if dist_to_player >= 200 but a bot might still fire a homing weapon if it is further away.
Add extra call to calc_gun_point in ai_do_actual_firing_stuff when a homing weapon is fired without visibility to make sure a valid gun point is used.

This is an original game bug, shown for example in this video: https://www.youtube.com/watch?v=2dP51znGh7o

### DIFF
--- a/similar/main/ai.cpp
+++ b/similar/main/ai.cpp
@@ -2792,7 +2792,7 @@ static int maybe_ai_do_actual_firing_stuff(object &obj)
 }
 
 #if defined(DXX_BUILD_DESCENT_I)
-static void ai_do_actual_firing_stuff(const d_robot_info_array &Robot_info, fvmobjptridx &vmobjptridx, const vmobjptridx_t obj, ai_static *const aip, ai_local &ailp, const robot_info &robptr, const fix dist_to_player, const vms_vector &gun_point, const robot_to_player_visibility_state &player_visibility, const int object_animates, const player_info &player_info, robot_gun_number)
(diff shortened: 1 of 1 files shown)


vLKp:
As I read the code, the path patched here is only sometimes wrong.  In some cases, the gun point will be set correctly by the caller, and in other cases it will not.  If so, it seems wrong to recompute the gun point when it is already filled in properly.  I would prefer not to compute the gun point when it has already been set up properly.  However, based on this pull request, I can adjust for that.

At this time, the YouTube video appears to be unavailable.

arbruijn:
Fair enough. I see now the code already skips firing if the gun_point is (0, 0, 0) in a few places, maybe that's a better fix?

Strange that the YouTube video doesn't play, it does for me. I've downloaded it here https://2ar.nl/dl/homing-mech-glitch.mp4

vLKp:
Reposting my closing comment, which Github seems to have lost:

---

I think testing for 0,0,0 is itself a bug, because it is theoretically possible for a robot to have its gun at that position.  This is very unlikely with the size of a robot relative to 1 unit of position in the engine, but not impossible.  I chose to fix both that and the double-computation by storing the gun point in a `std::optional<vms_vector>`, which is valueless when the the gun point has not been computed.  This both eliminates the need for a magic location of `0,0,0` to represent "unknown" and allows the game to defer computing the gun point until it is needed.

I ultimately did manage to get the YouTube video, once I switched to requesting a lower resolution copy.

PR #690, opened 2023-01-07
Fix macOS compilation

Kreeblah:
Looks like there are a couple instances of `std::ranges` that need to be switched over to the ranges backport for macOS compilation to succeed.


vLKp:
A few weeks ago I was wondering whether Apple had yet updated to a clang with better std::ranges support.  Now I know. :)  I will merge this when next I work on the code.

PR #693, opened 2023-01-14
Fix or disable compiler warnings warnings

JoeNotCharles:
The flatpak default build setup enables some extra warnings with -Werror. They had 4 failures: 2 of them detected valid bounds checking mistakes, 1 complained about some technically correct but dangerous code that I thought was worth changing, and 1 I had to disable because it was triggering a compiler bug in GCC 12.x.

### DIFF
--- a/common/include/window.h
+++ b/common/include/window.h
@@ -21,6 +21,7 @@
 #include "console.h"
 
 #include <assert.h>
+#include <memory>
 
 #include "fwd-window.h"
 #include "event.h"
@@ -36,7 +37,7 @@ private:
 	class window *next = nullptr;				// the next window in the doubly linked list
 	uint8_t w_visible = 1;						// whether it's visible
 	uint8_t w_modal = 1;						// modal = accept all user input exclusively
-	bool *w_exists = nullptr;					// optional pointer to a tracking variable
+	std::shared_ptr<bool> w_exists;					// optional pointer to a tracking variable
 public:
 	explicit window(grs_canvas &src, int x, int y, int w, int h);
 	window(const window &) = delete;
@@ -85,9 +86,9 @@ public:
 		return wind.prev;
 	}
 
-	void track(bool *exists)
+	void track(std::shared_ptr<bool> exists)
 	{
-		assert(w_exists == nullptr);
+		assert(!w_exists);
 		w_exists = exists;
 	}
 };
--- a/common/main/newmenu.h
+++ b/common/main/newmenu.h
@@ -354,7 +354,7 @@ struct newmenu : newmenu_layout, window
 		newmenu_layout(title, subtitle, filename, src, tiny_mode, tabs_flag, citem_init, draw_box), window(src, x, y, w, h)
 	{
 	}
-	int *rval = nullptr;			// Pointer to return value (for polling newmenus)
+	std::shared_ptr<int> rval;		// Pointer to return value (for polling newmenus)
 	virtual window_event_result event_handler(const d_event &) override;
 	static int process_until_closed(newmenu *);
 };
--- a/d2x-rebirth/main/movie.cpp
+++ b/d2x-rebirth/main/movie.cpp
@@ -23,6 +23,8 @@ COPYRIGHT 1993-1999 PARALLAX SOFTWARE CORPORATION.  ALL RIGHTS RESERVED.
  *
  */
 
+#include <memory>
+
 #include <string.h>
 #ifndef macintosh
 # include <sys/types.h>
@@ -379,8 +381,8 @@ movie_play_status RunMovie(const char *const filename, const std::span<const cha
 	}
 	const auto reshow = hide_menus();
 	auto wind = window_create<movie>(grd_curscreen->sc_canvas, std::move(mvestream));
-	bool exists = true;
-	wind->track(&exists);
+	auto window_exists = std::make_shared<bool>(true);
+	wind->track(window_exists);
 	init_subtitles(wind->SubtitleState, subtitles);
 
 #if DXX_USE_OGL
@@ -392,7 +394,7 @@ movie_play_status RunMovie(const char *const filename, const std::span<const cha
 	gr_set_mode(hires_flag ? screen_mode{640, 480} : screen_mode{320, 200});
 #endif
 
-	while (exists)
+	while (*window_exists)
 		event_process();
 	wind = nullptr;
 
--- a/similar/main/menu.cpp
+++ b/similar/main/menu.cpp
@@ -2337,10 +2337,13 @@ window_event_result browser::callback_handler(const d_event &event, window_event
 			{
 				const size_t len_newpath = strlen(newpath.data());
 				const size_t len_item = strlen(list[citem]);
-				if (len_newpath + len_item < newpath.size())
+				// Add a separator if the path doesn't already end with one.
+				const size_t len_sep = strlen(sep);
+				const char* sep_to_add = strncmp(&newpath[len_newpath - len_sep], sep, len_sep) ? sep : "";
+				const size_t len_sep_to_add = (sep_to_add == sep) ? len_sep : 0;
+				if (len_newpath + len_item + len_sep_to_add < newpath.size())
 				{
-					const size_t len_sep = strlen(sep);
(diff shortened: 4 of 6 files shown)


vLKp:
https://github.com/dxx-rebirth/dxx-rebirth/pull/693/commits/edfb3c4b77a5a68106771ea7901102e599f54370 :

> If newpath ends in a directory separator, the EVENT_NEWMENU_SELECTED handler appends another copy of the separate before appending list[citem].

I think that this is backwards.  `strncmp` returns `0` if the strings match, and non-zero otherwise.  Then: `strncmp(&newpath[len_newpath - len_sep], sep, len_sep) ? sep : ""` yields the true path (`sep`) if the strings mismatch, and the false path (`""`) if the strings match.  Therefore, if the string _does not_ end in the contents of `sep`, a copy of `sep` will be used.  If the string _does_ end in the contents of `sep`, then `""` will be used.

I do not get a truncation warning for this file in gcc-10, gcc-11, or gcc-12, all x86_64-pc-linux-gnu.  However, on inspection, I think the warning is valid (though failure is very unlikely, since `newpath` is `PATH_MAX` long, and users are very unlikely to have a path that deep).  `snprintf` needs `len_sep + len_item + len_newpath` < `newpath.size()` for this to avoid truncation, and the test does not ensure that.

---

https://github.com/dxx-rebirth/dxx-rebirth/pull/693/commits/1f1903a3b98000719c200454fdc87ef52741eec4 :

https://github.com/dxx-rebirth/dxx-rebirth/blob/1f1903a3b98000719c200454fdc87ef52741eec4/similar/main/net_udp.cpp#L5735

I think this is wrong.  `Error` always terminates the program, so it should be used only when an unrecoverable condition occurs.  In this case, `data.size()` can be constrained in the caller, and in practice already is constrained.  There is only one caller of `->send_data_direct`:

```
$ git grep -w send_data_direct
common/main/multi.h:241:    virtual void send_data_direct(std::span<const uint8_t> data, playernum_t pnum, int needack) const = 0;
common/main/net_udp.h:31:   virtual void send_data_direct(std::span<const uint8_t> data, playernum_t pnum, int needack) const override;
similar/main/multi.cpp:1068:    multi::dispatch->send_data_direct(buf, pnum, priority);
similar/main/net_udp.cpp:5706:void dispatch_table::send_data_direct(const std::span<const uint8_t> data, const playernum_t pnum, int needack) const
```

That one caller is a template function, which constructs `data` to refer to `buf`:

https://github.com/dxx-rebirth/dxx-rebirth/blob/9fd0cc96a58b5d5dc92e91ae375d0e9b025b4f17/similar/main/multi.cpp#L1065-L1068

`multi_command<C>` inherits `std::array<uint8_t, command_length<C>::value>`:

https://github.com/dxx-rebirth/dxx-rebirth/blob/9fd0cc96a58b5d5dc92e91ae375d0e9b025b4f17/common/main/multiinternal.h#L131

`command_length<C>` is `std::integral_constant<unsigned, SIZE>`, where `SIZE` varies based on the command code `C` in question.

`len` is constrained by how it is computed in the preceding lines, and will be either `2` or `6`.  Therefore, this error path can occur only if `6` + the longest command exceeds the buffer size.  A quick scan of `multiinternal.h` suggests that the largest message is probably `MULTI_PLAYER_DERES` due to its use of `DXX_MP_SIZE_PLAYER_RELATED`, which is 108 in Descent 2.  The size of `buf` is a function of `struct UDP_mdata_info`, which is `454` for its field `mbuf`, plus some control bytes.  From that, I believe this error path is impossible.  As a safety measure, `multi_send_data_direct` could be modified to include a `requires()` to validate the length at compile time.

---

https://github.com/dxx-rebirth/dxx-rebirth/pull/693/commits/10a2b2d33792bf7e40d2a9bc93bebb22804088ea :

This is a known issue, and one of the reasons I do not routinely test gcc-12.  I considered using shared_ptr to solve this, but I am not happy with pessimizing the runtime behavior just to convince the compiler that the code is safe.  When I first encountered this, I started on a project to clean up the nesting semantics around windows so that the `exists` hack could go away.  That is rather involved, and was never finished, since I tried not to break anything along the way.

When I build with gcc-12 here, I get a diagnostic for `exists`, but not for `rval`.  In what environment did you see a warning for `rval`?  It seems like `rval` ought to be more likely to trigger the warning.

---

https://github.com/dxx-rebirth/dxx-rebirth/pull/693/commits/2644e0cb93c1870141305be604186ec76ba09562 :

https://github.com/dxx-rebirth/dxx-rebirth/blob/1f1903a3b98000719c200454fdc87ef52741eec4/similar/main/render.cpp#L1051-L1055

I think a better citation is https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104165 , part of which was moved to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=107986 .  I wonder if inlining `sort_seg_children` and/or adding range hints could convince gcc not to warn here.

vLKp:
The above changes were merged to master, using the older version of the branch, so Github did not flag it as a merge due to the force push that came in while I was testing.  I believe the merge addresses the issues raised by this pull request.  Please comment or reopen if you disagree.

JoeNotCharles:
> I think that this is backwards. `strncmp` returns `0` if the strings match, and non-zero otherwise.

Argh. I thought doubling the separator didn't make much sense. Too long since I've used strcmp directly... Fixed.

> I think this is wrong. `Error` always terminates the program, so it should be used only when an unrecoverable condition occurs. In this case, `data.size()` can be constrained in the caller, and in practice already is constrained.

It turns out the bounds check warning was already fixed by 385277b833f409b5adbc275e99d810232614b169, which I rebased past. Removed this check.

> This is a known issue, and one of the reasons I do not routinely test gcc-12. I considered using shared_ptr to solve this, but I am not happy with pessimizing the runtime behavior just to convince the compiler that the code is safe.

Since this code isn't in any hot path, I think the overhead of a shared_ptr is worth it as defensive programming - not so much to convince the compiler that the code is safe, but to make it less likely that refactoring will introduce use-after-free's.

Environment is g++ 12.1.0, with flags

CPPFLAGS: ['-Wno-sign-compare']
CXXFLAGS: ['-g', '-O2', '-ftabstop=4', '-Wall', '-Werror=extra', '-Werror=format=2', '-Werror=missing-braces', '-Werror=missing-include-dirs', '-Werror=uninitialized', '-Werror=undef', '-Werror=pointer-arith', '-Werror=cast-qual', '-Werror=missing-declarations', '-Werror=vla', '-pthread', '-funsigned-char', '-O2', '-g', '-pipe', '-Wp,-D_FORTIFY_SOURCE=2', '-Wp,-D_GLIBCXX_ASSERTIONS', '-fexceptions', '-fstack-protector-strong', '-grecord-gcc-switches', '-fasynchronous-unwind-tables', '-fstack-clash-protection', '-fcf-protection', '-fno-omit-frame-pointer', '-std=gnu++20', '-Werror=unused', '-Werror=useless-cast', '-Wimplicit-fallthrough=5', '-fvisibility=hidden', '-Wduplicated-branches', '-Wduplicated-cond', '-Wsuggest-attribute=noreturn', '-Wsuggest-final-types', '-Wsuggest-override', '-Wlogical-op', '-Wold-style-cast', '-Wredundant-decls']

> I think a better citation is https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104165 , part of which was moved to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=107986 . I wonder if inlining `sort_seg_children` and/or adding range hints could convince gcc not to warn here.

Thanks, updated the link.

I don't think we should change trivially correct code to work around an incorrect compiler warning. Since we know the code is safe, disabling the warning in this narrow scope has the lowest impact. (Well, this disables it for all GCC versions, plus clang which obeys the GCC pragmas, even if they handle the warning correctly, but I think that's a fair balance for readability.)

In comparison the warning about lifetime of "exists" and "rval" are theoretically correct, and we only know the code is safe by reasoning about the lifetime of wind and the effects of the while/event_process() loop, so I think it's fair to change the code to make its safety more obvious to both the compiler and human readers.

JoeNotCharles:
> The above changes were merged to master, using the older version of the branch, so Github did not flag it as a merge due to the force push that came in while I was testing. I believe the merge addresses the issues raised by this pull request. Please comment or reopen if you disagree.

Indeed, thanks! I think my version of the path check is easier to read so I'll send another pull request for it, but that's just a cleanup.

PR #696, opened 2023-01-23
More precise bounds check in EVENT_NEWMENU_SELECTED handler

JoeNotCharles:
This only includes space for an added separator if the path doesn't already contain one.

### DIFF
--- a/similar/main/menu.cpp
+++ b/similar/main/menu.cpp
@@ -2344,9 +2344,11 @@ window_event_result browser::callback_handler(const d_event &event, window_event
 				 * succeed in all reasonable cases.
 				 */
 				const size_t len_sep = strlen(sep);
-				if (len_newpath + len_item + len_sep < newpath.size())
+				const char *sep_to_add = strncmp(&newpath[len_newpath - len_sep], sep, len_sep) ? sep : "";
+				const size_t len_sep_to_add = (sep_to_add == sep) ? len_sep : 0;
+				if (len_newpath + len_item + len_sep_to_add < newpath.size())
 				{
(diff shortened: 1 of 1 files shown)


vLKp:
I considered doing that in my prior commit, and opted against it on the basis that this buffer is already so large that users should never trigger this check, so depriving them of one extra byte in some cases was not worth the added complexity.  Further, since this was only motivated by gcc-12 insisting that there was a potential for string truncation, I wanted to minimize the complexity so that it would recognize that the logic was sufficient to avoid truncation.

Is there some user-visible benefit to this, other than having 1 more byte available when the path is already nearly `PATH_MAX` (4096) bytes long?

JoeNotCharles:
No user-visible benefits. I mainly think the code is an improvement because the temp variable is more readable than embedding a long expression inside snprintf.

vLKp:
My apologies for the very late response.  I did not receive a GitHub notification that there was a response here.

---

I can see that argument.  This area was subsequently reworked in e1d68f58ab6a2fe43eb011555ab9a1582d523bab because gcc-14 was issuing a truncation warning for the snprintf, possibly on the overly cautious assumption that the input strings could get longer after being measured and before being used.  This PR no longer merges cleanly, but I think the rework followed a similar design of breaking expressions out into local variables.  For completeness, I checked out the last commit where this PR would merge cleanly, and found that gcc-14 warns both with and without the change in this PR.  This is not surprising, based on gcc's apparent expectation of a time-of-check/time-of-use error.

Closing, on the basis that (1) this does not merge cleanly in the current `master`, (2) the rework implemented something similar to your proposal, which I think addressed the intent of your change.

Issue #695, opened 2023-01-22
Build failure: sizeof(callsign_t) doesn't necessarily equals to CALLSIGN_LEN+1


mikrosk:
Believe or not but on my (big endian) g++ 7.5.0 sizeof(callsign_t) equals to 10 and not 9 (=CALLSIGN_LEN+1). And indeed, the standard doesn't forbid it: https://stackoverflow.com/a/25173547/21009.

It's easy to change the `#define` however it causes more serious trouble in `struct player_rw` which is, I presume, used for serialization or network transfer as the its assert requires hardcoded size of 116 (and I get 117...).

vLKp:
gcc below gcc-10 is not supported on current `master`.  When reporting a build failure, please attach the build log showing the failure.

---

Yes, structures can in some cases contain padding.  I added static asserts about structure sizes specifically because so much of the legacy code assumes no padding, so these asserts catch the problem at build time instead of manifesting in weird bugs at runtime when the game reads the wrong number of bytes from the file.

mikrosk:
So what would you suggest in this case? `static_assert` worked as expected (and I'm grateful for that) but I wasn't sure how to fix it. When "real" sizeof() is 8, the actual sizeof() is 8. When "real" sizeof() is 9, the actual sizeof() is 10, so it seems there's some align process in the works (and yes, I tried to use `__pack__`, no difference).

About the gcc version, yes, that's another pain. I had to revert back to commit 1bcc5c40 because on my platform gcc7 is the most stable/proven compiler.

Let me be clear - if this kind of build failure reports annoy you (because you are not seeing them in normal environment), I understand. My original aim was to use dxx-rebirth as an experiment on a low end platform but from I have seen, the code is long gone from the 8bpp rendering anyway and full of modern C++ constructs.

If you see value in fixing stuff like this (another tip: try to compile dxx-rebirth with `opengl=0` on a platform _without opengl at all`, you'll see quite a lot of build failures because the project basically expects OpenGL always installed), I'm happy to help, otherwise feel free to close both my issues and I wont bother you with similar reports.

vLKp:
I suggest updating to a currently supported compiler.  I doubt gcc-7 will get any interest from the upstream gcc developers at this point, so if you encounter any compiler bugs, you will be on your own.  I am surprised any current generation platform is stuck with such an old gcc.  I only moved the minimum up to gcc-10 because even Debian stable had access to that, and Debian is well known for being slow to update.

---

[...]

PR #699, opened 2023-01-24
Fix build with opengl=0 and no GL headers available

mikrosk:
As mentioned in https://github.com/dxx-rebirth/dxx-rebirth/issues/695#issuecomment-1400986448, current dxx-rebirth fails without at least some GL installation present. Fortunately, the fix is easy.

### DIFF
--- a/similar/main/config.cpp
+++ b/similar/main/config.cpp
@@ -41,7 +41,9 @@ COPYRIGHT 1993-1999 PARALLAX SOFTWARE CORPORATION.  ALL RIGHTS RESERVED.
 #include "u_mem.h"
 #include "physfsx.h"
 #include "nvparse.h"
+#if DXX_USE_OGL
 #include "ogl_init.h"
+#endif
 #include <memory>
 
 namespace dcx {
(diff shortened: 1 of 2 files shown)


vLKp:
This appears to disable some uses of OpenGL-specific members, but does not actually remove them from the relevant structure.  I would prefer that unused members be compiled out.  If you wish, I can add that when I next work on the code (and merge this at that time).

mikrosk:
Up to you - if it makes your life easier, I can remove the struct members, too. If you plan to do further changes and it's easier for you to do it along with them, I'll happily leave to you.

mikrosk:
I've just tried to compile this project on master (now that we have gcc-13 available on our platform) and this one still doesn't seem to be fixed.

Luckily my patch still applies cleanly and fixes the build. Any chance you could look into proper fix instead of this quick one?

vLKp:
The pull request was closed before I came back to it, and it dropped off my queue.  I've been busy with other projects, but I can try to get this when I start sweeping open issues and pull requests again.

PR #709, opened 2023-04-06
Make sure SDL2 uses specified audio format

arbruijn:
Use extended open function Mix_OpenAudioDevice to set allowed_changes to 0.
Prevents higher pitched sound on Windows.


vLKp:
This looks reasonable, but could you explain why Windows exhibits problems with the default setting?  Rebirth only opens audio in this one place, and never resets it.

arbruijn:
Apparently Windows prefers a 48kHz sample rate (maybe depending on sound card/drivers) and SDL2 uses that if allowed. While Rebirth assumes the sample rate is 44.1kHz.

PR #710, opened 2023-04-06
Fix Apple Clang errors and warnings

Kreeblah:
Xcode 14.3 was released a few days ago, which bumped Clang to 14.0.3 and libcxx to 15.x.  Some part of these changes caused compilation to generate some new warnings, with a few of those (the ones for set-but-unused variables) being promoted to errors.  This fixes those warnings, so the tree compiles cleanly on Apple Clang again.

Separately, the update also means that the ranges backports are no longer necessary for up-to-date versions of Apple Clang.  I was able to successfully compile the tree with the changes in this pull request and an additional change to set the preprocessor checks in `backports-ranges.h` to a nonexistant macro.  I haven't changed anything regarding that in this pull request, though.

### DIFF
--- a/common/main/cli.cpp
+++ b/common/main/cli.cpp
@@ -320,7 +320,7 @@ void CLIState::execute_active_line()
 	const char *p = m_line.c_str();
 	con_printf(CON_NORMAL, "con%c%s", g_prompt_strings[0], p);
 	cmd_append(p);
-	m_lines[0] = move(m_line);
+	m_lines[0] = std::move(m_line);
 	m_lines.emplace_front();
 	m_history_position = 0;
 	if (m_lines.size() > m_maximum_history_lines)
@@ -394,7 +394,7 @@ void CLIState::history_move(unsigned position)
 {
 	if (position >= m_lines.size())
 		return;
-	m_lines[m_history_position] = move(m_line);
+	m_lines[m_history_position] = std::move(m_line);
 	auto &l = m_lines[m_history_position = position];
 	m_line_position = l.size();
 	m_line = l;
--- a/similar/2d/font.cpp
+++ b/similar/2d/font.cpp
@@ -1030,7 +1030,7 @@ static std::unique_ptr<grs_font> gr_internal_init_font(const char *fontname)
 	//set curcanv vars
 
 	auto &ft_filename = font->ft_filename;
-	font->ft_allocdata = move(ft_allocdata);
+	font->ft_allocdata = std::move(ft_allocdata);
 	strncpy(&ft_filename[0], fontname, ft_filename.size());
 	return font;
 }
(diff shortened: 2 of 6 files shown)


vLKp:
I am pleased to see that Apple updated to libcxx 15.x, as this gets them to be only one version behind, again. ;)  Rebirth development has been substantially reduced while I waited for Apple clang to catch up to C++20.

Your changes look fine, and I will merge them.  However, I have a question about the changes in `common/main/cli.cpp`.  What warnings/errors were you getting that prompted this change?  Argument-dependent lookup should have been finding `std::move` on its own, so this change looks like it should not be needed.  This file builds fine with gcc, and has not been changed since July 2022.  It builds fine with x86_64-pc-linux-gnu-clang-14, and from what I recall, also worked fine in clang-15 before I removed that to downgrade to match Apple, so I find it odd that Apple clang-14 with libc++-15 would have a problem with it.

Kreeblah:
When I was looking to see what version they'd moved up to, I did notice that libcxx 16 was available, so I was initially hoping they'd just decided to catch up.  Unfortunately not, though.

As far as the specific warnings go, both `common/main/cli.cpp` and `similar/2d/font.cpp` started throwing the same warning, though this warning isn't being promoted to an error:

```
CXX dxx build/ common/main/cli.cpp
common/main/cli.cpp:323:15: warning: unqualified call to 'std::move' [-Wunqualified-std-cast-call]
    m_lines[0] = move(m_line);
                 ^
                 std::
common/main/cli.cpp:397:32: warning: unqualified call to 'std::move' [-Wunqualified-std-cast-call]
    m_lines[m_history_position] = move(m_line);
                                  ^
                                  std::
2 warnings generated.
```

```
CXX d1x build/ similar/2d/font.cpp
similar/2d/font.cpp:1033:23: warning: unqualified call to 'std::move' [-Wunqualified-std-cast-call]
    font->ft_allocdata = move(ft_allocdata);
                         ^
                         std::
1 warning generated.
```

From what I've been able to find, it seems that the LLVM folks decided to do this in order to guard against people with similarly-named methods (both `move()` and `forward()`) and instead warn unconditionally if they're specified without the `std::` namespace prefix.  https://reviews.llvm.org/D119670?id=408276 has the discussion on it I saw.

Kreeblah:
As far as why this would be warning on Apple Clang 14 and not the general Clang 14 builds . . . I'm unsure.  Given that this was something that was added to the project, it makes me wonder whether Apple is backporting specific changes to the versions they release, but I don't see anything in the release notes at https://developer.apple.com/documentation/xcode-release-notes/xcode-14_3-release-notes about that.

vLKp:
Thank you for the research.  I tend to agree with one of the commenters in the thread you cited, who asserted that this ought not be enabled in `-Wall`.  From my perspective, this is another case of clang issuing questionable warnings at the default level, and thereby encouraging people to ignore warnings or ignore clang.  (I keep the clang build working as best I can, but I got tired of it choking on things that I considered stylistic choices rather than actual issues, so I now consider it a lower tier target that I build-test eventually, but I do not use regularly, and take a skeptical view to any new warnings it issues.)

PR #712, opened 2023-04-21
Fix inaccurate handling of D2 super laser damage with quad-lasers

AlumiuN:
Upon a review of D2 weapon damage, it would appear that D2 did not apply the quad laser per-bolt damage penalty to super lasers; the code was written with it clearly being intended, but as SUPER_MAX_LASER_LEVEL was inexplicably set to 5, it only checked for a laser id between 1 and 5 (LASER_ID to LASER_ID + 4), however the super lasers used ID 30 and 31, and as such did not lose damage. (see LASER.C line 644 for reference)

### DIFF
--- a/similar/main/laser.cpp
+++ b/similar/main/laser.cpp
@@ -656,14 +656,13 @@ static void do_omega_stuff(fvmsegptridx &vmsegptridx, const vmobjptridx_t parent
 	}
 }
 
+// Note that this is only used for determining if the quad laser per-bolt damage penalty should be applied, and consequently excludes super lasers
 static int is_laser_weapon_type(const weapon_id_type weapon_type)
 {
 	return weapon_type == weapon_id_type::LASER_ID_L1 ||
 		weapon_type == weapon_id_type::LASER_ID_L2 ||
 		weapon_type == weapon_id_type::LASER_ID_L3 ||
-		weapon_type == weapon_id_type::LASER_ID_L4 ||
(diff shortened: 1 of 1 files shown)


vLKp:
The description here is a bit confusing, but after looking into it, I agree with you, and I will merge this change when next I work on the code.

---

`MAX_SUPER_LASER_LEVEL` is supposed to be 5, not 6, since the levels were intended to be 0-based, so the starting lasers were level 0 internally, and the best non-super lasers were level 3 internally.

---

Early Descent 1 Rebirth, laser.c logic for when to enable the penalty (which as you say effectively ignored the player, because the ID was wrong):

https://github.com/dxx-rebirth/dxx-rebirth/blob/b15b093caa6e8cdc3bd68f610c0ccff0cc208f58/main/laser.c#L292

Early Descent 2 Rebirth, laser.c logic for when to enable the penalty:

https://github.com/dxx-rebirth/dxx-rebirth/blob/0e65fa0caceab0b38aff0235df083d5290ba4240/main/laser.c#L666

[Use helper to check for laser type weapons](/dxx-rebirth/dxx-rebirth/commit/553dc95b0575fdc87d2054032a216baad3f4e935) changed the logic to correctly test for player laser weapons of all levels.  [Only reduce laser power in D2 mode](/dxx-rebirth/dxx-rebirth/commit/fc46e9a97c21daf1c3f375a5a7aa072d31e7afda) changed the logic to apply the penalty only to Descent 2, since pull #507 noted that the Descent 1 logic was present but not effective.

PR #714, opened 2023-05-09
macOS cross compilation fixes/instructions

Kreeblah:
I figured I'd look into seeing about cross-compiling for macOS from Linux, and I wanted to contribute my findings (along with some related patches).

It seems that cross-compiling for macOS relies on the [osxcross project](https://github.com/tpoechtrager/osxcross), which uses clang and LLVM.  It also includes a self-contained instance of the MacPorts package manager for macOS dependency management, which can be used to satisfy the dxx-rebirth dependencies.

Unfortunately, `dylibbundler` depends on some tools which are currently macOS-only, so it can't be used from Linux.  The `install_name_tool` application is specifically what I'm finding to be a blocker there, as there doesn't seem to be a port that compiles on Linux (except one that's been modified to process ELF binaries instead).  What that means is that the builds that come out of this can be run on Macs, but only if they have the libraries in the same location that MacPorts puts them on a Mac (`/opt/local/lib`), or they can have the libraries bundled in using `dylibbundler` on a Mac for further distribution, but as produced, the resulting builds can only be run on a limited subset of systems.

Still, it was an interesting exercise, and I figured I'd see whether it'd be worth including this info in the project documentation.


vLKp:
I think recording this where it can be easily found in the future is worthwhile.  I will look at getting this merged when next I work on the code.  I hoped to get to it this past weekend, but did not.

Issue #689, opened 2023-01-05
[build error] opengl enabled, but sdl configured with opengl disabled


spyjoshx:
OS: Fedora 37 (GNOME) x86_64
I installed the packages in INSTALL.txt (SLD_image-devel was not found, but was listed as optional so I ignored) and ran scons, but it stopped with the following full output:

```
scons: Reading SConscript files ...
===== D1X-Rebirth v0.61.0 0.60.0-beta2-696-gd2e190ec3d5e d2e190ec3d5e6257e82773e331e467bd20545cf7 =====
===== D2X-Rebirth v0.61.0 0.60.0-beta2-696-gd2e190ec3d5e d2e190ec3d5e6257e82773e331e467bd20545cf7 =====
DXX-Archive.3: compiling on 'linux'/'x86_64' for 'linux' into build/
DXX-Archive.3: checking version of C++ compiler 'g++' ... 'g++ (GCC) 12.2.1 20221121 (Red Hat 12.2.1-4)'
DXX-Archive.3: checking path to assembler ... b'as'
DXX-Archive.3: checking version of assembler b'as' ... 'GNU assembler version 2.38-25.fc37'
DXX-Archive.3: checking path to linker ... b'ld'
DXX-Archive.3: checking version of linker b'ld' ... 'GNU ld version 2.38-25.fc37'
DXX-Archive.3: CHOST: None
DXX-Archive.3: CXX: 'g++'
DXX-Archive.3: CPPDEFINES: [('PHYSFS_DEPRECATED', '')]
DXX-Archive.3: CPPPATH: ['common/include', 'common/main', '.', 'build/']
DXX-Archive.3: CPPFLAGS: ['-Wno-sign-compare']
DXX-Archive.3: CXXFLAGS: ['-g', '-O2', '-ftabstop=4', '-Wall', '-Werror=extra', '-Werror=format=2', '-Werror=missing-braces', '-Werror=missing-include-dirs', '-Werror=uninitialized', '-Werror=undef', '-Werror=pointer-arith', '-Werror=cast-qual', '-Werror=missing-declarations', '-Werror=vla', '-pthread', '-funsigned-char']
DXX-Archive.3: LIBS: None
DXX-Archive.3: LINKFLAGS: []
DXX-Archive.3: $CCACHE_PREFIX: None
DXX-Archive.3: $DISTCC_HOSTS: None
DXX-Archive.3: checking whether C++ compiler and linker work...yes
[...]

PR #720, opened 2023-06-25
Skip SDL OpenGL build check when the SDL2 wrapper is detected

chewi:
The wrapper does not define SDL_VIDEO_OPENGL in SDL_config.h. This appears to be deliberate, although it's not entirely clear why.

    /* Don't define most of the SDL backends, under the assumption checking for these against the headers won't work anyhow.
       The exception is the X11 backend; you need its define to know if you can use its syswm interface. */

We could check SDL2's SDL_config.h instead, but that seems awkward to pull off.

Closes: https://github.com/dxx-rebirth/dxx-rebirth/issues/689


vLKp:
It's unfortunate that they chose to do it that way, but since they made that choice, merging your change to bypass the check seems like the least bad option.  Merged.

PR #734, opened 2023-10-02
Fix macOS builds with current clang version

Kreeblah:
The current version of clang on macOS yields this error when compiling D2X:

```
CXX d2x build/ similar/main/wall.cpp
similar/main/wall.cpp:1429:3: error: ignoring return value of function declared with 'nodiscard' attribute [-Werror,-Wunused-result]
        std::remove_if(r.begin(), r.end(), std::ref(rp));
        ^~~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
scons: *** [build/similar/main/.d2x-rebirth.wall.o] Error 1
```

So, this explicitly discards the return value from that `std::remove_if` call there.  Applying it results in a clean build.

### DIFF
--- a/similar/main/wall.cpp
+++ b/similar/main/wall.cpp
@@ -1426,7 +1426,7 @@ void wall_frame_process(const d_robot_info_array &Robot_info)
 		const auto &&r = partial_range(CloakingWalls, CloakingWalls.get_count());
 		auto &Walls = LevelUniqueWallSubsystemState.Walls;
 		cw_removal_predicate rp{Segments.vmptr, Walls};
-		std::remove_if(r.begin(), r.end(), std::ref(rp));
+		static_cast<void>(std::remove_if(r.begin(), r.end(), std::ref(rp)));
 		CloakingWalls.set_count(rp.num_cloaking_walls);
 	}
 #endif


vLKp:
I agree that this should be changed.  For future reference, what version of clang is this?  With the standard build options I use, clang is not presently warning about this here, with either clang-15 or clang-16.

---

For those looking only at the error message, and not at the surrounding context, I will note that while ignoring the return value from `remove_if` is often wrong (because discarding the return value deprives the caller of the new end iterator), in this case ignoring the return value is safe.  `rp` is an object that counts how many elements were not removed, and a later line uses that count to update the end iterator of the range `r`.  This is done so that the caller does not need to compute the count of elements through use of `std::distance(r.begin(), returned_iterator)`.

Kreeblah:
I suspect this is Apple pulling in upstream patches (or mismatching libcxx versions) again.  Here's what I get from `clang --version`:

```
Apple clang version 15.0.0 (clang-1500.0.40.1)
Target: arm64-apple-darwin23.0.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

And, if I check  the definition in `remove_if.h` I have is:

```
//===----------------------------------------------------------------------===//
//
// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
// See https://llvm.org/LICENSE.txt for license information.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
//
//===----------------------------------------------------------------------===//

#ifndef _LIBCPP___ALGORITHM_REMOVE_IF_H
#define _LIBCPP___ALGORITHM_REMOVE_IF_H

#include <__algorithm/find_if.h>
#include <__config>
#include <__utility/move.h>

#if !defined(_LIBCPP_HAS_NO_PRAGMA_SYSTEM_HEADER)
#  pragma GCC system_header
#endif

_LIBCPP_BEGIN_NAMESPACE_STD

template <class _ForwardIterator, class _Predicate>
_LIBCPP_NODISCARD_EXT _LIBCPP_HIDE_FROM_ABI _LIBCPP_CONSTEXPR_SINCE_CXX20 _ForwardIterator
remove_if(_ForwardIterator __first, _ForwardIterator __last, _Predicate __pred)
{
    __first = _VSTD::find_if<_ForwardIterator, _Predicate&>(__first, __last, __pred);
    if (__first != __last)
    {
        _ForwardIterator __i = __first;
        while (++__i != __last)
        {
            if (!__pred(*__i))
            {
                *__first = _VSTD::move(*__i);
                ++__first;
            }
        }
    }
    return __first;
}

_LIBCPP_END_NAMESPACE_STD

#endif // _LIBCPP___ALGORITHM_REMOVE_IF_H
```

I'm guessing that `_LIBCPP_NODISCARD_EXT` there is what's triggering this.  Checking libcxx 17.0.1 upstream, I also see it there, so I suspect building would also trigger this warning with that libcxx version.

Kreeblah:
Actually, I see that marker in libcxx 16.0.6 as well, so I guess that's not definitive for determining a version.  I'll need to do some more checking.

Kreeblah:
OK, I found https://reviews.llvm.org/D128267 which enabled `[[nodiscard]]` at all times, and it looks like that went out in libcxx 16.0.0: https://releases.llvm.org/16.0.0/projects/libcxx/docs/ReleaseNotes.html

I'm not sure why you're not seeing the same warning there, though.

vLKp:
Thank you for the thorough investigation.  I agree that `[[nodiscard]]` is the origin of this warning.  It looks like my install of `clang` is using gcc's libstdc++ headers, not clang's libc++ headers, hence the different implementation.  That was not my intent.

Regardless, a happy side discovery of your investigation is that Apple is shipping a newer libc++ than I had thought, and may now implement more of the C++20 features I wanted, like Ranges.  I have been holding off on publishing much C++20 specific work because Apple's clang was behind when last I looked.

PR #735, opened 2023-10-04
Merge 5c900ca7a stereo updates

IronicResearch:
Updated with stereo viewport support for intro screens, popup menus, automap windows, and D2 extra viewports for guidebot/rearview displays when stereo viewport mode is enabled.

Extra stereo viewport handling allows for more uniform experience when stereo display mode is active, whereas original Descent 1/2 switched in and out stereo modes between game play and higher-res graphics for menus and automaps. 

Conditionals for DXX_USE_STEREOSCOPIC_RENDER=1 added for all stereo support steps dependent on VR_stereo variables. Negligible impact when compiling default non-stereo version with DXX_USE_STEREOSCOPIC_RENDER=0.

Added conditional compilation for DXX_USE_OGLES=1 to omit glDrawBuffer() calls used for GL_STEREO quad buffering, so stereo viewport options also supportable with GLES. Need GLES alternative to gCopyPixels() for viewport copy ops.

Stereo viewport handling needs to adjust x,y / w,h coordinates in each of these special cases. Fullscreen bitmaps were easily handled by blitting bitmaps twice via overloaded ogl_ubitmapm_cs() function mainly used by show_fullscr().

Popup menus & dialogs were trickier since text strings needed to be matched with background graphics. Right-eye viewport is simply copied from rendered left-eye viewport (default OGL). Would have preferred to render popup menu text & graphics composited to offscreen buffer and blitting to left/right viewports similar to fullscreen case, which was unexpectedly problematic for default OGL (non-GLES).

### DIFF
--- a/common/2d/bitblt.cpp
+++ b/common/2d/bitblt.cpp
@@ -32,6 +32,7 @@ COPYRIGHT 1993-1999 PARALLAX SOFTWARE CORPORATION.  ALL RIGHTS RESERVED.
 #include "dxxerror.h"
 #include "byteutil.h"
 #if DXX_USE_OGL
+#include "fwd-gr.h"
 #include "ogl_init.h"
 #endif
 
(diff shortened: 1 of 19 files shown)


vLKp:
Welcome back.  This pull request is rather large, and may take some time to review.  I will try to get to it as soon as I can.

IronicResearch:
OK by me, since I had to catch up on changes in various stage merges.
Most impact would be in automap.cpp & newmenu.cpp modules which could potentially be revisited if offscreen bitmaps were implementable for composting text strings over background graphics.

PR #746, opened 2024-01-03
Web port

midzer:
Hi,

because there are no `Issues` enabled in this repo, I am opening a PR directly to show my changes.

The last days, I was working on a web port of dx1 and dx2.

Everything compiled pretty well soon, but there was flickering because of `SDL_Delay` calls in a loop. In the process, I optimized the main loop

* no manual `timer_delay()` calls any more in each class
* calc time for each frame and delay before swap

and event handling

* prefer `SDL_PollEvents` over `SDL_PeekEvents`
* Cleanup

As a result `::idle` event isn't much used any more and can be removed completely in the future.

Here's the working shareware of dx1 https://midzer.de/wasm/descent1/

Have a great week
midzer

### DIFF
--- a/EMSCRIPTEN.md
+++ b/EMSCRIPTEN.md
@@ -1,0 +1,29 @@
+# Emscripten
+
+## Build
+
+### Descent 1
+
+```
+emscons scons CXX=em++ CXXFLAGS="-sUSE_SDL=2 -Icommon/include/physfs/src -Icommon/include/gl4es/include" d1x=1 d2x=0 sharepath=""
+```
+
+### Descent 2
+
+```
+emscons scons CXX=em++ CXXFLAGS="-sUSE_SDL=2 -Icommon/include/physfs/src -Icommon/include/gl4es/include" d1x=0 d2x=1 sharepath=""
+```
+
+## Link
+
+### Descent 1
+
+```
+em++ -flto -O3 common/*/*.o common/*/*/*.o d1x-rebirth/main/*.o similar/*/*.o similar/*/*/*.o libphysfs.a libGL.a libGLU.a -o index.html -sUSE_SDL=2 -sUSE_SDL_IMAGE=2 -sSDL2_IMAGE_FORMATS='["pcx"]' -sUSE_SDL_MIXER=2 -sSDL2_MIXER_FORMATS='["mid"]' -sFULL_ES2 -lGL -sASYNCIFY -sENVIRONMENT=web --preload-file data/ --preload-file d1x.ini --preload-file dgguspat@/etc/timidity -sINITIAL_MEMORY=128mb -sSTACK_SIZE=524288 -Wl,-u,fileno -Wl,-u,htonl --closure 1
+```
+
+### Descent 2
+
+```
+em++ -flto -O3 common/*/*.o common/*/*/*.o d2x-rebirth/main/*.o d2x-rebirth/libmve/*.o similar/*/*.o similar/*/*/*.o libphysfs.a libGL.a libGLU.a -o index.html -sUSE_SDL=2 -sUSE_SDL_IMAGE=2 -sSDL2_IMAGE_FORMATS='["pcx"]' -sUSE_SDL_MIXER=2 -sSDL2_MIXER_FORMATS='["mid"]' -sFULL_ES2 -lGL -sASYNCIFY -sENVIRONMENT=web --preload-file data/ --preload-file d2x.ini --preload-file dgguspat@/etc/timidity -sINITIAL_MEMORY=128mb -sSTACK_SIZE=524288 -Wl,-u,fileno -Wl,-u,htonl --closure 1
+```
--- a/SConstruct
+++ b/SConstruct
@@ -971,26 +971,26 @@ help:assume C++ compiler works
 		#
 		# If they are not in use, this assignment is a no-op.
 		cenv['CXXCOM'] = cenv._dxx_cxxcom_no_prefix
-		if Link(context, text='', msg='whether C++ compiler and linker work', calling_function='ld_works'):
+		#if Link(context, text='', msg='whether C++ compiler and linker work', calling_function='ld_works'):
 			# If ccache or distcc are in use, this block is only reached
 			# when one or both of them failed.  `most_recent_error` will
 			# be a description of the failure.  If neither are in use,
 			# `most_recent_error` will be None.
-			return most_recent_error
+		#	return most_recent_error
 		# Force only compile, even if LTO is enabled.
-		elif self._Compile(context, text='', msg='whether C++ compiler works', calling_function='cxx_works'):
+		if self._Compile(context, text='', msg='whether C++ compiler works', calling_function='cxx_works'):
 			specified_LIBS = 'or ' if cenv.get('LIBS') else ''
 			if specified_LIBS:
 				cenv['LIBS'] = []
-				if Link(context, text='', msg='whether C++ compiler and linker work with blank $LIBS', calling_function='ld_blank_libs_works'):
+				#if Link(context, text='', msg='whether C++ compiler and linker work with blank $LIBS', calling_function='ld_blank_libs_works'):
 					# Using $LIBS="" allowed the test to succeed.  $LIBS
 					# specifies one or more unusable libraries.  Usually
 					# this is because it specifies a library which does
 					# not exist or is an incompatible architecture.
-					return 'C++ compiler works.  C++ linker works with blank $LIBS.  C++ linker does not work with specified $LIBS.'
+				#	return 'C++ compiler works.  C++ linker works with blank $LIBS.  C++ linker does not work with specified $LIBS.'
 			if cenv['LINKFLAGS']:
 				cenv['LINKFLAGS'] = []
-				if Link(context, text='', msg='whether C++ compiler and linker work with blank $LIBS and blank $LDFLAGS', calling_function='ld_blank_libs_ldflags_works'):
+				#if Link(context, text='', msg='whether C++ compiler and linker work with blank $LIBS and blank $LDFLAGS', calling_function='ld_blank_libs_ldflags_works'):
 					# Using LINKFLAGS="" allowed the test to succeed.
 					# To avoid bloat, there is no further test to see
 					# whether the link will work with user-specified
@@ -1000,8 +1000,8 @@ help:assume C++ compiler works
 					# different error on the next run.  If the user is
 					# lucky, fixing LINKFLAGS will allow the build to
 					# run.
-					return 'C++ compiler works.  C++ linker works with blank $LIBS and blank $LDFLAGS.  C++ linker does not work with blank $LIBS and specified $LDFLAGS.'
-			return f'C++ compiler works.  C++ linker does not work with specified {specified_LIBS}blank $LIBS and specified $LINKFLAGS.  C++ linker does not work with blank $LIBS and blank $LINKFLAGS.'
+				#	return 'C++ compiler works.  C++ linker works with blank $LIBS and blank $LDFLAGS.  C++ linker does not work with blank $LIBS and specified $LDFLAGS.'
+			#return f'C++ compiler works.  C++ linker does not work with specified {specified_LIBS}blank $LIBS and specified $LINKFLAGS.  C++ linker does not work with blank $LIBS and blank $LINKFLAGS.'
 		else:
 			if cenv['CXXFLAGS']:
 				cenv['CXXFLAGS'] = []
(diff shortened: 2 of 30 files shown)


vLKp:
> because there are no `Issues` enabled in this repo

How did you determine this?  This repository has always had the Issues feature enabled, and even has some very old wishlist issues still open.

> * prefer `SDL_PollEvents` over `SDL_PeekEvents`

There is no `SDL_PollEvents`.  Did you mean `SDL_PollEvent`?

There is no usage of `SDL_PeekEvents`.  Did you mean `SDL_PeepEvents`?

Could you explain how this is an improvement?  As I understand the SDL event mechanism, SDL_PeekEvents has the useful property that it can return more than one event at a time, saving on calls into the SDL event system.  A cursory read of the SDL code seems to agree with this, and confirms that SDL_PollEvent is implemented on top of SDL_WaitEventTimeout, which in turn will use the peep event logic.

---

As to the changes:

- `EMSCRIPTEN.md` shows several commands, but none of those are installed here, and it has no hints what readers should do to obtain those commands, nor explanations of why one would want to run those commands.  Top-level documentation files may be read by new users to see what is offered, and this file presently does not describe who should be reading it or what the reader should do to get started using it.
- `EMSCRIPTEN.md` seems to propose searching header paths that do not exist (`common/include/physfs/src`, `common/include/gl4es/include`).  This will likely trigger a warning from `-Wmissing-include-dirs`.
- `EMSCRIPTEN.md` link command looks completely wrong to me.  The default `builddir` will put object files outside the source tree, so you cannot find them using a glob in the source tree.  Linking in every file with the `.o` extension, regardless of whether it is provided by the current source code, is wrong.  The linker command should only include files that were actually generated by compialtion.  The default naming convention for files in the non-common code includes a leading dot for historical reasons, so your standard shell globbing cannot see those files.  Enabling bash's `dotglob` would match them, but the instructions do not mention that.  More generally, SCons already understands how to link object files into a final executable, so that should be used here instead of running a linker command manually.
- `EMSCRIPTEN.md` proposes linking to static libraries that do not exist, and has no instructions on how to obtain them.  Further, static libraries are generally discouraged, so if they are needed, there should be an explanation of why static linking is correct here.
- Your changes to SConstruct break some of the normal SConf logic.  This might be acceptable for a specialized build where you know all the tests are pointless, but it cannot be merged as-is into `master`.  Additionally, there are no inline comments nor commit message notes explaining why any of those changes are needed even for the Emscripten build.
- You changed the defaults for several SConf options, but there is no indication why.
- You removed a standard warning option (`-Wundef`).  Why?
- You removed default debug information (`-g`).  Why?
- You force-enabled `-flto`, which the default build supports but does not force.  Why?  If you need to enable LTO, you should enable it via the SConf switch `lto`, rather than setting the flag directly.
- You changed from `-O2` to `-O3`.  Generally, on desktop platforms, `-O3` is discouraged.  Why are you default-enabling this?
- You force-disabled inclusion of GL headers that, as far as I know, desktop platforms need.  I expect this will cause many build failures due to missing declarations.
- Why did you disable the compile-time selftest of `check_bswap` on `unsigned long long`?
- Why did you remove the sanity checks in the valptridx `requires` clauses?
- You added in `inferno.cpp` an include of a header I do not have, and do not see instructions on how to obtain.  What package provides `gl4esinit.h`?

Centralizing delay logic is a good change, provided that it does not cause the desktop platforms to busy-loop in places where it previously slept between frames.

Since all these changes are combined in a single commit, and several of them break the desktop build, I cannot test any of your changes without reverting the breaking ones.  This will impede my ability to try out the proposed changes.

midzer:
First of all, thanks for your long reply.

Somehow (?) I overlooked Issues ARE in fact enabled in this repo (you can disable them optionally in repo settings, there are repos around which do this). Sorry, for the irritation. My fault. Before showing any code (and making a PR) I wanted to talk about the possibility of porting `dxx-rebirth` to a browser.

It was *not* my intent to have this whole thing merged, just talk about it and find changes which are possible good for the project.

Let me go into your questions in a following post.

midzer:
> There is no SDL_PollEvents. Did you mean SDL_PollEvent?

Yes, sorry for misnaming, typed it out of my brain.

> There is no usage of SDL_PeekEvents. Did you mean SDL_PeepEvents?

Yes, same as previous answer.

>Could you explain how this is an improvement? As I understand the SDL event mechanism, SDL_PeekEvents has the useful property that it can return more than one event at a time, saving on calls into the SDL event system. A cursory read of the SDL code seems to agree with this, and confirms that SDL_PollEvent is implemented on top of SDL_WaitEventTimeout, which in turn will use the peep event logic.

After spending days in the code for debugging, I (tried to) optimize `common/arch/sdl/event.cpp`, mainly because of the statement

> SDL_PollEvent() is the favored way of receiving system events since it can be done from the main loop and does not suspend the main loop while waiting on an event to be posted.

in https://wiki.libsdl.org/SDL2/SDL_PollEvent

As you had a look in SDL source and figured out that this function is a on top of `SDL_PeekEvent`, all changes in `common/arch/sdl/event.cpp` are no benefit, just less lines of code, not supporting SDL1, too.

Answering further questions in next post.

midzer:
> EMSCRIPTEN.md shows several commands, but none of those are installed here, and it has no hints what readers should do to obtain those commands, nor explanations of why one would want to run those commands. Top-level documentation files may be read by new users to see what is offered, and this file presently does not describe who should be reading it or what the reader should do to get started using it.

True, there has to be an explanation or at least a link to https://emscripten.org/docs/getting_started/downloads.html

> EMSCRIPTEN.md seems to propose searching header paths that do not exist (common/include/physfs/src, common/include/gl4es/include). This will likely trigger a warning from -Wmissing-include-dirs.

Yes, those headers are not commited (right now).

> EMSCRIPTEN.md link command looks completely wrong to me. The default builddir will put object files outside the source tree, so you cannot find them using a glob in the source tree. Linking in every file with the .o extension, regardless of whether it is provided by the current source code, is wrong. The linker command should only include files that were actually generated by compialtion. The default naming convention for files in the non-common code includes a leading dot for historical reasons, so your standard shell globbing cannot see those files. Enabling bash's dotglob would match them, but the instructions do not mention that. More generally, SCons already understands how to link object files into a final executable, so that should be used here instead of running a linker command manually.

I am not so familiar with SCons yet. Yes, I had to remove the `.` from those files via a bash command, each time after compiling.

> EMSCRIPTEN.md proposes linking to static libraries that do not exist, and has no instructions on how to obtain them. Further, static libraries are generally discouraged, so if they are needed, there should be an explanation of why static linking is correct here.

Yes, at least links to the repos of those dependencies are necessary. https://emscripten.org/docs/compiling/Dynamic-Linking.html has insight why static linking is preferred right now for WebAssembly.

> Your changes to SConstruct break some of the normal SConf logic. This might be acceptable for a specialized build where you know all the tests are pointless, but it cannot be merged as-is into master. Additionally, there are no inline comments nor commit message notes explaining why any of those changes are needed even for the Emscripten build.

Yes, those `SConstruct` changes break things right now. As mentioned earlier, an issue would've been better before opening this rather large PR.

> You changed the defaults for several SConf options, but there is no indication why.

Later I realized I can pass those arguments to `scons` command as well, like `sdl2=1` for example.

> You removed a standard warning option (-Wundef). Why?

There was an undefined identifier in an #if directive which lead to an error while compiling. IIRC, it was related to `gl4es` GL headers, not sure.

> You removed default debug information (-g). Why?

To have a final "full optimized build". Is the debug information fully stripped out, when `-g` is *not* in the linker step?

> You force-enabled -flto, which the default build supports but does not force. Why? If you need to enable LTO, you should enable it via the SConf switch lto, rather than setting the flag directly.

IIRC, I tried via `SConf`, but it did not work.

> You changed from -O2 to -O3. Generally, on desktop platforms, -O3 is discouraged. Why are you default-enabling this?

I sticked to `Building with -O3 or -Os can produce an ever better build than -O2, and are worth considering for release builds.` via https://emscripten.org/docs/optimizing/Optimizing-Code.html

> You force-disabled inclusion of GL headers that, as far as I know, desktop platforms need. I expect this will cause many build failures due to missing declarations.

Yes, as mentioned earlier, those changes - as they are right now, break things for other systems.

> Why did you disable the compile-time selftest of check_bswap on unsigned long long?

I think, those tests made `emscons scons` fail. Have to retry to investigate.

> Why did you remove the sanity checks in the valptridx requires clauses?

```
requires(
			allow_nullptr /*|| !rhs.allow_nullptr*/
```

this looks like a merge mistake after I merged latest `master` changes from late December in my `emscripten` branch.

> You added in inferno.cpp an include of a header I do not have, and do not see instructions on how to obtain. What package provides gl4esinit.h?

Yeah, it is coming from https://github.com/ptitSeb/gl4es, not commited, again.

midzer:
> Centralizing delay logic is a good change, provided that it does not cause the desktop platforms to busy-loop in places where it previously slept between frames.

Yes, those few changes for passed variable `frameStart` should be tested individually for native platforms.

> Since all these changes are combined in a single commit, and several of them break the desktop build, I cannot test any of your changes without reverting the breaking ones. This will impede my ability to try out the proposed changes.

If you agree, I could pick only those changes regarding manual delays and `frameStart` passing in a separate branch. It should not break.

vLKp:
> I am not so familiar with SCons yet. Yes, I had to remove the `.` from those files via a bash command, each time after compiling.

If you are going to hack up SCons anyway, you could instead remove the `.` from where I tell SCons how to name the file. ;)  The below diff seems to achieve that:

```diff
diff --git a/SConstruct b/SConstruct
index a39c8cf16..80cdd2715 100644
--- a/SConstruct
+++ b/SConstruct
@@ -4923,7 +4923,7 @@ class DXXProgram(DXXCommon):
 	static_archive_construction = {}
 
 	def _apply_target_name(self,name):
-		return os.path.join(os.path.dirname(name), f'.{self.target}.{os.path.splitext(os.path.basename(name))[0]}')
+		return os.path.join(os.path.dirname(name), f'{self.target}.{os.path.splitext(os.path.basename(name))[0]}')
 
 	def _apply_env_version_seq(self,env,_empty={}):
 		if self.user_settings.pch:
```

I have considered applying this a few times over the years, but held back because it means that building a commit from before the change and building a commit from after it will use different names, and the old files would be left as orphans.

> As mentioned earlier, an issue would've been better before opening this rather large PR.

Understood.  Your PR is actually fairly small compared to another one that was opened recently without any warning.  That one is rather difficult to review in the Github web interface.  Don't worry about having started here.

> Later I realized I can pass those arguments to `scons` command as well, like `sdl2=1` for example.

True, and for options you keep using repeatedly, the `site` facility may be an even better choice.  I keep my preferred build settings there, so that I do not need to retype them on every run.  For your use case, running `echo sdl2=1 > site-local.py` would likely suffice.  The file is named `site-local.py` (but can be overridden with `scons` the command line variable assignment `site=my-preferences.py` (see the top of `main` in `SConstruct`)).  It is evaluated as full Python (so you can open files readable by your user - be careful with untrusted site scripts), though without access to the SCons active state.  Global variables set by the evaluation of the site script override the built-in defaults (but not command line assignments).

> There was an undefined identifier in an #if directive which lead to an error while compiling. IIRC, it was related to `gl4es` GL headers, not sure.

If you include the gl4es headers via `-isystem` instead of as application headers, the compiler should be less strict about warnings in them.  Personally, I would prefer to see such a warning fixed, but I recognize that means getting the gl4es maintainers to make a change.

> To have a final "full optimized build". Is the debug information fully stripped out, when `-g` is *not* in the linker step?

That depends on the toolchain, but generally, no, you would need either an explicit `strip` or to tell the linker to strip symbols (often via `-s` as a linker option).  Given how unusual Emscripten is compared to normal toolchains, it may behave differently in this regard, so if you are concerned about eliminating unnecessary debugging information, you should check the Emscripten documentation.

Also note that the presence or absence of debug information only affects the size of the output, but not its performance.  I routinely build with `-g -O2` and get an optimized build with debug symbols.  (Debugging experience can be impaired by how the optimizer rearranged code, but I can at least count on having function names and moderately accurate breakpoint-by-source-line.)

> IIRC, I tried via `SConf`, but it did not work.

LTO works for me on a traditional toolchain.  If it does not work for you, please file an issue reporting the problem.  I might be able to make a code change to get it to work for you.  I would expect that force-enabling it the way you did actually works worse, because some SConf tests need special handling when LTO is enabled, and your method of forcing it prevents them from noticing that they need to enable that handling.

> I sticked to `Building with -O3 or -Os can produce an ever better build than -O2, and are worth considering for release builds.` via https://emscripten.org/docs/optimizing/Optimizing-Code.html

I find their claim a bit questionable, but if they recommend it for this platform, I will assume they have tested it and found it suitable.  On traditional desktop systems, `-O3` has a reputation for increasing code size through excessive unrolling, and for breaking programs that are not perfectly standards-conformant.  I would like to believe I have not done anything in the Rebirth code that would cause a problem on that latter point, but I cannot say that I test `-O3` with any regularity, nor receive reports from anyone else who does.  Use it if you like, but beware that you may encounter problems that would be hidden under `-O2`.  If you do see problems, please report them.  They are likely worth fixing in the main tree.

> this looks like a merge mistake after I merged latest `master` changes from late December in my `emscripten` branch.

Some of those changes were subsequently rewritten in `master`, because `clang` cannot handle accessing `static constexpr` members through a reference received as an argument.  (`gcc` handles this without issue.)  Perhaps you pulled during the window between when I published the checks in that form and when I rewrote them to a less desirable form that `clang` understands.

> If you agree, I could pick only those changes regarding manual delays and `frameStart` passing in a separate branch. It should not break.

I would like to try it out before I commit to merging it, but yes, if you want to extract it out, I can take a look at it.  Making the game sleep more and poll less is generally desirable, in my opinion.

midzer:
Closing this one, but I keep my branch active for further development of the web port.

PR #748, opened 2024-01-06
Try a github actions package.yml

JeodC:
Can merge into a test branch if it makes you feel safer.

- Removes unnecessary .yml files
- Adds one package.yml file to build for Ubuntu 20.04, Ubuntu 22.04, MacOS, and Windows

Post-merge:
- Remove upstream-merge.yml as this is the upstream repo (not needed)
- Change the following in package.yml:
`
on:
  push:
    branches: [ package ]`
to
`
on:
  push:
    branches: [ master ]`

Thanks to maggu 2810 for original design. I only consolidated packaging.


vLKp:
The sheer number of commits here reinforces my suspicion that Github Actions are not ready for general use.  Needing to commit and publish to the world just to test a change seems wrong.  Is there no way to prototype these offline?

> Adds one package.yml file to build for Ubuntu 20.04, Ubuntu 22.04, MacOS, and Windows

What is the advantage of using one file instead of individual targets?  I would expect that keeping them broken out makes the logs cleaner to read.

> I only consolidated packaging.

It looks like you did a lot more than that.  The diffstat says `8 files changed, 461 insertions(+), 163 deletions(-)`.

---

I see several changes here that seem wrong.  There is no need to fix the build for gcc-10, because gcc-10 is no longer supported.  Similarly, there is no need to build with gcc-11, because gcc-11 is no longer supported.  I can sort this out when I squash the merge.

---

This seems to be long on changes and short on inline comments.  Is any of this modelled off a documented reference project I can read about to understand the design better?  I am particularly curious why a separate `Dockerfile` is added.  I thought Github Actions ran everything in a docker container already, and there would be no need to have another container inside that container.

JeodC:
All of these commits are me learning how actions work. I took @maggu2810's files from his fork and set about modifying them to reshape them to what I envisioned. The Dockerfile and GCC-10 commits are from before I moved those changes into a separate compatibility branch. I reverted those changes after I created a branch for them. The compatibility branch was needed to keep the application working on ArkOS. That branch has a modified readme.md to explain:

`This branch contains compatibility fixes for retro handheld devices running on RK3326 and RK3566 processors. Some of these devices use operating systems that rely on older GLIBs and GCC compiler rules. This branch is manually maintained with new commits slowly synced and builds tested.`

Back to the primary branch:

The packages are combined into one .yml file for ease of use. You can view the separate build logs easily when you open the action:

image

You just click on the target system you want to look at.

image

You get a summary of the job:

image

...which is a series of dropdown menus.

If you don't want to create releases, you can always remove the upload job at the bottom of  `package.yml`.

JeodC (commented):
Changes:
- Package-linux.yml is removed
- Package-macos.yml is removed
- Package-windows.yml is removed / renamed to package-sdl1.2.yml and modified
   - This was only used for testing sound issues in a separate discussion and is not really needed
- Package.yml was created
- Upstream-merge.yml was created but is not needed since this is the upstream repo it references
- linux/build_package.sh was modified to consolidate build files into one .zip file
- macos/build_package.sh was modified to consolidate build files into one .zip file
-  windows/build_package.sh was modified to consolidate build files into one .zip file

Output:
- DXX-Rebirth-ubuntu-20.04.zip
   - D1X-Rebirth.AppImage
   - D2X-Rebirth.AppImage
- DXX-Rebirth-ubuntu-22.04.zip
   - D1X-Rebirth.AppImage
   - D2X-Rebirth.AppImage
- DXX-Rebirth-macos.zip
   - D1X-Rebirth.app
   - D2X-Rebirth.app
- DXX-Rebirth-windows.zip
   - D1X-Rebirth/
   - D2X-Rebirth/

JeodC:
You know what, I will recreate the fork from scratch and add these files in so it's easier to digest. Sorry it's a mess.

vLKp:
> All of these commits are me learning how actions work.

I understand that.  My point was that Github really ought to provide a way for new users of the Github Actions feature to learn it more cleanly.  I've never had a working Github Action, so I too will need to learn how they work.  Seeing how many tries it took you makes me think this will be a major undertaking for me to get one working to my satisfaction.  I would be happier if I could run it in a container locally, where I can fully inspect the state at every step of the debugging process.

> The compatibility branch was needed to keep the application working on ArkOS.

I suggest getting ArkOS to update would be a better use of your time than trying to maintain a fork that runs on older gcc.  You may be able to keep the code working for now with just disabling build checks that require newer gcc, but future changes may not be so easy to handle.  In particular, changes to require more features from the C++20 standard library will need to be skipped or rewritten in your fork.

> You can view the separate build logs easily when you open the action:

Perhaps this is only because it is your repository and not mine, but I don't see a way to get the logs out of any of this.  I think I've seen other actions provide their logs before, but I don't know why yours would be absent.

> You know what, I will recreate the fork from scratch and add these files in so it's easier to digest. Sorry it's a mess.

That wasn't necessary.  You could have force-pushed over the branch with a new, cleaner history.

We can continue this in the new pull request #751 though.

PR #750, opened 2024-01-09
Replace manual delays by frame time measurement

midzer:
Hi,

According to good dialogue in https://github.com/dxx-rebirth/dxx-rebirth/pull/746, here's mentioned PR to improve render performance (should give more FPS) by replacing all manual delays.

Code is tested for Descent 1 Shareware.

Have a nice week
midzer

### DIFF
--- a/common/arch/sdl/timer.cpp
+++ b/common/arch/sdl/timer.cpp
@@ -48,30 +48,4 @@ void timer_delay_ms(unsigned milliseconds)
 	SDL_Delay(milliseconds);
 }
 
-// Replacement for timer_delay which considers calc time the program needs between frames (not reentrant)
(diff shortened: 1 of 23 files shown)


vLKp:
From an initial read of this work, it looks like you removed delay calls in a variety of places, but only added the ability to sleep in the OpenGL-specific rendering path.  Will that not cause the SDL-only build to spin, since it does not use the OpenGL renderer?  It looks like SDL-only may not even build, since you changed the declaration for `gr_flip`, but did not update the implementation in `similar/arch/sdl/gr.cpp` to match the new signature.

midzer:
Yeah, I forgot the SDL-only path.

Gonna try if I get it to work and push necessary changes.

PR #751, opened 2024-01-09
Add github actions to facilitate releases

JeodC:
Hopefully this is easier to digest than #748. View the action at [this fork](https://github.com/JeodC/dxx-rebirth/actions/runs/7466973182).

Original design by @maggu2810, further modifications by JeodC.

Changes:

- Package.yml was created
- Upstream-merge.yml was created but is not needed since this is the upstream repo it references
- linux/build_package.sh was modified to consolidate build files into one .zip file
- macos/build_package.sh was modified to consolidate build files into one .zip file
- windows/build_package.sh was modified to consolidate build files into one .zip file

Release output:

- DXX-Rebirth-ubuntu-20.04.zip
  - D1X-Rebirth.AppImage
  - D2X-Rebirth.AppImage
- DXX-Rebirth-ubuntu-22.04.zip
  - D1X-Rebirth.AppImage
  - D2X-Rebirth.AppImage
- DXX-Rebirth-macos.zip
  - D1X-Rebirth.app
  - D2X-Rebirth.app
- DXX-Rebirth-windows.zip
  - D1X-Rebirth/
  - D2X-Rebirth/

Potential tasks:

- If you would like, I can reshape this so that releases must be manually run. I would accomplish this by renaming `packages.yml` to `release.yml` and disabling the automated run every push. I would then create a `build.yml` file that *does* run on commit pushes, but does not upload artifacts or releases. This would enable you to test if new commits build successfully. However, you may already do this before pushing commits to the repo for all I know.
- I can add comments throughout the `package.yml` file to further clarify what is happening.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -464,6 +464,33 @@ class ConfigureTests(_ConfigureTests):
 	__python_import_struct = None
 	_cxx_conformance_cxx20 = 20
 	__cxx_std_required_features = CxxRequiredFeatures([
+		# As of this writing, <gcc-12 is already unsupported, but some
+		# platforms, such as Ubuntu 22.04, still try to use gcc-11 by default.
+		# Use this test both to verify that Class Template Argument Deduction
+		# is generally supported, since it is used in the game, and to exclude
+		# an unsupported compiler that does not accept CTAD in one of the
+		# contexts the game uses.
+		Cxx20RequiredFeature('class template argument deduction', '''
+struct Outer_%(N)s
+{
+	template <typename byte_type>
+	struct Inner
+	{
+		constexpr Inner(byte_type &)
+		{
+		}
+	};
+	/* <gcc-12 does not allow declaring a deduction guide in the immediately
+	 * enclosing scope:
+```
+error: deduction guide 'Outer_test_class_20template_20argument
+_20deduction::Inner(byte_type&) -> Outer_test_class_20template_20argument_20deduction::Inner<byte_type>' must be declared at namespace scope
+```
+	 */
+	template <typename byte_type>
+		Inner(byte_type &) -> Inner<byte_type>;
+};
+'''),
 		Cxx20RequiredFeature('explicitly defaulted operator==', '''
 struct A_%(N)s
 {


JeodC:
I'm not sure why you wouldn't be able to view a build log. The releases page provides it so it should be public. I don't think there's an option for privacy.

https://github.com/JeodC/dxx-rebirth/actions/runs/7467777209

You should be able to click around and see things from there.

vLKp:
> If you would like, I can reshape this so that releases must be manually run.

I appreciate this offer, but based on the utter failure of Github Actions to date, I think requiring a manual run would be a bad idea.  Every time I've ever tried to trigger a manual run, I get a blank pop up with a spinner that spins forever, no error messages, and the run never happens.  Since there are no errors, there is nothing I can do to fix it.  Since it never runs, no artifacts are ever built.

> This would enable you to test if new commits build successfully. However, you may already do this before pushing commits to the repo for all I know.

Historically, I was very picky about build testing every commit in a variety of supported environments before pushing them.  Any time I pushed a commit that was broken was an error on my part.  For days with a large push, this could require significant build time before I was ready to push, with additional delays if I found anything that required a rework.  More recently, with the revelation that most users just wait for someone to post a Windows binary for them, I have reconsidered whether this substantial use of CPU time is worth my time.  I have been seriously considering adopting a policy that I only build test `HEAD` before pushing it, and do not worry whether commits between the previous `master` and the new `HEAD` are all functional.

> I can add comments throughout the `package.yml` file to further clarify what is happening.

Thank you, but wait on that for now.  I can follow the shell scripting parts pretty well.  My concern has always been the meaning of the magical yml entries, particularly with strange constructs like:

```
          OUTPUT: ${{needs.globalenv.outputs.d1x_rebirth_appimage_output_u2004}}
```

I am still sour on the magic of `on: workflow_dispatch` to mean that the user is expected to click a button to trigger the workflow.  Reading the yml for that does not even hint to me that this was the intended meaning of that construct.

JeodC:
In GitHub Actions, jobs in a workflow run in isolated environments. Each job doesn't share its environment with other jobs, and information cannot be directly passed between them. To overcome this limitation, you can use output variables.

In the workflow:

The `globalenv` job sets various environment variables and performs some initial setup.

The `appimage` step within the `globalenv` job generates AppImage artifacts and defines output variables for the paths or names of these artifacts.

The `ubuntu2004` and `ubuntu2204` jobs need information about where the AppImage artifacts are located so they can package and upload them. This information is obtained by referencing the output variables defined in the `appimage` step.

Here's why the output variables are needed:

**Isolation between Jobs**: Each job runs in its own environment, and there is no direct sharing of information between jobs.

**Passing Information**: Output variables act as a mechanism to pass information (in this case, the paths or names of the generated AppImage artifacts) from one job to subsequent jobs.

**Consistency and Automation**: By using output variables, you can ensure consistency in the workflow, especially when dealing with multiple jobs and platforms. The values of these variables are determined dynamically during the workflow run and are used in subsequent steps to automate the packaging and uploading of artifacts.

For example, if the paths or names of the generated AppImage artifacts were hard-coded or manually specified in subsequent jobs, any changes to the generation process or filenames would require manual updates in multiple places. By using output variables, the workflow remains flexible and adapts to changes made in earlier steps, promoting maintainability and automation.

Why don't the ubuntu jobs just use their own appimage artifacts?

Each job could, in theory, handle its own artifacts without relying on output variables from a previous job. The use of output variables and cross-job communication is a design choice, and workflows can be structured in various ways based on project requirements and preferences.

Here are some reasons why the workflow might be designed this way:

**Modularization and Reusability**: The globalenv job may be responsible for common setup and configurations that are reused across multiple jobs. By centralizing this setup in one job, it can be easily maintained and modified without duplicating code in every job.

**Consistency**: Using output variables ensures that all jobs have access to the same information about the artifacts. This can help maintain consistency in the workflow and reduce the likelihood of errors due to discrepancies in file paths or names.

**Ease of Maintenance**: If there are changes to the artifact generation process or if additional platforms are added in the future, modifying a single job (globalenv) to handle these changes can be more convenient than updating multiple jobs individually.

**Clear Separation of Concerns**: Jobs in a workflow can be designed with specific responsibilities. The `appimage` step within the `globalenv` job might focus solely on generating artifacts, while subsequent jobs (`ubuntu2004` and `ubuntu2204`) focus on using those artifacts for packaging and uploading.

**Workflow Dependencies**: Output variables provide a way to express dependencies between jobs explicitly. By using the `needs` keyword and output variables, you can specify the order in which jobs should run and ensure that certain jobs complete successfully before others start.

If this reads like a ChatGPT response that's because it is. It's very useful in learning programming. In short, ask @maggu2810. He designed the yml. Maybe he took inspiration from other projects.

JeodC:
As for workflow_dispatch, I found documentation of that here: https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow

JeodC:
I wonder if your issues with Actions stem from the repo being for an organization rather than an individual.

https://docs.github.com/en/organizations/managing-organization-settings/disabling-or-limiting-github-actions-for-your-organization

vLKp:
> If this reads like a ChatGPT response that's because it is. It's very useful in learning programming.

From what I have read, it has an unfortunate tendency to invent plausible sounding garbage that will mislead the unwary.  For this reason, I stick to conventional search tools, which lead you to specific pages whose contents can be evaluated on its merits.  LLMs are unable to show their work, so it is very difficult to evaluate whether they are producing correct output.

---

> I wonder if your issues with Actions stem from the repo being for an organization rather than an individual.

Maybe?  It looks to me like the [settings](https://github.com/dxx-rebirth/dxx-rebirth/settings/actions) permit Github Actions freely right now.

---

I merged the commits in this pull request, with the exception of the two that turned sdlmixer off and back on.

JeodC:
The action worked, but you need to grant some permissions in order for the release job to function.

JeodC:
@maggu2810 This partion pull breaks your upstream-merge.yml. If you want to maintain it, you need to manually sync your fork and recreate it.

vLKp:
I found that I can get logs, albeit through a decidedly non-obvious path.  If I go to a specific phase of the job, then choose (Settings icon) -> View raw logs, I can get the output.  I cannot see why this is considered a setting, but at least logs are available.

---

> The action worked, but you need to grant some permissions in order for the release job to function. I'm not sure where exactly you find that in the Oragnization scope, but here it is in my fork's settings under Actions>General:

The organization has jobs set to be read-only by default.  I _think_ from some of the Github Actions documentation (which is still a chore to read) that the right additional stanza in the yml can opt in to more permissions.  This seems like a weird design, but I plan to experiment with it when I have time to fight with the Actions system.

JeodC:
That's easy You'll add two lines to the upload job at the bottom of package.yml.
```yml
 upload:
    #if: ${{ false }}
    name: Create release and upload artifacts
    needs:
      - ubuntu2004
      - globalenv
      - ubuntu2204
      - windows
      - macos
    runs-on: ubuntu-22.04
    permissions:
     contents: write
    steps:
      - name: Install AppImage requirements
        run: |
          sudo apt-get install -y libfuse2

      - name: Install AppImage stuff
        run: |
          mkdir ~/Applications
          curl -s -L --output ~/Applications/pyuploadtool-x86_64.AppImage https://github.com/TheAssassin/pyuploadtool/releases/download/continuous/pyuploadtool-x86_64.AppImage
          chmod 0755 ~/Applications/pyuploadtool-x86_64.AppImage

      - name: Download artifacts
        uses: actions/download-artifact@v3

      - name: Create release and upload artifacts
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          GITHUB_CONTINUOUS_RELEASE_TAG: ${{needs.globalenv.outputs.github_continuous_release_tag}}
          GITHUB_CONTINUOUS_RELEASE_TYPE: ${{needs.globalenv.outputs.github_continuous_release_type}}
          GITHUB_CONTINUOUS_RELEASE_NAME: ${{needs.globalenv.outputs.github_continuous_release_name}}
        run: |
          ~/Applications/pyuploadtool-x86_64.AppImage \
            linux-appimage-ubuntu-*/*.zip \
            linux-appimage-ubuntu-*/*.zip \
            windows/*.zip \
            macos/*.zip
```

JeodC:
Hey @vLKp you ok? Been a while.

vLKp:
Historically, I did most of my work on Rebirth on the weekends.  Lately, even that time has been scarce, and researching how to make Github Actions behave as I want felt like a chore that I would rather avoid, so my limited development time was instead spent on other tasks.  I finally got around to implementing what should have been a fix for this, and it instead broke the workflow even worse.  Github Actions really needs a proper offline validator for their syntax.  I thought from the documentation that the permissions could be applied at the step level, but does not seem to be true.  I pushed a fix for my mistake, but since the workflow ignores files under `.github/`, no replacement build was triggered.  I will need to find some other change to push to provoke Github Actions to try again.

JeodC:
Just nudge the readme?

vLKp:
There are some source code comments I could update for a forced change, though the activity in #759 seems relevant too.

---

I see 3 commits in this pull request that are not yet in `master`.

- https://github.com/dxx-rebirth/dxx-rebirth/pull/751/commits/b2a1fa3da817af58df4c81669a0bb276e6346001 - this is useful, but applies to a file that #759 completely replaces, so merging #759 would prevent merging this commit.
- https://github.com/dxx-rebirth/dxx-rebirth/pull/751/commits/47980d74d814a304fc31ae3afef81bf0a98e1b5b - this seems unnecessary.  As I understand GitHub Actions, each `run` statement runs in its own shell, so a `run` that is just an `export` of a fixed value sets that, then exits, discarding the result of the `export`.  Such a statement is not useful, and should be removed rather than modified.  However, this also is in a file that #759 replaces, so merging #759 would conflict with this too.
- https://github.com/dxx-rebirth/dxx-rebirth/pull/751/commits/21a7a0a5b0aa596b1572fb7c3fa4693f60668d03 - this could be useful for downstream forks, but is not useful in the Rebirth repository directly.  If there is a way to configure it to run only in forks, and not in the Rebirth repository, I am fine merging it.  I suspect that merging it as-is will lead to strange failures, as I have not set up a Personal Access Token in Rebirth for it to use.  The GitHub Actions `if` construct might (or might not) be usable for the purpose of forcing this not to run when evaluated in the Rebirth repository.

JeodC:
If #759 is more up-to-date, then by all means use that. As for https://github.com/dxx-rebirth/dxx-rebirth/commit/21a7a0a5b0aa596b1572fb7c3fa4693f60668d03 - I've already stated that file is meant for forks and should be deleted upon merging to the upstream repository (this one).

Issue #753, opened 2024-01-12
Sound effects too loud


JeodC:
### Description

This branches from a discussion in #711.

Sound effects like explosions are much louder in recent DXX versions than in 0.58.1. A video demonstration is attached. No .dxa addons are used.

https://github.com/dxx-rebirth/dxx-rebirth/assets/47716344/c751f3ee-dcc7-4ec7-a333-97327f7a3177

It appears explosion sound effects may play at least twice, causing the loudness. Since this issue occurs after 0.58.1, it likely occurs in D2X as well.

Other reported oddities that may be linked: weapon hit sound effects, incorrect directional audio, and sound effects randomly cutting out in water levels.

AlumiuN:
This is (at least part of) the sound issue I mentioned here - https://github.com/dxx-rebirth/dxx-rebirth/issues/711#issuecomment-1880385509

A quick test confirms that it isn't because older Rebirth only plays explosion sounds on a single projectile, because it does still play it for each impact on lava or robots regardless of which projectile it is. I can't say I've heard incorrect directional audio; IIRC the position of sounds isn't recalculated after they start, which means the player turning or moving may cause the sound to effectively be in the wrong position later on, but I believe that's consistent with DOS Descent (although that could be something that's worth changing anyway).

JeodC:
I am unable to build prior versions on windows for testing easily. However, I did find an old issue that looked interesting. Maybe this is an uncaught regression from #88 ?

JeodC:
AlumiuN tested without sdl mixer, issue still occurs. Mixer is not the problem.

https://www.youtube.com/watch?v=ZDJ93tL2jrw&feature=youtu.be

AlumiuN:
Well, SDL_mixer itself was never going to be the problem, because 0.58.1 uses it too; what it does mean is that either the issue is outside of the SDL_mixer specific path, or it was implemented on both the SDL_mixer path and the non-SDL_mixer path. I haven't had any luck finding anything obvious in either case, but there are a lot of changes to look through and I don't know what would be specifically causing this in the first place.

JeodC:
Here is a second comparison, this time between DXX-Redux and current DXX-Rebirth.

https://github.com/dxx-rebirth/dxx-rebirth/assets/47716344/38fae8ac-7f23-4fb8-9bb8-6f41720abccc

AlumiuN:
[...]

PR #758, opened 2024-03-09
Restore older Rebirth perceived digi_mixer volume

alexstrout:
This attempts to address the perceived volume issues noted in #753.

As @vLKp noted, old pre- 20a0166c builds had an unintentional effect of double-scaling certain (most) sounds, but not others. This PR attempts to emulate this behavior by instead double-scaling overall Mix_Volume, instead of the Mix_Distance call on individual sounds.

It also reverts any of the individual volume intensity changes made, restoring the overall expected mix.
(particularly with -nosdlmixer)

An interesting side-effect of the old double-scaling is that it actually set the perceived sound volume - at least on Windows - closer to the perceived midi volume? I couldn't find any obvious reason for this in the Rebirth code itself, but my best guess is the system's midi volume scaling may be accounting for [logarithmic intensity of sound](https://www.britannica.com/science/sound-physics/The-decibel-scale) whereas SDL may not?
As such, this change preserves that (subjectively pleasant) behaviour.

Example [test.dem.zip](https://github.com/dxx-rebirth/dxx-rebirth/files/14544574/test.dem.zip) on 0.58.1 vs e59df93f:

https://github.com/dxx-rebirth/dxx-rebirth/assets/3589969/ba969e59-637a-4441-834b-06dce171cc0a

https://github.com/dxx-rebirth/dxx-rebirth/assets/3589969/b0bb4409-9cb6-4a56-903e-c291586ffc26

### DIFF
--- a/similar/arch/sdl/digi_mixer.cpp
+++ b/similar/arch/sdl/digi_mixer.cpp
@@ -586,9 +586,14 @@ void digi_mixer_end_sound(const sound_channel channel)
 
 void digi_mixer_set_digi_volume( int dvolume )
 {
+	/*
+	* Double-scale Mix_Volume to more closely match old pre-20a0166c Rebirth builds
+	* This also preserves the (unintentionally?) more accurate percieved volume scaling of those builds
+	* (e.g. perceived sound volume will match perceived midi volume more closely)
+	*/
 	digi_volume = dvolume;
 	if (!digi_initialised) return;
-	Mix_Volume(-1, fix2byte(dvolume));
+	Mix_Volume(-1, fix2byte(fixmul(dvolume, dvolume)));
 }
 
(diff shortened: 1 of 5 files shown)


vLKp:
Thank you for implementing this.  Fixing this has been on my task list for some time, but I had ranked it below the Github Actions task, which proved rather time consuming for the research it required.  I had the sense, which is likely incorrect based on how short your change is, that fixing this would itself be a time consuming project.

Merged.

alexstrout:
I never included a video of a v0.61 build before this change - attaching here for comparison, for the discussion in #753.

These are with the same in-game audio levels, so fair warning that everything will sound pretty loud! :)

https://github.com/dxx-rebirth/dxx-rebirth/assets/3589969/d2569693-e255-4bcc-94ad-2ff259925eb3

Issue #691, opened 2023-01-08
Feature request: publish on flathub.org for SteamDeck


JoeNotCharles:
It would be great to package dxx-rebirth as a flatpak and publish it through flathub.org (https://github.com/flathub/flathub/wiki/App-Submission).

flathub.org is the main 3rd-party feed for the SteamDeck, so this would let users easily install dxx-rebirth there.

vLKp:
This was previously requested in #680, but the requester never responded.  It's not clear to me from the linked page how I would test that Flatpak built the program correctly, which makes me disinclined to have them redistribute it.

JoeNotCharles:
Looks like there are more complete docs at https://docs.flatpak.org/en/latest/building.html. I'll see if I can get a build set up.

vLKp:
Thank you.  To elaborate on my comment above: I am uncomfortable posting for distribution a program I cannot actually test.  For Windows and for traditional Linux distributions, I can readily create a virtual machine of the appropriate environment to try it out.  When a Linux package maintainer packages Rebirth without me (which I have no objection to them doing), they (usually) direct their users to report problems to them first, and to upstream second, so if the package is broken due to a problem in the distribution, the relevant distribution maintainer hears about it and can deal with it.  When I post something directly, users reasonably expect they can come straight to me with a problem.  Thus, I don't want to post things which may be broken and give people the impression that I expect it to work.  This is compounded in the case of Flatpak, which appears to have a strange distribution and sandboxing model, so there are more opportunities for things to go wrong.

---

I tried reading through their setup documentation, and it does not look very promising.  I started at the link you gave, which starts off by telling me to read [Building your first Flatpak](https://docs.flatpak.org/en/latest/first-build.html) first.  I went there, and it starts off by telling me that I first need to read and complete [setup guide on flatpak.org](https://flatpak.org/setup/).  I went _there_, and it's just an ugly mess of icons, several of them far too large (Rocky Linux alone is 4096 pixels tall), and most are not properly labelled.  At that point I decided the rabbit hole was too deep and gave up.  I don't care to spend this much time just trying to get set up to learn how to build yet another package format.

---

I wish you luck with the setup, and I'm happy to answer any questions you may have about build problems.  I hope the flatpak build environment is typical enough that following the usual Linux instructions will work for you.

JoeNotCharles:
Yeah, the docs for flatpak are pretty poor. Now that I've dug into then
[...]

PR #759, opened 2024-03-10
Update Workflows

alexstrout:
Among other things, this PR: 

* Splits monolithic `package.yml` into separate reusable workflows
* Updates various names and spacing for readability
* Reverts builds to SDL1 builds w/ default SCons parameters
  * Choosing this for now until SDL2 support is out of beta
    (except for MacOS, where it's needed)
* Updates Linux deployment:
  * Use GCC 13
  * Remove ubuntu-20.04 package job (seemed redundant for building similar image w/ GCC)
  * Use existing linux/build_package.sh and simplify job a bit
* Updates MacOS deployment:
  * Target OS 14 and simplify job a bit
    * For some reason, macos-13 refuses to work with an error installing scons? No idea
    * macos-12 works, but [generates a few errors during compile](https://github.com/alexstrout/dxx-rebirth/actions/runs/8224723834/job/22488900411#step:4:191)
* Adds `ci.yml`, which runs automatically on merges / PRs to `master` and simply provides build status
* Adds `release.yml`, which can be run manually to generate a release tag
  * See https://github.com/alexstrout/dxx-rebirth/releases/tag/v0.0.2 for an example release from this action
  * Alternatively, you can simply:
    * Download the build artifacts from any workflow run
    * Manually create a release in the GitHub UI with the appropriate builds attached

These _should_ work as-is - hit me up if there's any trouble getting these running under the dxx-rebirth org. :P

Also currently, it looks like there's an issue with the Linux build failing due to a change in b3b4d90b - for demonstration purposes, I reverted that change in my master branch, which is how that example release above is compiled.

Thanks!

### DIFF
--- a/.github/workflows/package-macos.yml
+++ b/.github/workflows/package-macos.yml
@@ -5,8 +5,10 @@ on:
 
 jobs:
   package:
-    name: macOS
+    name: OS 14
     runs-on: macos-14
+    env:
+        MACOSX_DEPLOYMENT_TARGET: 14
     steps:
       - name: Clone Repository
         uses: actions/checkout@v4
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,7 +36,7 @@ jobs:
       - name: Create Release
         uses: softprops/action-gh-release@9d7c94cfd0a1f3ed45544c887983e9fa900f0564
         with:
-          tag_name: ${{ inputs.tag_name || github.ref }}
-          prerelease: ${{ inputs.prerelease || contains(github.ref, '-') }}
+          tag_name: ${{ inputs.tag_name || github.ref_name }}
+          prerelease: ${{ inputs.prerelease || contains(github.ref_name, '-') }}
           generate_release_notes: true
           files: '*/*.zip'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,9 @@ on:
         description: Pre-release?
         default: true
         type: boolean
+  push:
+    tags:
+      - "v*.*"
 
 jobs:
   package:
@@ -33,6 +36,7 @@ jobs:
       - name: Create Release
         uses: softprops/action-gh-release@v2
         with:
-          tag_name: ${{ inputs.tag_name }}
-          prerelease: ${{ inputs.prerelease }}
+          tag_name: ${{ inputs.tag_name || github.ref }}
+          prerelease: ${{ inputs.prerelease || contains(github.ref, '-') }}
+          generate_release_notes: true
           files: '*/*.zip'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,7 +34,7 @@ jobs:
         run: ls -R
 
       - name: Create Release
-        uses: softprops/action-gh-release@v2
+        uses: softprops/action-gh-release@9d7c94cfd0a1f3ed45544c887983e9fa900f0564
         with:
           tag_name: ${{ inputs.tag_name || github.ref }}
           prerelease: ${{ inputs.prerelease || contains(github.ref, '-') }}
--- a/contrib/packaging/linux/build_package.sh
+++ b/contrib/packaging/linux/build_package.sh
@@ -1,59 +1,30 @@
 #!/bin/bash
-
 set -eux -o pipefail
 
+arch=$(uname -m)
+
 # Grab latest AppImage package
-curl	\
-	--silent	\
-	--show-error	\
-	--location	\
-	--output '#1'	\
-	https://github.com/AppImage/AppImageKit/releases/download/continuous/'{appimagetool-x86_64.AppImage,AppRun-x86_64}'	\
-	|| exit 3
-chmod a+x appimagetool-x86_64.AppImage AppRun-x86_64
+curl --silent --show-error --location --output "#1" \
+    https://github.com/linuxdeploy/linuxdeploy/releases/download/continuous/"{linuxdeploy-x86_64.AppImage}" || exit 3
+chmod a+x linuxdeploy-x86_64.AppImage
 
 build_appimage() {
     name="$1"
     prettyname="$2"
 
-    appdir="${name}.appdir"
-    appimagename="${prettyname}.AppImage"
-
-    # Install
-    mkdir "${appdir}"
-
-    # Copy resources into package dir
-    mkdir -p "${appdir}/usr/bin"
-    cp --link "build/${name}/${name}" "${appdir}/usr/bin"
-
-    mkdir -p "${appdir}/usr/share/pixmaps"
-    cp --link "${name}/${name}.xpm" "${appdir}/usr/share/pixmaps"
-    cp --link "${name}/${name}.xpm" "${appdir}/"
-
-    mkdir -p "${appdir}/usr/share/icons/hicolor/128x128/apps/"
-    cp --link "${name}/${name}.png" "${appdir}/usr/share/icons/hicolor/128x128/apps/"
-    cp --link "${name}/${name}.png" "${appdir}/"
-
-    mkdir -p "${appdir}/usr/share/applications"
-    cp --link "${name}/${name}.desktop" "${appdir}/usr/share/applications"
-    cp --link "${name}/${name}.desktop" "${appdir}/"
-
-    # Package
-    cp --link "AppRun-x86_64" "${appdir}/AppRun"
-
     # Package!
-    "./appimagetool-x86_64.AppImage" --no-appstream --verbose "${appdir}" "${appimagename}"
-
-    # Clean
-    rm -rf "${appdir}"
+    export OUTPUT="${prettyname}.AppImage"
+    ./linuxdeploy-x86_64.AppImage \
+        --output appimage \
+        --appdir="${name}.appdir" \
+        --executable="build/${name}/${name}" \
+        --desktop-file="${name}/${name}.desktop" \
+        --icon-file="${name}/${name}.png"
 }
 
-# Build each subunit
+# Build each app
 build_appimage "d1x-rebirth" "D1X-Rebirth"
 build_appimage "d2x-rebirth" "D2X-Rebirth"
 
-# Consolidate into a single zip file
-zip -r -X DXX-Rebirth-Linux-AppImage-`uname -m`.zip D1X-Rebirth.AppImage D2X-Rebirth.AppImage
-
-# Clean
-rm -f appimagetool* AppRun* D1X-Rebirth.AppImage D2X-Rebirth.AppImage
+# Consolidate both apps into a single zip file
+zip -r -X "DXX-Rebirth-Linux-AppImage-${arch}.zip" "D1X-Rebirth.AppImage" "D2X-Rebirth.AppImage"
--- a/contrib/packaging/linux/build_package.sh
+++ b/contrib/packaging/linux/build_package.sh
@@ -4,8 +4,13 @@ set -eux -o pipefail
 arch=$(uname -m)
 
 # Grab AppImage package at specific version
-curl --silent --show-error --location --output "#1" \
-    https://github.com/linuxdeploy/linuxdeploy/releases/download/1-alpha-20240109-1/"{linuxdeploy-x86_64.AppImage}" || exit 3
+curl \
+    --silent \
+    --show-error \
+    --location \
+    --output "#1" \
+    https://github.com/linuxdeploy/linuxdeploy/releases/download/1-alpha-20240109-1/"{linuxdeploy-x86_64.AppImage}" \
+    || exit 3
 chmod a+x linuxdeploy-x86_64.AppImage
 
 build_appimage() {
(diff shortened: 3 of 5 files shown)


alexstrout:
@vLKp dunno what your notifications are on new PRs, but thought I'd tag you just in case you're also combating the release pipeline. It might also be easier to review the "final product" of these changes at https://github.com/alexstrout/dxx-rebirth/tree/workflows/.github/workflows instead of the raw diff.

FWIW, I'd like to look into possibly getting a [Flatpak](https://docs.flatpak.org/en/latest/introduction.html) distribution rolling, but packaging apps with that is new to me and I'm sure a Flathub distribution is a whole other can of worms too. I mostly suggest it because my end-user experience installing native apps from Flathub on a Steam Deck has been fantastic.

JeodC:
There is an existing PR  #751 that he's already trying to work on.

vLKp:
> Updates Linux deployment:
> *  Use GCC 13

I have been meaning to change that, since gcc before gcc-12 is no longer supported.  However, I wanted to see a workflow work once before I started making major changes.

> Remove ubuntu-20.04 package job (seemed redundant for building similar image w/ GCC)

I think the original intent of the author of that job was to have a library that was linked to the older libraries of Ubuntu 20.04, so that the game could be run on an Ubuntu 20.04.  However, since 20.04 does not have a nice way to get a new enough gcc, I had been considering dropping Ubuntu 20.04 from the automation.  @JeodC: do you see a reason to retain Ubuntu 20.04?

> macos-12 works, but generates a few errors during compile

MacOS has been a problem ever since I took an interest in using newer C++ standards, because clang has been badly behind, especially with regard to C++20 ranges support.  From a quick read of the linked log, I think this is more of the same.  The clang on macos-12 (which, for confusing historical reasons, is invoked as `g++` despite not being GNU GCC) is not able to build programs which use C++20 std::ranges.  Per the output in the log (`checking version of C++ compiler 'g++' ... 'Apple clang version 14.0.0 (clang-1400.0.29.202)'`), this clang is too old.  I thought I had made the SConf test strict enough that you would get a configure error for using a clang that lacks this support, but apparently not.  A future change might force a configure error for compilers which lack `std::ranges::subrange`.  I have no plans to remove use of `subrange` from the main program.

> Adds `release.yml`, which can be run manually to generate a release tag

Historically, the manual activation workflows have always failed.  They produce an in-page pop-up with a spinner that spins forever, no status messages, no error messages, and they never actually run the workflow.  This utter failure, in particular its lack of error reporting, soured me to Github Actions when the first actions were submitted to the repository.  If the site can't even report when something doesn't work, it's very difficult to debug.

> Adds release.yml, which can be run manually to generate a release tag

From a manual review, I see one problem already.  You removed the `contents: write` permission.  The previous submission of an automated workflow failed to create releases because it did not set `contents: write`, and the default permissions in this repository do not imply that permission.  My attempt to fix it in fdb41d2597b23ea5b8d13505af100f8e8235029f then broke the workflow because I added it in a scope that I thought would grant it as narrowly as possible, but Github Actions does not recognize that particular scope as a valid place for that directive.  I fixed my mistake in 3d4bfe5146e3bb343051bd4ae5fbd55898f51291, but as a change in the `.github` directory, that did not trigger a new workflow run.  Addressing that was high on my task list, until I saw that you are reworking the whole area.

> Also currently, it looks like there's an issue with the Linux build failing due to a change in b3b4d90

The build worked for me before I published it.  I see in the build log that you are trying to use gcc-11.  This is no longer supported or tested.  As an odd coincidence, the specific deduction guide on which gcc-11 is choking is only present at all because clang (all tested versions) failed to implicitly deduce the class template arguments, and needed the explicit deduction guide to tell it what to do.  gcc-12 and gcc-13 deduced the arguments correctly without a guide.

---

> dunno what your notifications are on new PRs

Github notifications stopped working about a year ago.  I always have "No unread events" according to the iconography in the corner, no matter how much work is actually pending.  This includes not being notified of `@username` mentions, not being notified about new or updated issue comments, and not being notified about pull request activity.  I only discover pending activity by looking at the unread notifications query page.  This is inconvenient, and I keep hoping Github will eventually fix the notification system.

> FWIW, I'd like to look into possibly getting a [Flatpak](https://docs.flatpak.org/en/latest/introduction.html) distribution rolling,

You are welcome to try.  Flatpak was previously requested in https://github.com/dxx-rebirth/dxx-rebirth/issues/691 ; at the time, I determined that Flatpak looked like too much trouble for something I would never be able to test, would never directly support, and had no personal use for.

alexstrout:
Whoops, I agree `permissions` go back into the workflow, so that jobs can't do anything unexpected. However - it was my impression that this keyword limited the available scope: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions
From my current understanding, I don't believe this can _grant_ any permissions that GitHub Actions aren't already allowed to perform on the repository? I'll try locking down access on my fork and seeing what the effects are.

I see the point on the older Ubuntu, but wouldn't that still exclude older or alternate distributions with different library versions, unless the necessary dependencies were included with the package? (I was thinking the AppImage script would address that, but it didn't seem to bundle the dependencies as expected? I wouldn't be surprised if I broke something though!)

vLKp:
@JeodC wrote:
> I could see no reason dxx-rebirth should not be able to work on older systems.

I agree, it would be nice if it worked on older systems, but from my perspective, the point of the Rebirth project is to make the old game (Descent) work on new systems.  For older systems that are still maintained well enough to support the specific tools I choose to use, I am happy to see the game run on those systems.  However, I do not keep old versions of compilers around, so keeping the game working with old tools is more trouble than I care to undertake.  I set the line at keeping the latest Debian (a distribution infamous for being outdated) working, and for a long time, I thought that was very generous.  I was surprised to see people expecting work I'd done recently to work on Ubuntu 18, which was already quite old by the time of the incident.  The Ubuntu 18 bit is in reference to a different exchange, not to the activity in this pull request.

@alexstrout wrote:

> However - it was my impression that this keyword limited the available scope: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

That is how I would design a security-focused system, yes.  However, my interpretation of the linked documentation is that this is not how GitHub Actions works:

> > You can use permissions to modify the default permissions granted to the GITHUB_TOKEN, adding or removing access as required,

Since it specifically mentions `adding`, I take it to be that a `permissions` block can increase a workflow's access relative to the repository default.

> I see the point on the older Ubuntu, but wouldn't that still exclude older or alternate distributions with different library versions, unless the necessary dependencies were included with the package?

Yes, it would exclude such systems.  However, from my reading, it is already not well supported to put on Ubuntu 20 a gcc new enough to meet the requirements I set for compiler support.  There exists an Ubuntu PPA that, for the time being, offers an adequate gcc.  There is no package in the Ubuntu 20 core repositories that provides an adequate gcc.  I have not tried to determine what, if any, other libraries would likewise be difficult to support.

Version handling for GNU libc is a problem when dealing with old systems.  As a general rule, a program linked to GNU libc version N cannot be run with a GNU libc version less than N, so supporting an old system requires that the build be done with an appropriately old GNU libc.  Using the currently active system libc is far easier than linking to a separate, older libc, so building for an old system strongly encourages building *on* an old system.  That in turn discourages me from trying to support old systems without a clear and convincing reason, such as a sizable userbase that cannot upgrade to a newer system.

> (I was thinking the AppImage script would address that, but it didn't seem to bundle the dependencies as expected? I wouldn't be surprised if I broke something though!)

Maybe.  I merged the AppImage support when someone else proposed it, and I saw neither security nor obvious correctness defects in the proposed implementation.  I've never run any AppImage on my system though, so I cannot vouch for whether the AppImage worked before your changes.

---

I merged this pull request, as it existed when I last fetched it.  I see that @alexstrout added further commits after my fetch and before my push.  I have glanced at them, but would like more time consider the linuxdeploy commit.

alexstrout:
The "Actually use GCC 13" commit e74473f may be worth applying, as that will fix the build issue with still using 11 to compile. That was an oversight on my part when I initially updated the job. (I thought it would use latest w/o needing to specify)

I've done a bit more work on a throwaway "workflow_test" branch, where I was attempting (among other things) to get an older MacOS deploy running - in particular on MacOS 12 using a newer compiler than Apple's outdated Clang, but this proved to be more trouble than expected.

I also consolidated the linuxdeploy changes into the actual contrib/packaging script, though that needs a bit more work. I'm also curious as to why the original AppImage scripts weren't bundling libraries properly, as I'd prefer to use the first-party tools if possible.

Comparison here, if interested: https://github.com/alexstrout/dxx-rebirth/compare/workflows...workflow_test
(Ignore the silly commits - I'd squash all that down to something readable before considering it release-ready. 😆)

I won't have time for it the rest of the weekend as I'll be out of town, but some time next week I'll keep hammering on it. 👍

JeodC:
I had based my pull request on maggu2810's work in his fork: https://github.com/maggu2810/dxx-rebirth

That is the original source for automation -- before that, there were platform specific packaging files in `/contrib/packaging` which are still there and used to bundle the final builds into more user-friendly releases.

vLKp:
I cherry-picked in most of @alexstrout's work, with the exception of the linuxdeploy change.  I pushed the result, and I am pleased to report that the GitHub Action ran to completion: https://github.com/dxx-rebirth/dxx-rebirth/actions/runs/8317230985 ; as far as I can tell, everything built correctly.  I have not tried running any of the built programs.  I notice that the `linux.zip` artifact is a zip that contains another zip, which is not great.  I may try to change that at some point.

On principle, I don't like downloading build tools as part of the run, which is why I have not moved forward with the linuxdeploy changes (and yes, I recognize this is restoring an old rule that was previously used; I was not happy with it then, and was happy to see it removed).  I have not ruled out accepting them, but I am hopeful I can find a way to build the requisite artifacts without downloading those tools.  In theory, if the tools were part of an appropriate GitHub Action (preferably one that GitHub endorses, or failing that, a wildly popular community Action), the workflow would be less vulnerable to a leftpad incident.  For now, I intend to leave the pull request open to deal with the linuxdeploy change, but I am also open to using it to discuss any other workflow updates either of you would like to propose.

JeodC:
> I notice that the `linux.zip` artifact is a zip that contains another zip, which is not great. I may try to change that at some point.

All artifacts are as `platform.zip` that contain the real .zip file. The release part of the actions is what will take them out of the `platform.zip` files and add them as releases. My fork does the same thing.

alexstrout:
I made some quick updates to this tonight - not sure how these changes will fly, but I cleaned up the package scripts a bit and also switched the AppImage to use the linuxdeploy method @maggu2810 originally used in their fork. The main reasoning here is that this appears ready-made to save some of the manual work building an AppImage - particularly, in this case, including the libraries necessary for the AppImage to function as a "portable app" across different distros as intended. There's some info from the AppImage documentation here: https://docs.appimage.org/packaging-guide/from-source/linuxdeploy-user-guide.html

I also switched the linuxdeploy curl to use a specific release tag (instead of the continuous tag) and the gh-release action to use a specific commit SHA to at least offer some supply-chain protection.

Finally, I enabled tag triggers on the release action. Some samples of this are available here: https://github.com/alexstrout/dxx-rebirth/releases
These trigger off a specific v*.* tagging pattern. The presence of a dash "-" is used to hint at a pre-release - e.g. "v0.0.5-beta" as shown in the examples above. The "Full Changelog" bit is auto-generated but can of course be edited as usual after the release is created.

All pipelines can still be triggered manually via dispatch, since it was useful for testing changes - but there should be no manual interaction required for a release besides simply tagging a release-worthy commit. LMK any thoughts / suggestions on this!

vLKp:
@JeodC wrote:
> All artifacts are as `platform.zip` that contain the real .zip file. The release part of the actions is what will take them out of the `platform.zip` files and add them as releases.

Right, but I think it would be nicer if the artifact stage was not double-zipping them in the first place, since it looks weird to download an artifact, get a zip file, and find within that another zip file.

It is nice that this would be better for the generated releases, but I'd like to see it right in both places.  I have not taken the time to look at it yet, though.

---

@alexstrout wrote:

> I made some quick updates to this tonight - not sure how these changes will fly, but I cleaned up the package scripts a bit

Some of the quoting changes appear unnecessary to me, but I am not inclined to request a rework over that.  The multiline curl was intentional so that each parameter was on its own line, making it easy for `git blame` to show changes over time.

> I also switched the linuxdeploy curl to use a specific release tag (instead of the continuous tag) and the gh-release action to use a specific commit SHA to at least offer some supply-chain protection.

Considering the events of the last day or so ( [xz backdoor](https://lwn.net/Articles/967180/) ) this phrasing seems eerily prescient.

> LMK any thoughts / suggestions on this!

I have not looked at the code beyond the first commit yet, but your summary sounds promising.

alexstrout:
Wow, that sucks - thanks for the link. Good intuition on the reporter. I feel bad for the original xz maintainer - hopefully they get some good support even after this all dies down...

I'll restore the multi-line curl bit, I agree that makes sense.
The quoting is definitely overkill, I just found it nice at a glance when using a syntax-highlighting editor.

PR #764, opened 2024-03-15
Add game enemy descriptions to comment

olivier1980:
Adding enemy descriptions as used in the game to robot_id list

### DIFF
--- a/similar/main/ai.cpp
+++ b/similar/main/ai.cpp
@@ -229,29 +229,30 @@ point_seg_array_t::iterator       Point_segs_free_ptr;
 
 // ------ John: End of variables which must be saved as part of gamesave. -----
 
-//	0	mech
-//	1	green claw
-//	2	spider
-//	3	josh
-//	4	violet
-//	5	cloak vulcan
-//	6	cloak mech
-//	7	brain
-//	8	onearm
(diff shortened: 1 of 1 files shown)


vLKp:
There are a number of robots that were never documented in the comment you propose to update, including I think all of the Descent 2 robots.  What prompted you to want to adjust this comment?  I think it would be better to merge this comment with the `enum class robot_id` in `robot.h`, and in the process define corresponding enum members for all the known robot types.  For example:

```diff
diff --git a/common/main/robot.h b/common/main/robot.h
index f22e3cb7c..37a302bda 100644
--- a/common/main/robot.h
+++ b/common/main/robot.h
@@ -50,6 +50,8 @@ enum class robot_animation_state : uint8_t
 
 enum class robot_id : uint8_t
 {
+	mech = 0,	// Medium Hulk
+	green_claw = 1,	// Medium Lifter
 	brain = 7,
 	toaster = 10,
 	baby_spider = 14,
```

and so on, for all the robots.  This way, the debugger can show the symbolic name for any robot.  Presently, the `enum` only has entries for the robots that were handled specially in the code.

olivier1980:
Yes, that a better solution, I'll close this PR

Issue #747, opened 2024-01-06
Shareware content missing from website


theofficialgman:
shareware content is missing from the website

https://www.dxx-rebirth.com/download/dxx/content/descent-pc-shareware.zip
https://www.dxx-rebirth.com/download/dxx/res/d1xr-hires.dxa
https://www.dxx-rebirth.com/download/dxx/res/d1xr-sc55-music.dxa
https://www.dxx-rebirth.com/download/dxx/content/descent2-pc-demo.zip

and potentially more

JeodC:
I've stored most addons at https://github.com/JeodC/Portmaster-Descent/tree/main/addons

If vLKp wants to upload them here he's more than welcome to.

vLKp:
It looks like @zicodxx switched the site to a simpler goodbye page than had been used, and these may have been lost in the process.

@zicodxx: will these be put back or do they need to be mirrored somewhere else going forward?  If the latter, some of the build scripts will need to be changed not to refer to `dxx-rebirth.com` anymore.

theofficialgman:
For now I will be downloading using the archived versions from wayback machine https://web.archive.org/web/*/https://www.dxx-rebirth.com*

vLKp:
It appears zico's Github account no longer exists.  My questions above may go unanswered.

PR #765, opened 2024-03-17
Add freedesktop metadata for D1X

fpiesche:
This adds a `metainfo.xml` file for D1X which can be used by packages to display additional data about it, as well as screenshots showing off various features to present in this. This information is used e.g. by Flathub's web listings, GNOME's Software app and KDE's Discover.


vLKp:
Adding freedesktop metadata seems like a reasonable change, but I am reluctant to merge screenshots into the main repository.  The entire history to date is ~41M.  The screenshots alone are ~4.7M, and would be carried forever once merged into the history.  Do you have a citation for this being standard practice?

Separately, this repository covers both D1X-Rebirth and D2X-Rebirth.  I think if D1X-Rebirth needs metadata, then it would be appropriate to also add metadata for D2X-Rebirth.

fpiesche:
I am in the process of creating metadata for D2X-Rebirth. ;)

I don't think there's any particular precedent re: screenshots; I've mostly added them here as the ones on dxx-rebirth.com went away once zico retired. I was also planning to add a PR to add the add-ons formerly hosted over there in here (with git-lfs enabled)... if you want to keep this repository clean of such things, might it be worth adding another repo to the dxx-rebirth "org" for assets like that instead, which would remain pretty static?

alexstrout:
This is awesome to see! I agree that a separate DXX-Rebirth-Assets repo might be nice for raw assets, versus checking them into the main code repo.

vLKp:
Zico's retirement caught me by surprise too.  I had not taken a full inventory of what disappeared when he replaced the site with the retirement announcement.

Yes, I think an assets repository would be a good idea.  That would also be a way to address issue #747.  Unfortunately, many non-programming maintenance tasks like this have accumulated, and the GitHub web UI's generally broken behavior has stopped me from dealing with it.  Many things which once worked now just fail, without any error message.  Fighting broken websites is not high on my list of ways to spend my limited time, so such tasks are on hold until GitHub's UI improves.

fpiesche:
Will re-create without screenshots.

vLKp:
I have not forgotten this pull request.  However, it has been on hold until I can create the asset repository for the screenshots.  Presently, the page that should do that comes up with only the GitHub navigation banner, and no content, so I cannot create new repositories.  I rarely create new repositories, so I don't know how long this has been broken, or when GitHub will fix their page.

fpiesche:
I've created a separate repository for assets (screenshots, add-ons, demo/shareware data, etc) and done so over on Gitlab rather than Github - my reasoning there is that Github has fairly stringent limitations on maximum repository size for non-Enterprise accounts (1GB), which the assets repository skirts very close to already. Gitlab, on the other hand, cap their repo size at 10GB for free accounts.

I've run into this problem with another project I was making Flatpak builds for who don't have file storage with fixed URLs ;)

PR #782, opened 2024-06-14
Fix GitHub Actions Linux builds

fpiesche:
Apparently at some point in the last few weeks, the GCC 13 packages seem to have vanished from Ubuntu 22 repos. From looking around these are now only available on Ubuntu 24.04, the current LTS release.

I've updated the Linux workflow to build on Ubuntu 24 instead - GitHub's runners for this are still in beta and have long queues for jobs, so I've moved the build into an Ubuntu 24 Docker container running on GitHub's `ubuntu-latest` runners (being the standard ones, but currently still on 22.04) and then use the artifacts in a separate job to run the `AppImage` packaging.

### DIFF
--- a/.github/workflows/package-linux.yml
+++ b/.github/workflows/package-linux.yml
@@ -4,11 +4,9 @@ on:
   workflow_call:
 
 jobs:
-  package:
+  build:
     name: GCC 13
-    runs-on: ubuntu-latest
-    container:
-      image: "ubuntu:24.04"
+    runs-on: "ubuntu:24.04"
     steps:
       - name: Clone Repository
         uses: actions/checkout@v4
@@ -27,11 +25,5 @@ jobs:
           CC: gcc-13
           CXX: g++-13
 
-      - name: Package
-        run: ./contrib/packaging/linux/build_package.sh
-
-      - name: Upload Artifact
-        uses: actions/upload-artifact@v4
-        with:
-          name: linux
-          path: '*.zip'
+      - name: Package AppImage
+        run: contrib/packaging/linux/build_package.sh
--- a/.github/workflows/package-linux.yml
+++ b/.github/workflows/package-linux.yml
@@ -38,7 +38,11 @@ jobs:
     runs-on: "ubuntu-latest"
     steps:
       - name: Install AppImage dependencies
-        run: sudo apt-get update && sudo apt-get install -y libphysfs1 libfuse2 wget curl
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y \
+            libphysfs1 libsdl1.2 libsdl-image1.2 libsdl-mixer1.2 libpng \
+            libfuse2 wget curl
 
       - name: Clone Repository
         uses: actions/checkout@v4
--- a/.github/workflows/package-linux.yml
+++ b/.github/workflows/package-linux.yml
@@ -52,6 +52,5 @@ jobs:
 
       - name: Package AppImage
         run: |
-          mv linux-loose/build .
-          rmdir linux-loose
+          ls -la
           contrib/packaging/linux/build_package.sh
--- a/.github/workflows/package-linux.yml
+++ b/.github/workflows/package-linux.yml
@@ -49,8 +49,9 @@ jobs:
         uses: actions/download-artifact@v4
         with:
           name: linux-loose
+          path: build/
 
       - name: Package AppImage
         run: |
-          ls -la
+          ls -la ./build
           contrib/packaging/linux/build_package.sh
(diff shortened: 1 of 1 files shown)


fpiesche:
You'll want to squash&merge this, sorry for the messy commit history.

vLKp:
I noticed the gcc 13 build failure, and thought it odd that the contributor who had switched the GitHub Action to use gcc-13 could have missed that.  I like your explanation better, that it once worked, and now fails.  I have a pending change that switched the GitHub Action down to gcc-12, since that is still supported by Rebirth, and likely will be until I can expect Debian users to have access to gcc-13 (or higher).

For my edification, could you explain why you chose to have this run in two separate jobs?  The GitHub Actions were largely written by outside contributors, and have only minor changes from me.  I have relatively little experience using GitHub Actions, so I am curious if there is some technical value in splitting these.

alexstrout:
gcc-13 on u22 was [working as of early May](https://github.com/alexstrout/dxx-rebirth/actions/runs/9006253496), and as fpiesche noted seems to have disappeared since then - weird, but good find. 👍 

However, I'm not sure I follow the remaining changes in the pipeline file? (everything below line 11)
Are these needed to support packaging in the container?

fpiesche:
Yeah - the appimage builder uses `fuse` which isn't available inside a Docker container by default due to the sandboxing there. As such I generate the actual build inside a Docker container, export it as an artefact and then run the appimage builder tool on the runner directly.

alexstrout:
Gotcha - that makes sense. Thanks!

Is there any merit to instead just using gcc-12 on the stable u22 runner, until later versions make their way down? (Genuine question - seems like either approach works)

fpiesche:
None I can think of; I had assumed there was some reason to the build using GCC13 so figured I'll fix that rather than downgrade it :D

Having the build containerised like this, though, makes it a hell of a lot easier to switch build toolchains (or even build multiple different ones via a [matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs)), as the build workflow now simply builds with whatever the default gcc/g++ executables are rather than specifying them.

If we're not married to GCC 13, it might even be worth trying to switch the container over to one of the `manylinux` ones - this is a project intended mostly for Python packages with binary components, but generally aims to provide a "lowest common denominator" C/C++ Linux build environment to maximise compatibility with different glibc versions across distributions etc.

vLKp:
Both gcc-12 and gcc-13 are supported and will be for some months yet.  Hopefully, by the time I drop gcc-12 support, GitHub runners will be able to provide gcc-13 (or higher) without the need for using Docker to download yet another OS runtime.

I pushed my switch to gcc-12 for now, but am leaving this pull request open because the more involved changes here seem worth considering.  Currently, I direct people who want a prebuilt binary to the artifacts from the CI build, which works since those artifacts are a final product.  If the artifacts start being used as an inter-stage communication mechanism, I may need to avoid directing people to the artifacts page.

Improving compatibility across multiple distributions is a nice goal.  I don't have any insight yet into how many Linux users get Rebirth from these builds versus through a distribution package.

alexstrout:
> Currently, I direct people who want a prebuilt binary to the artifacts from the CI build, which works since those artifacts are a final product.

@vLKp One important note on this: Users must be logged into a valid GitHub account to download build artifacts, which are only retained for a [period of up to 90 days on public repositories](https://docs.github.com/en/organizations/managing-organization-settings/configuring-the-retention-period-for-github-actions-artifacts-and-logs-in-your-organization).

A release tag is a bit more user-friendly by comparison - see the [example releases](https://github.com/alexstrout/dxx-rebirth/releases) used both to test the pipelines, and to provide myself with a working build for my Steam Deck :P

These were created just by pushing a `v*.*` tag - the workflow took care of the rest and automatically created the release w/ default change notes. (see the push:tags condition in release.yml and the actions it performs after)

---

Random anecdote: Deck seems like it's been a big kick in the rear for 'nix gaming, particularly retro - landing this on Flathub (e.g. via @fpiesche's recent work) would probably find a decently healthy (if slightly less patient / technical) audience. For what it's worth, Rebirth runs fantastically on it, and likely similar / much less powerful devices!

fpiesche:
I have a workflow locally here that will create a new release tag when any code is pushed to `master`, automatically parsing commit messages to figure out whether the bump should be major, minor or patch - combined with the existing "create release" workflow, this has now put my local repo into a state where merging a PR creates a tag, runs a build, and then creates a GitHub release with the linux/Windows/macOS executables from the build :)

I'll create a PR to add this soon, at which point we also have everything in place here we need for me to submit DXX-Rebirth to FlatHub!

PR #785, opened 2024-06-17
Add FreeDesktop metainfo file for d2x

fpiesche:


fpiesche:
Augh, branched off the wrong place here.

PR #792, opened 2024-07-06
Add d2x metainfo

fpiesche:
Actually branched off of current master this time so there's no other guff polluting the PR.


vLKp:
You removed the extra commits (which could have been done in-place by force-pushing over the old branch), but the last image URL still points into the dxx-rebirth repository, and is therefore broken.  Would you like for me to fix that up as part of the merge?

fpiesche:
Sorry about that - looks like I pasted one too many `<screenshot>` segments in there by accident.

Issue #711, opened 2023-04-07
Last compiled version is from 2022/09/29


snytek:
Hi

From time to time I check for new versions on:
https://www.dxx-rebirth.com/download-dxx-rebirth/

Since a few months there seems to be no update. The last version is dated 2022/09/29.
`commit: [e3faab0](https://github.com/dxx-rebirth/dxx-rebirth/commit/e3faab0899bc86184fdd67568ccc4f237c5f6807), date: 2022/09/29`

Is there maybe an issue with an auto deployment?

Thanks

vLKp:
I am not aware of an auto-deployment.  As far as I know, the builds posted on the website are prepared by @zicodxx manually, and then uploaded.

LordKane1701:
> I am not aware of an auto-deployment. As far as I know, the builds posted on the website are prepared by @zicodxx manually, and then uploaded.



> Hi
> 
> From time to time I check for new versions on: https://www.dxx-rebirth.com/download-dxx-rebirth/
> 
> Since a few months there seems to be no update. The last version is dated 2022/09/29. `commit: [e3faab0](https://github.com/dxx-rebirth/dxx-rebirth/commit/e3faab0899bc86184fdd67568ccc4f237c5f6807), date: 2022/09/29`
> 
> Is there maybe an issue with an auto deployment?
> 
> Thanks

I had a brief talk with @zicodxx on twitter, about this issue before this ticket was submitted. he hasnt had alot of time as of late with some compilation issues has made new versions slow down. he doesnt know when he can get back to it, but you can fork and use github actions to create a local version.

maggu2810:
I created a fork of this repo that uses a GitHub action to create Linux, Windows and Mac packages.
I did not test the Mac and Windows build result, but feel free to do so.

snytek:
Thanks @maggu2810! 
I've tested the windows compile and it works!
If this is a fork, it is a copy, right? How do you make sure the executables are up to date when code changes?

maggu2810:
It is a fork but only for packaging using GitHub actions and not to work on the code base.
ATM merge upstream changes and trigger the build involves a few clicks, but I plan to automate this step so every day a new build (amd merge of upstream changes) is done automatically.

snytek:
This sounds great. 
[...]

PR #797, opened 2024-08-01
Add automatic release tagging

fpiesche:
This adds a new workflow, `tag.yml`, which will automatically create a new semver tag on the repository when changes are merged to `master` (or the more modern equivalent `main` in case `master` ever gets renamed there); additionally this will then run the existing `release.yml` workflow to create a new GitHub release for the new tag.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,7 +40,7 @@ jobs:
         run: ls -R
 
       - name: Create Release
-        uses: softprops/action-gh-release@v2
+        uses: softprops/action-gh-release@9d7c94cfd0a1f3ed45544c887983e9fa900f0564
         with:
           tag_name: ${{ inputs.tag_name || github.ref_name }}
           prerelease: ${{ inputs.prerelease || contains(github.ref_name, '-') }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,6 +14,7 @@ on:
     inputs:
       tag_name:
         description: Tag Name (new or existing)
+        required: true
         type: string
   push:
     tags:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,7 +34,7 @@ jobs:
         run: ls -R
 
       - name: Create Release
-        uses: softprops/action-gh-release@9d7c94cfd0a1f3ed45544c887983e9fa900f0564
+        uses: softprops/action-gh-release@v2
         with:
           tag_name: ${{ inputs.tag_name || github.ref_name }}
           prerelease: ${{ inputs.prerelease || contains(github.ref_name, '-') }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,11 @@ on:
         description: Pre-release?
         default: true
         type: boolean
+  workflow_call:
+    inputs:
+      tag_name:
+        description: Tag Name (new or existing)
+        type: string
   push:
     tags:
       - "v[0-9]+.[0-9]+.[0-9]+"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,7 +12,7 @@ on:
         type: boolean
   push:
     tags:
-      - "v*.*"
+      - "v[0-9]+.[0-9]+.[0-9]+"
 
 jobs:
   package:
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -21,6 +21,8 @@ jobs:
   release:
     name: Create release
     needs: tag
+    permissions:
+      contents: write
     uses: ./.github/workflows/release.yml
     with:
         tag_name: ${{ needs.tag.steps.tag_version.outputs.new_tag }}
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -1,0 +1,19 @@
+name: Create version tag
+on:
+  push:
+    branches:
+      - master
+      - main
+
+jobs:
+  tag:
+    name: Create tag
+    runs-on: ubuntu-22.04
+    permissions:
+      contents: write
+    steps:
+      - name: Bump version and push tag
+        id: tag_version
+        uses: mathieudutour/github-tag-action@v6.2
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -25,4 +25,4 @@ jobs:
       contents: write
     uses: ./.github/workflows/release.yml
     with:
-        tag_name: ${{ needs.tag.steps.tag_version.outputs.new_tag }}
+      tag_name: ${{ needs.tag.steps.tag_version.outputs.new_tag }}
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -8,6 +8,7 @@ on:
 jobs:
   tag:
     name: Create tag
+    id: create_tag
     runs-on: ubuntu-22.04
     permissions:
       contents: write
@@ -18,7 +19,9 @@ jobs:
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Create release
-        uses: release.yml
-        with:
-          tag_name: ${{ steps.tag_version.outputs.new_tag }}
+  release:
+    name: Create release
+    needs: create_tag
+    uses: ./.github/workflows/release.yml
+    with:
+        tag_name: ${{ needs.create_tag.steps.tag_version.outputs.new_tag }}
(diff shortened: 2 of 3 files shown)


vLKp:
I prefer that the use of `softprops/action-gh-release` remain pinned on a digest, rather than a tag, so that it cannot change unexpectedly.  If it needs to be updated to a newer version, that is fine, but it should be a newer digest, rather than switch to a tag.

I would like to understand the automatic tagging code before merging this.  It appears to be some unrecognized (to me) person's personal action, and is again identified by tag rather than a pinned digest.

I am not enthusiastic about polluting the repository with tags on every push, though I recall from one of your other posts that the Flatpak project strongly prefers use of tags, and will be difficult to integrate without them.

Removing zico's contact information doesn't really belong here.  He's never indicated whether he wants to be dropped.  However, I can see an argument for dropping the section labelled "Contact" since the registered domain does not provide any way for users to get help or learn about this version.

fpiesche:
> I prefer that the use of `softprops/action-gh-release` remain pinned on a digest, rather than a tag, so that it cannot change unexpectedly. If it needs to be updated to a newer version, that is fine, but it should be a newer digest, rather than switch to a tag.

> Removing zico's contact information doesn't really belong here. He's never indicated whether he wants to be dropped. However, I can see an argument for dropping the section labelled "Contact" since the registered domain does not provide any way for users to get help or learn about this version.


Both reverted.


> I would like to understand the automatic tagging code before merging this. It appears to be some unrecognized (to me) person's personal action, and is again identified by tag rather than a pinned digest.

This is indeed the case. I picked up on the existence of this action via other projects I've contributed to and I can at least vouch for its functioning. It's also still actively maintained and [fairly widely used](https://github.com/search?q=mathieudutour%252Fgithub-tag-action) across github, including by [RedHat](https://github.com/redhat-marketplace/redhat-marketplace-operator/blob/2ab9172a72cb3e1574ca5a43723ab9414f71a754/internal/ci/workflows.cue#L93)  and the [government of British Columbia](https://github.com/bcgov/bcregistry-sre/blob/5f252d61de5db3d8cd393b1f6c39bb3c8e4c30b5/composite-actions/ui-cd/action.yaml#L59). I'd previously looked at a number of other GitHub Marketplace actions for this purpose and decided this was by some margin the most trustworthy and reliable option.

> I am not enthusiastic about polluting the repository with tags on every push, though I recall from one of your other posts that the Flatpak project strongly prefers use of tags, and will be difficult to integrate without them.

I see your thinking here; my reasoning for doing this is twofold:

- Flathub as an organisation specifically don't want "development snapshot" builds, i.e. builds happening whenever someone commits to master - I appreciate that this clashes with your workflow for the project, so I figured having releases/tags created on commits to `master` at least signals the intent that `master` will be in a releasable state at any given point.
- more importantly I'm thinking about maintainability of the Flatpak here - Flathub's CI infrastructure includes a feature that can monitor source repositories for the builds for new releases - via e.g. anitya or just arbitrary JSON web APIs, and the "default" option for this is git tags - and automatically creates a PR to update the Flatpak build with that release and runs a CI build to a testing repository for verification, then automatically publishes the update to Flathub once the PR is merged. Without tags or otherwise publicly "visible" releases of some shape, the Flatpak build manifest would need to be manually updated whenever you want to do so.

Would you prefer a workflow you can run manually to create a new tag and release? From conversations in other PRs I thought someone else had prepared workflows to basically do this automatic release process so I assumed this would be your preferred option.

If you'd rather use the github web UI to manually run a workflow that'll do the whole release creation thing I'll be happy to help get that set up also, though I also get the drift that you're understandably not a fan of github's web UI. The other, nearer-to-development, option would be to just create tags in git manually when you feel you've reached some kind of milestone in development that's worth "advertising" as a release and having a workflow that then flags this tag as a "Release" on Github. Updates for the Flatpak build would also automatically fall out of this process change as mentioned above (re: how Flathub's CI infrastructure handles updates).

jvalecillos:
Hello, I found this PR while looking for the latest version of this port to play these classic games and also saw #711 about getting newer builds.

From the discussion here, it sounds like @fpiesche and others are suggesting that a new release (with tags) should be created automatically every time a change is merged into the main branch (`master`). This makes sense if we're following a workflow where each feature or bug fix is done on its own branch, tested, then merged into main to mark a new version. But I get that the maintainers here are working a bit differently —often committing directly to main— so they may prefer a way to decide when a new release actually happens.

Here’s a middle-ground approach I’d like to suggest:

1. **Flexible Commit Style**: Authors can either merge feature branches or commit directly to main, whatever works best.
2. **Manual Tagging for Releases**: When maintainers feel the main branch is ready for a new release, they can manually add a tag (using semantic versioning if possible, like major, minor, patch).
3. **Automated Build & Release**: A GitHub Action workflow kicks in whenever a new tag is added. It would:
   - Build the project for all supported platforms.
   - Create a release with the tag.
   - Attach the compiled versions and zipped source code to the release.
4. **Flatpak Manifest Update**: The FlatHub manifest would be updated to point to this new tag. This should be done separately.
5. **(Optional) Continuous Release**: We could set up another workflow that makes a "continuous" release tag with each merge to main. This would always point to the latest commit and overwrite the previous one, providing a kind of pre-release for anyone wanting the latest version.

This setup gives maintainers control over official releases while offering a way for users to access the latest changes. Let me know what you think.

fpiesche:
For the sake of getting more recent builds into circulation, I've started looking at Flathub's beta branch which may be useful for the workflow we've got here - the main Flathub repository is meant for stable, well-tested releases and thus they're not fond of having anything other than tagged releases in there, *but* they offer a "beta" channel which can be more in-flux.

As of right about now, the current state of DXX-Rebirth's `master` branch is available on the Flathub `beta` channel, with automatic updates configured over there so when a change lands on `master` here it will be available there ASAP after:

`flatpak remote-add --user flathub-beta https://flathub.org/beta-repo/flathub-beta.flatpakrepo
flatpak install --user flathub-beta io.github.dxx_rebirth.dxx-rebirth`

jvalecillos:
> As of right about now, the current state of DXX-Rebirth's master branch is available on the Flathub beta channel, with automatic updates configured over there so when a change lands on master

@fpiesche Thank you! That’s great news for Linux users on Flatpak.

That said, it would still be beneficial to have tagged releases for users on other operating systems like Windows, macOS, or Linux (for those using AppImage). With proper versioning, the project could also be integrated with distro package managers and compiled for other architectures, such as ARM.

The more accessible it is to get the latest version up and running, the more likely we’ll see a growing community and potentially attract new contributors to keep improving this project.

vLKp:
I never wanted to be a release manager for the project, though that seems to have fallen to me by default as the only remaining active person with direct commit access.  I don't have a good answer for preferred process, though I can address some of the questions about this area.

Yes, I have historically tried to hold the position that `master` is always release-ready.  Marking every point where I publish as its own release feels like it would irritate users who don't want to see a new must-upgrade release every few weeks, since I usually don't make big upgrade-worthy changes that often.  Most activity is more subtle.

Someone did prepare a GitHub Actions workflow for creating releases in response to a trigger from the web UI.  It has never worked for me.  Specifically, when I tell the GitHub Web UI to run it, I get a modal that spins forever and nothing happens, not even an error message.  I gave up trying to get it working, but left it because apparently it works for some people under some circumstances.

Considering the comments above about how widely used the auto-tagging action is, I am inclined to withdraw my earlier concern and use the tagging action that @fpiesche suggested.

@fpiesche wrote:

> though I also get the drift that you're understandably not a fan of github's web UI.

That is very true.  GitHub's web UI has been getting steadily more frustrating to use with each "improvement" that GitHub rolls out for it, with the result that I use the web UI less and less.  I have far less time for Rebirth than I once did, and I don't like spending what time I have fighting with the web UI.

---

@jvalecillos: my workflow is a bit of a hybrid in that regard.  I develop locally, do whatever testing I think is appropriate / needed, then push that whole batch of changes to `master` in a single step.  Thus, while it looks to the outside world like I commit directly to `master`, internally it's a bit more controlled.  Some projects get deferred, reworked, or even abandoned, before anyone watching GitHub even knows I was working on them.

As it stands, there is a GitHub Actions workflow that reacts to each push to `master` by building for Windows/Linux/Mac, and uploading the built artifacts.  This is not quite as thorough as what you are suggesting, but provides some basic prebuilt binaries for the major platforms.

As I understand git tag conventions, a rolling tag would confuse people and tooling.  Tags are supposed to reference a specific point in history, and never change.

---

Regarding FlatHub's preference for stable, well-tested releases: I understand their position, and from the perspective of a user consuming software, I like that position.  However, that position almost argues against having an automated tagging mechanism for Rebirth at this time.  I test as best I can before pushing, but I don't have any users who provide feedback before I push commits to master, so I can't really say that tags pointing to master are "well tested."  Some areas, such as multiplayer code, get very little testing, so I have stayed away from making further changes there.

PR #798, opened 2024-08-01
Fix package names

fpiesche:
I got the package names wrong here, as I've found out when trying to get this through Flathub's CI. The first four components should be io.github.owner.repository, and of course there are no `dxx-rebirth/d1x-rebirth`/`dxx-rebirth/d2x-rebirth` repositories. Fixed that...


vLKp:
If the last component is repository, shouldn't it be `dxx-rebirth`, not `dxx_rebirth`?  The form with a dash is how GitHub actually presents it.  Using the form with an underscore returns a 404:

- https://github.com/dxx-rebirth/ (works)
- https://github.com/dxx_rebirth/ (broken)

PR #809, opened 2025-02-07
Fix release pipeline / missing Windows ogg playback

alexstrout:
While trying a recent pipeline build, I noticed the release workflow no longer attaches any .zip files.

On closer inspection, it looks like the build artifacts are now folders instead of the pre-packaged .zips; see "Display Artifacts" step of [job/36825594735](https://github.com/alexstrout/dxx-rebirth/actions/runs/13191581990/job/36825594735) (note GitHub requires sign-in to view log details / artifacts)

This seems reasonable enough, so I added a final zip step to the release job. (and changed the path accordingly)

I also noticed that on Windows, .ogg playback (e.g. addon music) was failing due to missing `libvorbisfile-3.dll` (and its dependencies) - since it's dynamically loaded, ldd seems to miss it, so as a workaround I added a manual copy command to grab it (+ deps).

Here's a result of the above actions:
https://github.com/alexstrout/dxx-rebirth/releases/tag/v0.0.11

Open to suggestions / alternatives etc. Thanks!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,10 +33,13 @@ jobs:
       - name: Display Artifacts
         run: ls -R
 
+      - name: Compress Artifacts
+        run: ls . | while read dir; do cd ${dir}; zip -r ../${dir}.zip *; cd ..; done
+
       - name: Create Release
         uses: softprops/action-gh-release@9d7c94cfd0a1f3ed45544c887983e9fa900f0564
         with:
           tag_name: ${{ inputs.tag_name || github.ref_name }}
           prerelease: ${{ inputs.prerelease || contains(github.ref_name, '-') }}
           generate_release_notes: true
-          files: '*/*.zip'
+          files: '*.zip'
--- a/contrib/packaging/windows/build_package.sh
+++ b/contrib/packaging/windows/build_package.sh
@@ -20,7 +20,7 @@ build_app() {
 	local name prettyname
     name="$1"
     prettyname="$2"
-    
+
     # Create a subdirectory for each app at the top level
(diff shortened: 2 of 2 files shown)


vLKp:
I suspect that the zip change was from [Include project and architecture in Linux artifact zip filename](/dxx-rebirth/dxx-rebirth/commit/7223dc525c657ec27f9bea06ac2a64f2d41cc254) and [Include project and architecture in Windows artifact zip filename](/dxx-rebirth/dxx-rebirth/commit/93d6512ca6ab6f7507d56d57a4b25ce956c3205f).  Prior to those, the artifact zip contained another zip, which then contained the actual files.  I changed it to have only a single level of zip.

I have never gotten the release workflow to even start successfully (it always hangs with a spinner, and no error), so I tend not to think about it.

As regards the proposed changes:
```diff
+        run: ls . | while read dir; do cd ${dir}; zip -r ../${dir}.zip *; cd ..; done
```

Using `ls` in a pipeline is usually wrong.  In this case, I think it would be better to write this as:
```
for dir in */; do (cd "$dir" && zip -r "../${dir%/}.zip" .); done
```

By using `*/`, you cause the shell to find all directories, and ignore all non-directories.  By using a sub-shell for the loop body, you avoid the need to `cd ..` at the end.  The bash parameter transformation is required since my form has a trailing `/` on the directory name, which needs to be removed to avoid mangling the path passed to `zip` as the desired zip filename.

All the other changes look good to me.

alexstrout:
D'oh, I forgot to return to this and make that suggested change. Thanks for fixing that!

PR #825, opened 2025-12-11
Fix unchecked optionals from causing a crash on level load for objects with mistakenly used AI control type

AlumiuN:
If an object has the wrong control type, it will be overwritten later in the load process, making crashing here bad both because it results in an inconsistent and unhelpful crash (for me it was just a hang, for others it could be a clean crash, or a window with a "bad optional access" exception message) but also because the level will load fine anyway. This specifically fixes level 4 of [PTMC: Castaway](https://sectorgame.com/dxma/mission/?m=2270) but potentially also other levels.

There may also be other optionals that are similarly unchecked before attempting to extract the value from them (and I'm sorry for not checking deeper than this, I am _very_ tired) ; I think in every case it would be more reasonable to check that the value doesn't exist and produce a warning/error message instead of crashing or reporting the extremely vague "bad optional access" error.

### DIFF
--- a/similar/main/gamesave.cpp
+++ b/similar/main/gamesave.cpp
@@ -456,15 +456,15 @@ static void read_object(const vmobjptr_t obj, const NamedPHYSFS_File f, int vers
 				const uint8_t gun_num = ai_info_flags[0];
 				obj->ctype.ai_info.CURRENT_GUN = (gun_num < MAX_GUNS) ? robot_gun_number{gun_num} : robot_gun_number{};
 			}
-			obj->ctype.ai_info.CURRENT_STATE = build_ai_state_from_untrusted(ai_info_flags[1]).value();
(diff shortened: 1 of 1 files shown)


vLKp:
>  I think in every case it would be more reasonable to check that the value doesn't exist and produce a warning/error message instead of crashing or reporting the extremely vague "bad optional access" error.

Yes.  I have wavered as to whether the game should throw exceptions on invalid inputs and, if so, whether to try to catch them or let the program terminate since the inputs should not happen.  Historically, I preferred a fairly pointed reporting mechanism, on the theory that invalid inputs that are quietly coerced to valid may never get reported or, if they are, are reported with such poor detail that they cannot be handled.  Conversely, inputs that provoke a fatal exception are easily traced (the game stops right where it found the problem), and are more likely to be reported.  More recently, I have come to the belief that quietly coercing inputs is a better choice.  However, there is a great deal of code that needs to be revisited to complete the conversion to use of value coercion.

Issue #772, opened 2024-04-22
[Feature Request] Play the Descent 1 PSX movies if present


JeodC:
Add the PSX movies to D1X-Rebirth if they exist. They can be sourced from https://www.descent2.de/files/movies/extra-movies.7z as `.mvl` files. There are five total:

- Interplay logo
- Intro Briefing (comes in two parts)
- Mine exit sequence
- Ending

These can also be viewed [on YouTube](https://www.youtube.com/watch?v=onk85Z6NGRI). If the `.mvl` files are not usable, I can look into obtaining them in a different format.

JeodC:
I wonder if it would also be possible to allow briefing audio to be played if movies aren't present. See https://github.com/RyuKojiro/d2x-xl/blob/5b61c4c12118a66b1c09b53baf70cbd21f488463/main/briefings.cpp#L359 where briefing audio is played when `extra.hog` is present, sourced from https://www.descent2.de/files/gamedata/extra-hog.7z

I think ideally these would be toggleable in the ini files. I am happy to redesign the `extra.hog` file into a `dxa` file, the same with the `extra-movies` archive. At present, `extra.hog` is structured like so:

```
brf01.wav
brf02.wav
brf03.wav
brf04.wav
brf05.wav
brf06.wav
brf07.wav
brf08.wav
brf09.wav
brf10.wav
brf11.wav
brf12.wav
brf13.wav
brf14.wav
brf15.wav
brf16.wav
brf17.wav
brf18.wav
brf19.wav
brf20.wav
brf21.wav
brf22.wav
brf23.wav
brf24.wav
brf25.wav
brf26.wav
brf27.wav
brf28.wav
brf29.wav
brf30.wav
brfcine.wav
enemy01.wav
enemy02.wav
enemy03.wav
enemy04.wav
enemy05.wav
enemy06.wav
enemy07.wav
enemy08.wav
enemy09.wav
enemy10.wav
enemy11.wav
enemy12.wav
enemy13.wav
enemy14.wav
```

It contains all the audio for Descent 1 briefings (27 brf + 3 secret brf) and the 14 robot briefings.

PR #839, opened 2026-04-09
Add d1x psx movie playback and briefing audio

JeodC:
This addresses #772 by supporting dxa archives for psx mve files and wav briefing audio. The libmve used with d2x has been copied and expanded to accommodate building, and briefing audio has been set up to play in correspondence with message numbers to avoid creating a mapping table in code. Instead, the following changes were made for the dxa addons:

- When sourcing the psx movies from https://www.descent2.de/files/movies/extra-movies.7z the mve files were extracted from the mvl archives and added directly to a dxa
- When sourcing briefing audio from https://www.descent2.de/files/gamedata/extra-hog.7z the wav files were extracted from the hog and renamed to align with message nums and robot ids in rebirth code, and then added directly to a dxa

Addons are available [here](https://github.com/JeodC/RHH-Ports/tree/main/ports/released/descent-1-2/addons/descent/other) to test with. I tested with Windows and verified functionality, although I noted an application hang when exiting via the window close button during briefings. This appears to be a pre-existing hang.

Binaries: https://github.com/JeodC/dxx-rebirth/actions/runs/24197653609

Test:

- Add dxa archives to data folder or exe folder
- View movies
- Add `-nomovies` to `d1x.ini` to test briefing audio instead or don't use the movies dxa archive at all


vLKp:
I am not aware of any bugs that would cause a hang on exit like that.  Could you elaborate on how to trigger it, so that I can look into it?  Is it visible on Linux too, or only on Windows?

---

This seems to introduce a `-Werror=unused` failure with `clang` due to `static constexpr bool valid_palette_color_range(const unsigned start, const unsigned count)` being defined in d1x by your commit, but none of the uses were copied over.  I will deal with this when next I work on this area.

---

Thank you for this feature.  I had never seen the PSX movies before.

JeodC:
> I am not aware of any bugs that would cause a hang on exit like that. Could you elaborate on how to trigger it, so that I can look into it? Is it visible on Linux too, or only on Windows?

I only tested it with Windows, but I had d1x-rebirth in windowed mode, not fullscreen, and closed the application via the title  bar's close button during a briefing. Simple to reproduce.

Issue #741, opened 2023-11-20
[Request] Use joystick or keyboard to scroll through level select


JeodC:
It appears there is no way to scroll through a list of levels via integers at level select either with joystick or with keyboard keys (up/down arrow maybe). Could this be easily implemented? Would benefit raspberry-pi and other devices without keyboard.

JeodC:
This would occur around here: /similar/main/menu.cpp#L993 and look something like this:

#include <conio.h>

int main() {
    int number = 1;

    while (true) {
        // Display the current value of the integer
        std::cout << "Current value: " << number << std::endl;

        // Check for keyboard input
        if (_kbhit()) {
            char key = _getch();

            // Check for arrow key
            if (key == 0xE0) {
                key = _getch();  // Read the actual key code

                // Handle arrow keys
                switch (key) {
                    case 72:  // Up arrow
                        number += 1;
                        break;
                    case 80:  // Down arrow
                        number -= 1;
                        break;
                }
            }
        }
    }

    return 0;
}

vLKp:
I believe typing the first character of the target listbox item will jump to it.  Since items can start with a number, adding integer-based selection would conflict with this.  It would also need special rules for how to handle when the user has more than 9 items.

---

The line number you cite seems to be in the middle of `do_new_game_menu` which, despite its name, is for prompting the user what level of a mission to play.  I initially read your original request to be about picking which mission to play.  Is that wrong?  Are you trying to start on a level higher than level 1 after picking a mission, but you are unable to do so because you have no keyboard?

Are you proposing the use of `_getch` in Rebirth, or just giving this as an example?  `_getch` does not seem to be standard on Linux.  ncurses provides `getch`, but that would likely not work on Windows.  More generally, `getch` is about reading from a keyboard, and your request seems to indicate you have no working keyboard, so I do not understand how adding an alternate way to read the keyboard is helpful.

JeodC:
It's just an example. But yeah I ported this to a retro handheld with no kbm controls, so level select (where you type a number at New Game) isn't working.

JeodC:
![image](https://github.com/dxx-rebirth/dxx-rebirth/assets/47716344/1569a332-2ace-4244-af4e-11c426841734)

vLKp:
I see.  So when you press the down arrow on the controller, it generates a keyboard event, not a joystick hat motion?  What keyboard key does it simulate?  The down arrow that numpad 2 generates when numlock is off?

[...]

PR #840, opened 2026-04-10
Joystick input overhaul

JeodC:
This is a pretty big overhaul and will need some extensive testing, but if it works well then it'll make DXX-Rebirth far more user-friendly with modern hardware and consoles.

---

- If built with SDL2, use `SDL_GAMECONTROLLER` over raw joystick. This means `BUTTON A` is always the same logical button regardless of button index.
- Bundle `gamecontrollerdb.txt` with builds, CI actions fetch latest from upstream
- CI actions build with sdl2, sdl1.2 still an option
- When building with SDL1.2, raw joystick behavior is retained
- New file `gamecontroller.cpp` for modern input handling, separate from `joy.cpp` to make organization and troubleshooting easier
- Guidebot menu changed to `newmenu_do2` to allow for gamepad navigation
- Guidebot menu button mapped to `L3` for gamepad (hardcoded via button-to-key)
- Default joystick controls remapped for modern controllers (SDL2 only)
- Modern button labels for joystick control config (SDL2 only)
- Joystick config menu help text re-labeled for gamepad support for deleting and resetting mappings
- Joystick config mapping prompt moved down a bit so it doesn't overlap mapping cells
- Level select in New Game is now a slider for gamepad support (resolving #741)
- Save games have a default name for gamepad support (KBM can still delete the default name and type custom names)
- In menus, `BACK` and `B` are mapped to `KEY_ESC`

---

I tested this with Windows built on top of the latest commit, with a wired PowerA Xbox One controller. Linux and Windows artifacts available at https://github.com/JeodC/dxx-rebirth/actions/runs/24242395539. MacOS is failing to build pending a separate pull request that's already submitted.


JeodC:
Screenshot 2026-04-10 070100

Screenshot 2026-04-10 070122

vLKp:
Thank you for this.  I had been thinking for quite some time that the SDL2 hotplug support for joysticks would be nice to have, but I never worked on it.  Your changes here look good in that regard.

---

I have not merged this yet, because your changes caused issues elsewhere:

- The guidebot menu no longer accepts keyboard shortcuts to pick the action.  I will fix this.
- The single player level slider triggers a gcc _FORTIFY_SOURCE warning because gcc fails to prove that you allocated enough space.  (Your code is fine.  gcc just cannot prove it.)  I will fix this.
- The single player level slider no longer takes numeric input, so I cannot type `20` to start on level 20.  I have an idea how to fix this, but will not be able to complete it today.

JeodC:
> Thank you for this. I had been thinking for quite some time that the SDL2 hotplug support for joysticks would be nice to have, but I never worked on it. Your changes here look good in that regard.
> 
> I have not merged this yet, because your changes caused issues elsewhere:
> 
> * The guidebot menu no longer accepts keyboard shortcuts to pick the action.  I will fix this.
> * The single player level slider triggers a gcc _FORTIFY_SOURCE warning because gcc fails to prove that you allocated enough space.  (Your code is fine.  gcc just cannot prove it.)  I will fix this.
> * The single player level slider no longer takes numeric input, so I cannot type `20` to start on level 20.  I have an idea how to fix this, but will not be able to complete it today.

If you'd like me to resolve any of these before merge, let me know.

similar/main/menu.cpp
```diff
@@ -1011,7 +1011,10 @@ window_event_result do_new_game_menu()
 				}
 				virtual window_event_result event_handler(const d_event &event) override
 				{
-					auto result = newmenu::event_handler(event);
+					const auto alive = track();
```
vLKp, line 1014: Under what circumstances was this crash observed?

JeodC: Linux arm. Reproduced by navigating menus quickly.

vLKp:
My above push brings in the originally proposed commit, as well as some changes and fixes on top of it.  In the interim, you pushed new commits, so GitHub did not automatically close this MR when I pushed.  I will look at the new commits soon.

While testing this, I found that the original MR caused Rebirth to ignore my joystick completely.  I eventually traced this down to being that my joystick is not considered a gamecontroller, so the MR code caused Rebirth not to open it as a gamecontroller (because it is not one) nor as a joystick (because the MR disabled opening traditional joysticks).  To address this, one of my changes restores the use of the traditional low-level joystick code when the game detects a joystick that is not a gamecontroller.  I think that the game will still not try to open a gamecontroller as a joystick, but I have no gamecontroller with which to test this.

JeodC:
> My above push brings in the originally proposed commit, as well as some changes and fixes on top of it. In the interim, you pushed new commits, so GitHub did not automatically close this MR when I pushed. I will look at the new commits soon.
> 
> While testing this, I found that the original MR caused Rebirth to ignore my joystick completely. I eventually traced this down to being that my joystick is not considered a gamecontroller, so the MR code caused Rebirth not to open it as a gamecontroller (because it is not one) nor as a joystick (because the MR disabled opening traditional joysticks). To address this, one of my changes restores the use of the traditional low-level joystick code when the game detects a joystick that is not a gamecontroller. I think that the game will still not try to open a gamecontroller as a joystick, but I have no gamecontroller with which to test this.

Ah, I do not have a retro joystick, so of course this would be an unforseen problem. Hopefully my other commits speak for themselves.

Issue #841, opened 2026-04-13
Runtime bug: Title screen music doesn't play


Kreeblah:
### Environment

Git commit https://github.com/dxx-rebirth/dxx-rebirth/commit/3df6691c41b4ff7de252e6c2ea4afc10fda1446d with patches from https://github.com/dxx-rebirth/dxx-rebirth/commit/53130e805a57cf3875cfb6be690541c84fa0ec54 and https://github.com/dxx-rebirth/dxx-rebirth/commit/b749eadb4080f596ce90ef7b2be97d7c1213567f, as well as deleting the `valid_palette_color_range` `constexpr` definition in `d1x-rebirth/main/movie.cpp`.

#### Operating System Environment

* Microsoft Windows
  * [ ] Microsoft Windows XP
  * [ ] Microsoft Windows Vista
  * [ ] Microsoft Windows 7
  * [ ] Microsoft Windows 8
  * [ ] Microsoft Windows 8.1
  * [ ] Microsoft Windows 10

* Mac OS X
  * [ ] Mac OS X 10.10
  * [ ] Mac OS X 10.11
  * [ ] Mac OS X 10.12
  * [ ] Mac OS X 10.13
  * [ ] Mac OS X 10.14
  * [ ] Mac OS X 10.15
  * [x] macOS 26.4.1

* Debian
  * [ ] Debian Bookworm
  * [ ] Debian Trixie
  * [ ] Debian Sid
* Fedora
  * [ ] Fedora 36
  * [ ] Fedora 37
  * [ ] Fedora 38
  * [ ] Fedora 39
  * [ ] Fedora 40
  * [ ] Rawhide
* Ubuntu
  * [ ] Ubuntu 20.04 LTS (Focal Fossa)
  * [ ] Ubuntu 22.04 LTS (Jammy Jellyfish)
  * [ ] Ubuntu 23.10 (Mantic Minotaur)
  * [ ] Ubuntu 24.04 LTS (Noble Numbat)

* [ ] Arch
* [ ] Gentoo

* [ ] OpenBSD
* [ ] FreeBSD
* [ ] NetBSD

#### CPU environment

* [ ] x86 (32-bit Intel/AMD)
* [ ] x86\_64 (64-bit Intel/AMD)
* [ ] ARM (32-bit)
* [x] ARM64 (64-bit; sometimes called AArch64)

#### Game environment

Title screen

### Description

[...]

PR #842, opened 2026-04-13
D1X movies patch

JeodC:
Addresses #841

- Title music plays after movies
- Movies stretch to fill screen (sized differently than d2x)
- Clicking X in window bar during briefing closes briefing and skips to gameplay (not ideal, should just close the app, but avoids the hang)
- Fix macos build failure and ci failure where build would succeed but ci would fail

https://github.com/JeodC/dxx-rebirth/actions/runs/24340838318

### DIFF
--- a/common/include/compiler-poison.h
+++ b/common/include/compiler-poison.h
@@ -1,9 +1,6 @@
 #pragma once
 
-#if DXX_HAVE_POISON
 #include <algorithm>
-#endif
-
 #include <cstdint>
 #include <memory>
 #include <span>
--- a/common/include/event.h
+++ b/common/include/event.h
@@ -10,6 +10,7 @@
 
 #include <SDL_version.h>
 #include "fwd-event.h"
+#include "fwd-window.h"
 #include "maths.h"
 
 #if SDL_MAJOR_VERSION == 2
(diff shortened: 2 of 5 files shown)


vLKp:
I merged up through https://github.com/dxx-rebirth/dxx-rebirth/pull/842/commits/69cf6044b13b089e1cfd328700709d58f0711b91 back in a33de9c17f10cb4393c5b4629f232c8b77b078f5.

For the two commits in this pull request that have not been merged:
- https://github.com/dxx-rebirth/dxx-rebirth/pull/842/commits/7a9506790001f991fbedef3d735981d1ec36f072 - when does the application reach a state of having zero windows, but still being in this function?
- https://github.com/dxx-rebirth/dxx-rebirth/pull/842/commits/4d1c491ced5273ac615eb04e5051cb2cb0d6cacf - this tries to solve two problems:
  - `compiler-poison.h`: I made commit 8c4aff8695b657804d5dda51d0ced46d16c61810 to address this because it was also failing on Linux when using clang's libc++.
  - `loadgl.h`: I made commit 1ed93296b7e31bff94582d88c6fec8d268609c94 to address this.  I was wrong to have the code ever reach the point that OS X even tried to include a GL header via `loadgl.h`, and that only happened when OS X ran a `check_headers_include=1` build, which tries to test every header it can find.  The game does not normally need this header to work on OS X, so I broadened the preprocessor exclude to prevent OS X from trying to pull any GL header here.  That seems more future-proof than referring to an OS X header that is not used by the game for anything.

JeodC:
https://github.com/dxx-rebirth/dxx-rebirth/commit/7a9506790001f991fbedef3d735981d1ec36f072 - Four months later, I couldn't tell you. I'd have to assume it's due to how movie playback functions. I vaguely recall hitting the exit option from the main menu and then experiencing a hang, but my memory on the commit is foggy.

PR #849, opened 2026-08-03
MVE: preserve SDL2 movie audio resampler state

data8504:
## Problem

The SDL_mixer movie path creates a fresh `SDL_AudioCVT` for every MVE audio segment.  When movie audio is resampled to the mixer rate, that resets the resampler at every segment boundary and produces quiet clicks, most noticeably during loud passages in the Descent II intro.  Using `-nosdlmixer` avoids this path, but also gives up SDL_mixer music playback.

## Changes

- Serialize the movie producer and audio callback on the ring, and refuse a new buffer before it would overwrite an unread entry.
- Use one SDL2 `SDL_AudioStream` for the lifetime of the movie and drain converted audio after each segment, preserving resampler state across MVE boundaries.
- Leave the SDL1 and non-SDL_mixer conversion paths unchanged.
- Add a Boost runtime contract test confirming that persistent SDL2 conversion produces identical output for different input segment boundaries.

## Validation

- Full D2X build with GCC 15, SDL 2.32.10, and SDL_mixer 2.8.1.
- D2X MVE playback source compiled with SDL_mixer disabled.
- D1X shared MVE playback source compiled with SDL2 and SDL_mixer enabled.
- `test-mve-audio-stream` built through `register_runtime_test_link_targets=1` and passed.
- Repeated manual Descent II movie playback, including early aborts and the main-menu idle loop: clicks are gone, movie audio remains normal, and music and game audio remain normal after each playback.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -3637,10 +3637,10 @@ class DXXCommon(LazyObjectConstructor):
 	class RuntimeTest(LazyObjectConstructor):
 		target: str
 		source: collections.abc.Sequence[str]
-		nodefaultlibs: bool = True
-		def __init__(self, target: str, source: collections.abc.Sequence[str]):
+		def __init__(self, target: str, source: collections.abc.Sequence[str], nodefaultlibs: bool = True):
 			self.target = target
 			self.source = LazyObjectConstructor.create_lazy_object_getter(source)
+			self.nodefaultlibs = nodefaultlibs
 
 	@cached_property
 	def program_message_prefix(self):
@@ -4930,6 +4930,9 @@ class DXXArchive(DXXCommon):
 		RuntimeTest('test-enumerate', (
 			'common/unittest/enumerate.cpp',
 			)),
+		RuntimeTest('test-mve-audio-stream', (
+			'd2x-rebirth/unittest/mve_audio_stream.cpp',
+			), nodefaultlibs=False),
 		RuntimeTest('test-serial', (
 			'common/unittest/serial.cpp',
 			)),
--- a/d2x-rebirth/libmve/mveplay.cpp
+++ b/d2x-rebirth/libmve/mveplay.cpp
@@ -32,7 +32,6 @@
 #endif
 
 #include "config.h"
-#include "d_sdl_audio.h"
 #include "mvelib.h"
 #include "mve_audio.h"
 #include "byteutil.h"
@@ -237,6 +236,7 @@ static void mve_audio_callback(void *userdata, unsigned char *stream, int len);
 static void mve_audio_callback(void *const vstream, unsigned char *stream, int len)
 {
 	auto &mvestream = *reinterpret_cast<MVESTREAM *>(vstream);
+	const std::lock_guard lock{mvestream.mve_audio_mutex};
 #ifdef DXX_REPORT_TOTAL_LENGTH
 	int total{0};
 #endif
@@ -298,6 +298,7 @@ static void mve_audio_callback(void *const vstream, unsigned char *stream, int l
 
 bool MVESTREAM::queue_mve_audio_buffer(::dcx::unique_span<int16_t> &&buffer)
 {
+	const std::lock_guard lock{mve_audio_mutex};
 	auto next = mve_audio_buftail + 1;
 	if (next == mve_audio_buffers.size())
 		next = 0;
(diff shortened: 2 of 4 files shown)


SConstruct
```diff
@@ -3593,9 +3593,10 @@ class DXXCommon(LazyObjectConstructor):
 		target: str
 		source: collections.abc.Sequence[str]
 		nodefaultlibs: bool = True
```
vLKp: I think the class-level variable can be removed now that the `__init__` method always sets an instance-level variable `nodefaultlibs`.

d2x-rebirth/libmve/mveplay.cpp
```diff
@@ -296,6 +296,53 @@ static void mve_audio_callback(void *const vstream, unsigned char *stream, int l
...
+		return false;
+	}
+	if (converted & 1)
+	{
+		con_printf(CON_URGENT, "%s:%u: SDL_AudioStreamGet returned odd byte count: %d", __FILE__, __LINE__, converted);
+		return false;
+	}
+	if (!converted)
+		return true;
+	::dcx::unique_span<int16_t> output{static_cast<std::size_t>(converted) / sizeof(int16_t)};
```
vLKp: `unique_span` is able to take ownership of an existing compatible `std::unique_ptr`.  It looks like this function could do:

```c++
    auto output{std::make_unique<int16_t[]>((available + 1) / 2)};
    const auto converted{SDL_AudioStreamGet(mve_audio_stream.get(), output.get(), available)};
    // Same error handling tests
    // No `memcpy`
    return queue_mve_audio_buffer(unique_span<int16_t>{std::move(output), converted / 2});
```

Is there a reason to use the `temporary` buffer?  Writing directly to an allocated `int16_t[]` would avoid one memory copy, although it would mean the buffer remains sized to `available`, which could be greater than `converted`.

d2x-rebirth/libmve/mveplay.cpp
```diff
@@ -350,19 +398,42 @@ MVESTREAM::handle_result MVESTREAM::handle_mve_segment_initaudiobuffers(unsigned
...
 
 MVESTREAM::handle_result MVESTREAM::handle_mve_segment_startstopaudio()
 {
-	if (mve_audio_spec && !mve_audio_playing && mve_audio_bufhead != mve_audio_buftail)
+	bool have_queued_audio;
+	{
+		const std::lock_guard lock{mve_audio_mutex};
+		have_queued_audio = mve_audio_bufhead != mve_audio_buftail;
+	}
+	if (mve_audio_spec && !mve_audio_playing && have_queued_audio)
```
vLKp: I see the intent to acquire the lock before examining the state variable, and why that would be useful.  How often will this function be entered when `mve_audio_spec && !mve_audio_playing` is false?  If that is relatively common, it would be nice to reorder this so that the lock is only acquired if `have_queued_audio` needs to be computed.  When one of the first two conditions fails, the function skips this block without needing to know the value of `have_queued_audio`.

vLKp (commented):
Thank you for the analysis, and the well written pull request.  I have some minor questions before merging, but can handle the adjustments if you would prefer not to work on this further.

For my curiosity, was this written with LLM assistance or is this entirely manual?  This repository presently has no policy on LLM usage, and I ask mainly because I want to understand whether there are LLMs that can produce code of the quality seen here.

SConstruct
data8504: Agreed. Now that every construction path assigns `self.nodefaultlibs`, the class-level default is redundant. I have removed it.

data8504:
Thank you for the generous feedback and the careful review. I was happy to continue working on the adjustments; all three suggestions make sense, and I have pushed them and revalidated the affected build and test paths.

Yes, this was developed with substantial LLM assistance. I used OpenAI Codex running GPT-5.6 Sol at medium reasoning. Codex produced almost all of the implementation and much of the analysis and PR write-up. I identified the problem, set the direction and constraints, required careful validation and consistency with the project's existing patterns, and reviewed the resulting work closely. For context, I am a software engineer at Intel, though this is a personal contribution and I speak only for myself.

The process was iterative: Codex inspected the surrounding implementation and project conventions, proposed and implemented changes, and I reviewed the assumptions, code, and validation results at each stage. I am glad to disclose that process and give Codex appropriate credit. I appreciate the project and the time you have taken to review the contribution so thoughtfully.

d2x-rebirth/libmve/mveplay.cpp
data8504: There was no functional need for the temporary buffer. I introduced it so that a legal short result from `SDL_AudioStreamGet` could be represented with the exact converted extent, but I overlooked that `unique_span` can own the larger allocation while exposing a smaller logical extent. Your suggestion preserves the short-read handling and removes the copy. I have changed it to allocate `int16_t[]` directly and construct the `unique_span` with `converted / sizeof(int16_t)` after the existing error and even-byte checks. Thank you for catching this.

d2x-rebirth/libmve/mveplay.cpp
data8504: I do not have a reliable frequency measurement across the movie corpus. This handler runs only when a start/stop-audio opcode is decoded, rather than for every audio frame, but there is no reason to acquire the mutex when either of the first two conditions fails. I have reordered the checks so that the ring state is examined under the mutex only when an audio specification exists and playback has not started.

vLKp:
This looks good.  I will get it merged when next I work on the code.

If you have records of the interactions with the LLM, and are willing to share them, they might be edifying for future users on how to properly guide an LLM to work on this project.  Sharing those interactions is not a requirement for merging the PR, since the changes look valid on their own.  They are primarily for my curiosity, and possibly the benefit of other users later.

data8504:
Sure—here it is: [DXX-Rebirth Codex transcript (.md)](https://gist.github.com/data8504/90bf6e5b70ee5b8e26b8a36139fc5e5f)

For orientation, it runs to 1,162 lines:

- Lines 1–58: scope, redaction notes, session map, and relevant AGENTS/environment context.
- Lines 59–298: initial discovery—upstream architecture, original media, LAN conventions, and the deployment plan.
- Lines 299–453: publishing the retail assets, installing the native engines, tightening permissions, and integrating the shared data with the stock desktop launchers.
- Lines 454–457: an unrelated tangent, omitted for privacy and relevance.
- Lines 458–735: reproducing and diagnosing the Descent II movie-audio clicks, isolating SDL_mixer, implementing the fix, testing it, packaging it, and validating it on the canary.
- Lines 736–824: assessing the patch’s suitability for upstream and the project’s maintenance activity.
- Lines 825–828: a credential-management tangent, omitted for privacy.
- Lines 829–944: shaping the three-patch series, independent review, and creation of PR #849.
- Lines 945–1151: responding to your review, revising and rebuilding, live validation, and updating the PR and its comments.
- Lines 1152–1162: a final unrelated omission and editorial notes.

It has been redacted with a light but deliberate hand: private LAN identities, topology, and paths are generalized; tool calls and outputs, private model reasoning, and three unrelated tangents are omitted. The visible prompts, steers, progress updates, and replies remain.

To keep it honest, I left the misspellings, agent idiosyncrasies, and my personality intact.

I hope it is useful.

PR #851, opened 2026-08-15
Silence -Wuseless-cast on i386

joshuamegnauth54:
Some platforms still use an i32 for time_t for i386. Those platforms fail to build with -Werror, so I silenced the lint since the cast is sound and the warning is spurious.

---

Hi, and thanks for the amazing source port! The direct cause of this PR is the i386 build for FreeBSD [failing](https://portsfallout.com/port/11440/). Fixing it seemed simple enough. I opted to fix it here instead of squelching the lint in FreeBSD's Makefile because it captures the intent better than simply silencing _all_ of `-Wuseless-cast`. Let me know if you'd prefer that I just fix it for FreeBSD directly. :grin:

### DIFF
--- a/similar/main/game.cpp
+++ b/similar/main/game.cpp
@@ -992,7 +992,7 @@ void save_screen_shot(int automap_flag)
 	unsigned tm_year;
 	const auto t = time(nullptr);
 	struct tm *tm = nullptr;
-	if (t == static_cast<time_t>(-1) || !(tm = gmtime(&t)))
+	if (t == time_t{-1} || !(tm = gmtime(&t)))
 		tm_year = tm_mon = tm_mday = tm_hour = tm_min = tm_sec = 0;
 	else
 	{


vLKp:
Thank you for the report, and the pull request.  I appreciate that you tried to avoid changing the code.  However, before suppressing the warning, I want to experiment with a minor code change.  I cannot reproduce the reported warning on 32-bit Linux or 64-bit Linux.  The literal `-1` will be an `int`.  I suspect that the problem is that on FreeBSD, `time_t` is a typedef for `int`.  You mention `time_t` being an `i32`, but on 32-bit platforms, `long int` is also _often_ an `i32`, so I cannot be sure from your description which underlying type FreeBSD is using.  On 32-bit Linux, `time_t` is a typedef for `long int`, which is why this cast is allowed on Linux, since it casts from `int` to `long int`, which the compiler considers distinct even when they are both 32-bit wide signed integers.

Please try the below patch in your environment.  If it works, I will push it to `master` and cite this pull request.  If it does not help, please post the output of the failure.  If we cannot get this to work readily, I will merge the pull request to disable the warning here.

```diff
diff --git a/similar/main/game.cpp b/similar/main/game.cpp
index 504de7589..51cd44203 100644
--- a/similar/main/game.cpp
+++ b/similar/main/game.cpp
@@ -992,7 +992,7 @@ void save_screen_shot(int automap_flag)
 	unsigned tm_year;
 	const auto t = time(nullptr);
 	struct tm *tm = nullptr;
-	if (t == static_cast<time_t>(-1) || !(tm = gmtime(&t)))
+	if (t == time_t{-1} || !(tm = gmtime(&t)))
 		tm_year = tm_mon = tm_mday = tm_hour = tm_min = tm_sec = 0;
 	else
 	{
```

joshuamegnauth54:
Hi again! Replacing the `static_cast` with `time_t{-1}` works and looks cleaner too! I force pushed with that update.

vLKp:
This change is now in `master`.  Unfortunately, I had written and committed my change before your force-push, so the history does not record your commit (though history does point to this PR).

Closing, since the required change is now pushed.

PR #853, opened 2026-08-22
`newmenu_mouse`: do not activate prior item when clicking text

### LINKED ISSUE #830: If you save to slot 10, the load screen shows 11 slots
See the red highlight below, "manual" was saved to slot 10. Also clicking on the 11th slot ( -empty- ) loads slot 10.

Image

data8504:
## Problem

On mouse-down, a non-selectable text row leaves `citem` pointing to the previous selectable row.  On mouse-up, the menu could then dispatch a select event for that previous row while the pointer was over the text row.  In the load-game menu, clicking the empty autosave row could therefore load save slot 10.

## Change

Only dispatch the mouse select event when the hit row is also the current menu item.  This leaves non-selectable text rows inert without changing their display or the normal activation path for selectable rows.

Fixes #830.

## Validation

- Built D1X-Rebirth and D2X-Rebirth with GCC 15, SDL 2, and SDL_mixer on Ubuntu 26.04.
- In both games, saved in slot 10, aborted to the main menu, and reopened the load screen.
- Clicking the 11th, empty autosave row did nothing; clicking slot 10 loaded it; and leaving the load screen still worked normally.

> AI assistance disclosure: I prepared this change and PR description with assistance from OpenAI Codex, operating as a distinct coding agent. Codex analyzed the issue and source, helped implement and review the fix, and verified the recorded build evidence. I performed the D1X/D2X runtime validation described above before submission.

### DIFF
--- a/similar/main/newmenu.cpp
+++ b/similar/main/newmenu.cpp
@@ -989,22 +989,20 @@ static window_event_result newmenu_mouse(const d_event &event, newmenu *menu, co
 			{
 				const auto [mx, my, mz] = mouse_get_pos();
 				const int line_spacing = static_cast<int>(LINE_SPACING(*canvas.cv_font, *GAME_FONT));
-				for (int i = menu->scroll_offset; i < menu->max_on_menu + menu->scroll_offset; ++i)
+				auto &iitem = *std::next(menu->items.begin(), menu->citem);
+				x1 = canvas.cv_bitmap.bm_x + iitem.x - fspacx(13);
+				x2 = x1 + iitem.w + fspacx(13);
+				y1 = canvas.cv_bitmap.bm_y + iitem.y - (line_spacing * menu->scroll_offset);
+				y2 = y1 + iitem.h;
+				if (((mx > x1) && (mx < x2)) && ((my > y1) && (my < y2)))
 				{
-					auto &iitem = *std::next(menu->items.begin(), i);
-					x1 = canvas.cv_bitmap.bm_x + iitem.x - fspacx(13);
-					x2 = x1 + iitem.w + fspacx(13);
-					y1 = canvas.cv_bitmap.bm_y + iitem.y - (line_spacing * menu->scroll_offset);
-					y2 = y1 + iitem.h;
-					if (((mx > x1) && (mx < x2)) && ((my > y1) && (my < y2))) {
-							// Tell callback, allow staying in menu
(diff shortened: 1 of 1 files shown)


similar/main/newmenu.cpp
```diff
@@ -996,7 +996,7 @@ static window_event_result newmenu_mouse(const d_event &event, newmenu *menu, co
 					x2 = x1 + iitem.w + fspacx(13);
 					y1 = canvas.cv_bitmap.bm_y + iitem.y - (line_spacing * menu->scroll_offset);
 					y2 = y1 + iitem.h;
-					if (((mx > x1) && (mx < x2)) && ((my > y1) && (my < y2))) {
+					if (i == menu->citem && ((mx > x1) && (mx < x2)) && ((my > y1) && (my < y2))) {
```
vLKp: I think this is not the right solution.  This patch is inside a `for` loop that iterates over some slice of the menu items.  There will be at most one element in that slice for which the new test of `i == menu->citem` is true.  Therefore, if the fix is to only accept that one element, then the `for` loop should be removed completely, because there is no value to computing the bounding boxes of elements that will be disqualified by this test of `i == menu->citem`.

I have not attempted to determine whether accepting only the item indicated by `menu->citem` is the right fix for the underlying problem.

vLKp (changes requested):

similar/main/newmenu.cpp
data8504: You’re right—the loop became redundant once activation was constrained to `citem`. I replaced it with a direct bounds check for the current item, matching the mouse-up pattern already used by `listbox_mouse`.

I also verified the underlying behavior: mouse-down selects an actionable row, while a text row preserves the previous selection; mouse-up now activates only when released within the current item’s bounds. This keeps text rows inert and prevents activation when dragging from a selectable row onto another row.

I rebuilt both engines and manually validated D1X and D2X: the empty autosave row remains inert, slot 10 loads normally, Escape still exits the load screen, and pressing slot 10 then dragging to and releasing over the empty row does not load anything.

vLKp: That looks good.  I will merge this when next I am in the area.  While setting up to test this, I encountered an unrelated (and likely *very* old) bug around loading a game during combat.  I got side-tracked reproducing that bug, and working on a fix.

Issue #488, opened 2020-01-05
D2X-Rebirth Dec 27 2019: Homing projectiles/possibly robot projectiles in general (?) seem easier to dodge versus 0.58.1


kakhome1:
In the latest Rebirth, it seems somewhat easier than in 0.58.1 (or Dos) to evade Homing Missiles or other homing projectiles. At least in Descent II, haven't tested Descent I. I'm not sure if they are moving a bit slower than they should or don't home as strongly. Might be slightly easier to dodge robot fire in general in fact but noticed it most for homing shots.

kakhome1:
(Probably need to check robot/projectile/ship speeds and homing 'box of sight' to confirm if anything is amiss here or not...I don't know how to do that myself...it just seemed easier to dodge)

vLKp:
Homing weapons are a perpetual source of argument in the Descent community.  @zicodxx reworked them after 0.58.1 to try to make them more like DOS.

If you can present an analysis of their performance in both 0.58.1 and tip, I will look at it.

zicodxx:
I have spent the better part of the day comparing the homing code (turn, tracking) and the majority of the weapon code in general with the original Descent and Descent 2 source releases. Apart from the obvious fixes to only execute homing related code parts (i.e. turning and tracking) based on a 30 FPS interval, Rebirth seems to match the original game fine. 0.58.1 used a different approach in that it tried to execute the homing code in every frame and instead scale the vector added to target by FrameTime. The scaling was off and 0.58.1 didn't produce faithful results. 

To my best knowledge and belief, I would say the current base reproduces the original game "as if running on a constant 30 FPS setting" faithfully.

I did also try and compare to the DOS game and apart from it being really hard to run them consistently at 30 FPS (and considering that the Descent 2 tracking behavior changing at 32 FPS), I cannot tell a difference either. 

If there is something off, I cannot find it. I'm sorry.

vLKp:
Given that zico has reviewed this again in detail, and I'm not inclined to doubt his work, I think the only remaining thing that could be done here is if someone wants to collect and present data showing that DOS and tip behave substantially differently at 30 FPS, which is generally considered the standard environment for homing weapons.  @kakhome1: is that something you would be interested in collecting?

kakhome1:
I fully trust zico...but the difference in dodge-ability between Dos and 0.60 is incredible. I posted two videos to YouTube (kakhome1) illustrating.

[...]

PR #860, opened 2026-08-31
homing: acquire and retain targets

data8504:
## Problem

Robot-fired homing weapons begin with `track_goal` unset.  A later target scan can find a valid target, but the returned object was used only for that turn and was not stored, so the missile usually returned to an untracked state on the next homing tick.

Persisting targets also makes the existing object lifecycle checks important: a player can become `OBJ_GHOST`, a remote player can cloak, and the live parent object type can change after a player-fired missile is launched.

## Changes

- Select an eligible target when a robot directly fires a homing weapon, leaving `object_none` when no target is available.  Smart children keep their explicit target assignment path.
- Store later target updates in `track_goal`, so an acquired target remains available on subsequent homing ticks.
- Reject ghost and cloaked player targets, and use the recorded firing identity for cooperative and team filtering after the shooter dies.
- Extract the existing fixed-point turn kernel and add deterministic tests for target selection, initial acquisition, retained reacquisition, no-target behavior, player death/replacement, and D1/D2 trajectory calculations.

Potentially addresses #488, pending gameplay evidence from players familiar with the original behavior.

## Validation

- Clean GCC 15 SDL2 builds of D1X-Rebirth and D2X-Rebirth
- Complete common, D1, and D2 header-include matrix
- 10 focused Boost homing tests
- Independent adversarial review of the final three-commit series

The trajectory coverage is a deterministic functional model of the production fixed-point code; it is not a claim of complete DOS/gameplay parity.

> AI assistance disclosure: I prepared this change with assistance from OpenAI Codex, operating as a distinct coding agent. Codex researched the implementation history, developed the tests and patch series, ran the stated validation, and performed an independent adversarial review. Gameplay evaluation and maintainer review remain external to that work.


vLKp:
I merged this as https://github.com/dxx-rebirth/dxx-rebirth/commit/40ed1ce3055b876bee5fc52b1815ab6cef887b4e.  Due to a last minute mistake on my part while removing an unrelated commit, the actual merged commits were rebased with me as the committer (but the original author is still listed as the author), so the merged commits have different commit IDs, and GitHub did not automatically close this PR.

data8504:
GitHub reports that the original PR head can still be merged cleanly. Since the equivalent changes are already in `master`, I believe merging it now should produce an ancestry-only merge commit, leave the tree unchanged, and allow GitHub to record the PR—and its relationship to the linked issue—as merged. I’d appreciate that for the project history, but if you’d prefer not to add the extra commit, I’ll close it manually.

PR #862, opened 2026-08-31
savegames: preserve cooperative object mappings

### LINKED ISSUE #737: Runtime bug: Powerups can potentially be replicated after loading a savegame in Multiplayer games
### Environment

Git Commit: 3ad507fad88f7475751e2ca94d911c97c62cbc65

NOTE: Both computers are on the same environment.

Operating System: Arch Linux
Kernel Version: 6.5.9-arch2-1 (64-bit)

#### CPU environment

* [X] x86\_64 (64-bit Intel and AMD)

#### Game environment

descent2.hog:

- Obtained from: GoG
- Size: 7.2 MiB (7,595,079)
- SHA256: f1abf516512739c97b43e2e93611a2398fc9f8bc7a014095ebc2b6b2fd21b703

### Description

Multiplayer games can replicate powerups only for the client-side, but affecting the multiplayer game. This happens after loading a savegame.

#### Regression status

Unknown.

### Steps to Reproduce

1. Start a multiplayer game on "Descent 2: Counterstrike!" on map24
2. Kill the robot that drops one Invulnerability
3. Save the game
4. Load the game
5. Check if both players can take one Invulnerability each

* Engines affected (potentially both. I saw this happening on D2X-Rebirth):
  * [X] D1X-Rebirth
  * [X] D2X-Rebirth

* Game modes affected (potentially all multiplayer modes, couldn't test):
  * [ ] Single player
  * [X] Multiplayer competitive
  * [X] Multiplayer cooperative

data8504:
## Problem

Cooperative savegames restore the object table, but not the network identity mapping for dynamically created objects. A robot-dropped powerup can therefore be object 1 on its creator and object 259 on a peer before the save; after restore, each peer treats its own local slot as authoritative. When one player collects the powerup, the removal message names the wrong identity for the other peer, allowing both players to collect it.

## Changes

- Append each cooperative object's creator and creator-local object number to the savegame.
- Remap saved creator slots to the current player slots by callsign and rebuild the network object tables after restore.
- Validate the complete mapping before replacing live network state, and fail a cooperative restore cleanly on truncated, out-of-range, duplicate, dead-object, or inconsistent mappings.
- Add focused tests for preplaced, local, remote, permuted-player, duplicate, invalid, and dead-object cases.

This bumps the D1 save version from 7 to 8 and the D2 save version from 22 to 23. The new cooperative data is appended after the prior format.

Fixes #737.

## Validation

- Reproduced the reported D2 Counterstrike Level 24 failure on upstream `e5744892` with two isolated UDP peers and the GOG data identified in the issue: after a real robot drop and cooperative save/restore, both peers received the same invulnerability.
- Repeated the same scenario with this change. The creator restored the drop as local object 1 and the peer as local object 259, while both restored the shared identity `(owner 0, remote 1)`. In both guest-first and host-first pickup orders, only the first peer received the invulnerability and the second peer no longer found the object.
- Loaded the prior D2 version-22 cooperative save into the new build on both peers and resumed gameplay.
- Passed five focused Boost test cases for mapping validation and player-slot permutation.
- Built D1X-Rebirth and D2X-Rebirth in release mode with GCC and SDL2.
- Passed the complete common, D1, and D2 header-include matrix.
- Completed an independent adversarial review of the final change and validation evidence.

The direct gameplay reproduction was performed in D2. D1 shares the repaired save/mapping path and was covered by build, header, and focused mapping tests, not a separate gameplay run.

> AI assistance disclosure: I prepared this change and PR description with assistance from OpenAI Codex, operating as a distinct coding agent. Codex investigated and reproduced the two-peer failure, implemented the save-format and restore changes, developed the focused tests and deterministic gameplay fixture, ran the validation listed above, and performed an independent adversarial review. Gameplay evaluation beyond the reported D2 scenario and maintainer review remain external to that work.


vLKp:
I concur that the underlying problem exists, and needs to be fixed.  I have always been wary of modifying the savegame format, because the loader is so fragile.  I still need to review this.

